PR #10, opened 2018-05-30
Updated to allow homework via PR

remarcmij:
This PR is based on the JS3 module as given to class 14. Up to Laurens whether he want to use this or something completely else. A worked-out solution of the homework can be found in the (private) repo [GitHubApp](https://github.com/HackYourFuture/GitHubApp).


Razpudding:
Thanks Jim! The worked out solution looks pretty fancy. I'm not sure what the current method is for handing in homework but I'm sure Tamim will be able to help out with that

PR #21, opened 2018-06-11
adding homework js3-w1

amudar:
Hi,
Please review my work.
Thanks - Mudar

### DIFF
--- a/Week1/js3-hw-w1/app.js
+++ b/Week1/js3-hw-w1/app.js
@@ -1,0 +1,116 @@
+'user strict';
+
+
+const url = "https://api.github.com/orgs/HackYourFuture/repos?per_page=100";
+
+function fetchJSON(url, cb) {
+
+    const request = new XMLHttpRequest();
+
+    request.open("GET", url);
+    request.responseType = "json";
+
+    request.onreadystatechange = () => {
+        if (request.readyState === 4) {
+            if (request.status < 400) {
+                cb(null, request.response);
+            } else {
+                cb(new Error(request.statusText));
+            }
+        }
+    };
+
+    request.send();
+}
+
+
+function callback(error, data) {
+    if (error !== null) {
+    } else {
+        renderRepo(data);
+    }
+}
+fetchJSON(url, callback);
+
+function renderRepo(repo) {
+    const root = document.getElementById('root');
+    const listItem = createAndAppend('div', root);
+    listItem.id = 'listItem';
+    const listItemName = createAndAppend('p', listItem);
+    listItemName.innerHTML = 'HackYourFuture Repositories';
+    const select = createAndAppend('select', listItem);
+    const repoBox = createAndAppend('div', root);
+    repoBox.id = 'repo-box';
+    const contributorBox = createAndAppend('div', root);
+    contributorBox.id = 'contributors-box';
+    select.addEventListener("change", () => repoInfo(select.value));
+    repo.forEach(repo => {
+        const option = createAndAppend('option', select);
+        option.innerHTML = repo.name;
+        option.setAttribute('value', repo.name);
+    });
+    repoInfo(select.value);
+}
+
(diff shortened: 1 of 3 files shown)


Week1/js3-hw-w1/index.html
```diff
@@ -0,0 +1,18 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <meta http-equiv="X-UA-Compatible" content="ie=edge">
+    <link rel="stylesheet" href="Style.css" />
```
harditsingh: Somehow your CSS file was not linked, turns out that stylesheet is 'style.css' and not 'Style.css' (yep, the value for the source is case-sensitive)

Week1/js3-hw-w1/app.js
```diff
@@ -0,0 +1,116 @@
...
+        }
+    };
+
+    request.send();
+}
+
+
+function callback(error, data) {
+    if (error !== null) {
+    } else {
```
harditsingh: You didn't handle the error! If the server is unreachable, the page doesn't load at all. This isn't the best thing to do as the user would be wondering what happened. You should put text on the page saying that there was an error retrieving the data from the server.

Week1/js3-hw-w1/app.js
```diff
@@ -0,0 +1,116 @@
...
+    const select = createAndAppend('select', listItem);
+    const repoBox = createAndAppend('div', root);
+    repoBox.id = 'repo-box';
+    const contributorBox = createAndAppend('div', root);
+    contributorBox.id = 'contributors-box';
+    select.addEventListener("change", () => repoInfo(select.value));
+    repo.forEach(repo => {
+        const option = createAndAppend('option', select);
+        option.innerHTML = repo.name;
+        option.setAttribute('value', repo.name);
```
harditsingh: You forgot to sort the elements of the list in an alphabetical order!

Week1/js3-hw-w1/app.js
```diff
@@ -0,0 +1,116 @@
...
+
+function createAndAppend(tagName, parent) {
+    const element = document.createElement(tagName);
+    parent.appendChild(element);
+    return element;
+}
+
+function repoInfo(repoName) {
+    const urlContributor = 'https://api.github.com/repos/HackYourFuture/' + repoName + '/contributors';
+    const repoUrl = 'https://api.github.com/repos/HackYourFuture/' + repoName;
```
harditsingh: You should not put static links inside functions, you should save these strings in a variable and then get the value from the variable. The way you did it gets the work done, but it will cause problems when you try to change the link (and forget to change it on some places).

Week1/js3-hw-w1/app.js
```diff
@@ -0,0 +1,116 @@
...
+    const repoBox = document.getElementById('repo-box');
+    repoBox.innerHTML = '';
+    const p = createAndAppend('p', repoBox);
+    const repoName = createAndAppend('p', repoBox);
+    const forks = createAndAppend('p', repoBox);
+    const updated = createAndAppend('p', repoBox);
+    repoName.innerHTML = 'repo: &nbsp;&nbsp;&nbsp;' + repoInfo.name;
+    forks.innerHTML = 'Forks: &nbsp;&nbsp;&nbsp;' + repoInfo.forks_count;
+    updated.innerHTML = 'Updated: &nbsp;&nbsp;&nbsp;' + repoInfo.updated_at;
+    p.innerHTML = 'Description: &nbsp;&nbsp;&nbsp;' + repoInfo.description;
```
harditsingh: You shouldn't use spaces, tabs, etc. to format the text on the webpage, always use CSS for this. (Can you please correct this in the next assignment when you get time?)

PR #22, opened 2018-06-12
added JS3 homework week1

Dumie1:
Added javascript3 homework week1. Not really confident with my index.js file or code...

### DIFF
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -3,8 +3,7 @@
 {
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
-    xhr.open('GET', url);
-    xhr.responseType = 'json';
+    xhr.open('GET', url, true);
     xhr.onload = () => {
       if (xhr.status < 400) {
         cb(null, xhr.response);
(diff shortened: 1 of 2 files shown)


homework/src/index.js
```diff
@@ -16,6 +15,98 @@
     xhr.send();
   }
 
+  function repoData(error, data) {
+    if (error !== null) {
+      console.error(error);
```
harditsingh: You should not print the error to the console as the user would never know what went wrong. Instead, you should show the error on the webpage

homework/src/index.js
```diff
@@ -16,6 +15,98 @@
...
+        const optionItem = createAndAppend('option', select);
+        optionItem.innerText = obj.name;
+        optionItem.value = 'https://api.github.com/repos/HackYourFuture/' + obj.name;
+      }
+      select.addEventListener('change', request => {
+        const newUrl = event.target.value;
+        const root = document.getElementById('root');
+
+        root.innerHTML = null;
+        fetchJSON("https://api.github.com/orgs/HackYourFuture/repos?per_page=100", repoData);
```
harditsingh: You should not use URLs like this, you should put them in a variable instead. Without a variable, it can get confusing when you try to reuse the code or try to change something

homework/src/index.js
```diff
@@ -16,6 +15,98 @@
...
+      const heading = createAndAppend('h1', header);
+      heading.innerHTML = 'HYF Repositories :';
+      const select = createAndAppend('select', header);
+      select.setAttribute('class', 'select_menu');
+      const optionItem = createAndAppend('option', select);
+      optionItem.innerText = 'Choose a repository';
+      for (const obj of arrayOfObjects) {
+        const optionItem = createAndAppend('option', select);
+        optionItem.innerText = obj.name;
+        optionItem.value = 'https://api.github.com/repos/HackYourFuture/' + obj.name;
```
harditsingh: You forgot to sort the elements of the drop-down menu

PR #24, opened 2018-06-13
homework week1 added

amirno1:
I know my code looks like a chaos but everything works perfectly.  I started in this way and i had nothing to do but continue. I really appriciate it if you can tell me how and where in my code, could i make it easier .Last thing to say ,the css part is really simple because it has been done in no time. thanks for your feedback in advance.

### DIFF
--- a/Week1/app.js
+++ b/Week1/app.js
@@ -1,0 +1,205 @@
+'use strict';
+function main() {
+  //making the main div
+  let mainDiv = document.createElement('div');
+  document.body.appendChild(mainDiv);
+
+  //giving an id to main div
+  let mainDivAttribute = document.createAttribute('id');
+  mainDivAttribute.value = 'root';
+  mainDiv.setAttributeNode(mainDivAttribute);
+
+  function createHTMLElement(tag, parent) {
+    let newElement = document.createElement(tag);
+    if (parent) {
+      parent.appendChild(newElement);
+    }
+    else {
+      document.body.appendChild(newElement);
+    }
+    return newElement;
+  }
+
+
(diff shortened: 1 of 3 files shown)


Week1/app.js
```diff
@@ -0,0 +1,205 @@
...
+  contributorsParagraph.innerText = 'Contributions';
+  let contributorsList = createHTMLElement('ul', contributorsInformationBox);
+
+
+
+
+  function renderContributorsInformationToDom() {
+
+    let hyfRepositoryAtTheMoment = repositoryNameLink.innerText;
+    const gitHubUrl = 'https://api.github.com/repos/HackYourFuture/' + hyfRepositoryAtTheMoment + '/contributors';
```
harditsingh: You should not put the URL inside functions and in multiple places, it can get confusing really quickly

Week1/app.js
```diff
@@ -0,0 +1,205 @@
...
+  repositoryInformationBox.setAttribute('id', 'rep_div');
+
+
+  //Requesting HYF repositories
+  const user = 'HackYourFuture';
+  const gitHubUrl = 'https://api.github.com/users/' + user + '/repos?per_page=100';
+  const hyfRepositoriesRequest = new XMLHttpRequest();
+  hyfRepositoriesRequest.open('GET', gitHubUrl, true);
+  hyfRepositoriesRequest.send();
+  hyfRepositoriesRequest.onreadystatechange = repositoriesRequest;
```
harditsingh: You have not handled any errors if server is not available. This way the user doesn't know why the website is not working when a server is not available.

PR #26, opened 2018-06-16
home work week1

zmagzoub:
I am really sorry because I handed in my homework too late. Today's evening I will hand in week 2 homework. 

My page is not styled, but the code works well. I will style my page today and hand in week 2 homework styled. Sorry again.


homework/src/index.js
```diff
@@ -1,47 +1,150 @@
...
+            err.innerHTML = error.message;
+        } else {
+            renderRepositories(null, data);
+        }
+    }
+
+    function renderRepositories(error, data) {
+        if (error !== null) {
+            const err = document.getElementById('root');
+            err.innerHTML = error.message;
```
harditsingh: You are handling errors in two places, should handle only at one place to avoid complexity

homework/src/index.js
```diff
@@ -1,47 +1,150 @@
...
+
+            select.addEventListener('change', (event) => {
+                const newURL = 'https://api.github.com/repos/HackYourFuture/' + event.target.value;
+                fetchJSON(newURL, renderInformation);
+            });
+
+            data.forEach(repository => {
+                const listItem = createAndAppend('option', select);
+                listItem.innerHTML = repository.name;
+                listItem.setAttribute('value', repository.name);
```
harditsingh: The options in the drop-down menu are not sorted

homework/src/index.js
```diff
@@ -1,47 +1,150 @@
...
+            information.setAttribute('id', 'information');
+
+            const contributors = createAndAppend('div', container);
+            contributors.setAttribute('id', 'contributors');
+
+            const contributorsList = createAndAppend('ul', contributors);
+            contributorsList.setAttribute('id', "contributors-list");
+
+            select.addEventListener('change', (event) => {
+                const newURL = 'https://api.github.com/repos/HackYourFuture/' + event.target.value;
```
harditsingh: You should avoid using URLs inside functions, it can get hard to reuse the code without debugging (especially in bigger projects)

PR #28, opened 2018-06-16
Homework week2

ThreePalmTrees:


ThreePalmTrees:
Thanks for taking the time Hardit,
For sorting the dropdown aka the select tag.. this is the closest I got to it:
```
sortedSelect = selectTag.sort((opt1, opt2) => opt1.localeCompare(opt2)); // but the selectTag is an object and sort() works only on arrays, so I'm not sure how to handle that task
```

PR #37, opened 2018-06-21
Week3

Meazer:
I didn't do ARIA BONUS, I did the Alternative BONUS : [Code Kata Race](https://www.codewars.com/users/Meazer/completed_solutions) instead.
Thanks in advance.


homework/index.js
```diff
@@ -0,0 +1,144 @@
+'use strict';
+
+function main() {
+
+    const mainUrl = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+
+    // initial page: create header and main container.
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root);
```
remarcmij: I like that you have used semantic HTML5 elements, such as `<header>`, `<main>` and `<section>` instead of just `<div>`'s.

homework/index.js
```diff
@@ -0,0 +1,144 @@
...
+    createAndAppend('h3', header, { text: 'HYF Repositories' });
+    const mainContainer = createAndAppend('main', root, { class: 'main-container' });
+
+    fetchAndRender(mainUrl);
+
+    async function fetchAndRender(url) {
+
```
remarcmij: I would personally not define all of the other functions inside the `main` function. While this helps to keeps those function out of the global scope, I would prefer to use a block scope for that. That would then look like this:

```js
'use strict';

{
  let mainContainer;

  function main() {
    const mainUrl = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';

    // initial page: create header and main container.
    const header = createAndAppend('header', root);
    createAndAppend('h3', header, { text: 'HYF Repositories' });
    mainContainer = createAndAppend('main', root, { class: 'main-container' });

    fetchAndRender(mainUrl);
  }

  async function fetchAndRender(url) {
    //...
  }

  // all your other functions

  window.addEventListener('load', main);
}
```

This has the same effect, but now I do not have to hunt around going over your large main function to see where you define functions and where you might have some hidden code that is executed directly. In your case there is no such hidden code, but as a reviewer I had to check it anyway. (It is good to keep an audience in mind when you write code).

homework/index.js
```diff
@@ -0,0 +1,144 @@
...
+                }
+            };
+            //if XHR not loaded reject the promise
+            XHR.onerror = () => reject(new Error('Network request failed'));
+            XHR.send();
+        });
+
+    }
+
+    async function renderReposList(reposObj) {
```
remarcmij: You named the parameter `reposObj`, implying that you expect an object, but in fact you get an array. So the name `repos` (as plural of `repo`) would be a better name.

homework/index.js
```diff
@@ -0,0 +1,144 @@
...
+        reposObj.sort((a, b) => { return a.name.localeCompare(b.name); });
+        //create list and append it to header
+        const selectList = createAndAppend('select', header, { id: 'selectList' });
+
+        for (const rep in reposObj) {
+            // add an option for each repo with value that is his index at reposObject 
+            createAndAppend('option', selectList, { value: rep, text: reposObj[rep].name });
+
+        }
+        //add a listener when select a new option
```
remarcmij: You could then replace your `for` loop as follows, making it more explicit that you are working on an array (using `for...in` on an array is a bad practice):

```js
repos.forEach((repo, index) => {
  // add an option for each repo with value that is his index at reposObject 
  createAndAppend('option', selectList, { value: index, text: repo.name });
});
```

homework/index.js
```diff
@@ -0,0 +1,144 @@
...
+                renderContributions(contributors);
+            } catch (err) {
+                renderError(err);
+            }
+
+        };
+        return reposObj[selectList.value];
+
+    }
+
```
remarcmij: I think you could further improve your code by reorganizing some functions and introduce a new function to render a given repo (remove blank lines and comments for brevity):

```js
async function fetchAndRender(url) {
  try {
    const data = await fetchJSON(url);
    renderReposList(data);
  } catch (err) {
    renderError(err);
  }
}

async function renderReposList(repos) {
  const header = document.querySelector('header');
  repos.sort((a, b) => { return a.name.localeCompare(b.name); });
  const selectList = createAndAppend('select', header, { id: 'selectList' });
  repos.forEach((repo, index) => {
    createAndAppend('option', selectList, { value: index, text: repo.name });
  });
  selectList.onchange = async () => renderRepo(repos[selectList.value]);
  // show first repo at startup
  renderRepo(repos[selectList.value]);
}

async function renderRepo(repo) {
  try {
    const contributors = await fetchJSON(repo.contributors_url);
    renderInfo(repo);
    renderContributions(contributors);
  } catch (err) {
    renderError(err);
  }
}
```

homework/index.js
```diff
@@ -0,0 +1,144 @@
...
+            { href: rep.html_url, target: '_blank', text: rep.name });
+
+    }
+
+    function renderError(error) {
+        mainContainer.innerText = '';
+        createAndAppend('div', mainContainer, { text: error.message, class: 'error' });
+    }
+
+    function renderContributions(contributions) {
```
remarcmij: Your function `renderContributions` might perhaps better be called `renderContributors`, to avoid confusion with the `contributions` property, which is just a number. Similarly, the parameter name could be `contributors`. Note that this parameter receives an array, so the line `if (contributions)` serves no purpose. An empty array is also `truthy`, so the condition is always `true`. So you can leave it out as the `for` will be executed zero times if the array is empty.

Meazer: Actually here I'm trying to cover the case when the repo is empty as [hyfer-infra](https://github.com/HackYourFuture/hyfer-infra) so the contributors_url return `null`.
I corrected the wrong named function and parameter.

Meazer: Actually here I'm trying to cover the case when the repo is empty as hyfer-infra so the contributors_url return null.
I corrected the wrong named function and parameter.

Meazer:
Hi Jim,
thank you for your feedback, I will try to implement it.

homework/index.js
remarcmij: You are completely right. My own app shows this for that repo:

hyfer-infra

PR #49, opened 2018-06-24
Homework-Part2  OOP-Classes Week3

PKnar:
The link to the website -> https://pknar.github.io/JavaScript3/Week3/


Week3/Contributor.js
```diff
@@ -0,0 +1,41 @@
...
+  render(parent) {
+    parent.innerHTML = '';
+
+    try {
+      for (let prop in this.data) {
+        const liContribInfo = Util.createHTMLElement('li', parent);
+        let [...contribDetails] = liContribInfo.children;
+        contribDetails[0] = Util.createHTMLElement('a', liContribInfo);
+        const contributorsAvatar = Util.createHTMLElement('img', contribDetails[0]);
+        contribDetails[1] = Util.createHTMLElement('li', liContribInfo, 'Username: '.bold() + this.data[prop].login);
```
remarcmij: Don't use this `.bold()` string method. It is deprecated: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/bold

Week3/Repository.js
```diff
@@ -0,0 +1,41 @@
...
+    parent.innerHTML = '';
+
+    const liRepInfo = Util.createHTMLElement('li', parent);
+    let [...repositoryDetails] = liRepInfo.children;
+    repositoryDetails[0] = Util.createHTMLElement('i', liRepInfo);
+    repositoryDetails[1] = Util.createHTMLElement('li', liRepInfo, 'Repository: '.bold());
+    const a = Util.createHTMLElement('a', repositoryDetails[1], this.data.name);
+
+    if (this.data.description !== null) {
+      repositoryDetails[2] = Util.createHTMLElement('li', liRepInfo, 'Description: '.bold() + this.data.description);
```
remarcmij: Don't use `.bold()`. It is deprecated.

PKnar:
Hello @remarcmij , thank you very much for your constructive feedback!! I modified the files based on your feedback. :+1: Between I have few questions, I will ask those under your comments.

Week3/Contributor.js
PKnar: I personally did not want to use those too, but unfortunately I could not find any useful information on how to make the first word bold, CSS does not have such property and the majority of Javascript solutions are either not clear enough or are presented using jQuery. Do you maybe have any suggestions?

Week3/App.js
```diff
@@ -0,0 +1,72 @@
...
+    });
+    this.renderData(repos[0]);
+  }
+
+  async renderData(repositoryData) {
+
+    const repository = new Repository(repositoryData);
+    repository.render(ulRepInfo);
+    const contribData = await repository.fetchContributors();
+    const contributors = new Contributor(contribData);
```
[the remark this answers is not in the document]
remarcmij: You could do it like this (check the zip file that I posted):

```js
contributors
  .map(contributor => new Contributor(contributor))
    .forEach(contributor => contributor.render(contributorList));
```

In general, a singular or plural form of a name is to be understand to represent a single item or collections of items respectively.

Week3/Contributor.js
remarcmij: You can either embed the word to made bold with  the standard element `<strong>`  which is by default styled to be bold or you can use a `<span>` with a class that you style yourself and set the `font-weight: bold;`. As your `createHTMLElement()` function uses the third parameter to set  the `innerHTML` of the element, you could simple do something like this:

```
contribDetails[1] = Util.createHTMLElement('li', liContribInfo, `<strong>Username:</strong> ${contributor.login}`);
```

I haven't tried it but I think it should work.

Week3/App.js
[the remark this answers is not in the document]
PKnar: Ah, I see! I will definitely check out your version. We are currently busy with Node.js. :sweat_smile:  Once I have time I will check it out, I am sure I will learn a lot from that. I am sorry I forgot to ask this last question. In week1 and week2 I used forEach() loop for contributors info, however, it did not work once classes were added. Is there a specific reason for that?

Week3/Contributor.js
PKnar: Yes this does work :+1:  interesting , before using bold() I did this  '<strong> Username: </strong>' + contributor.login, and  it did not work. I guess it has something to do with the function I used. Thanks for this solution!

PR #52, opened 2018-08-08
Week1

mehmetscoskun:


homework/src/index.js
```diff
@@ -32,13 +32,106 @@
...
       } else {
-        createAndAppend('pre', root, { html: JSON.stringify(data, null, 2) });
+
+        data.sort((a, b) => a.name.localeCompare(b.name));
+
+        data.forEach((obj) => { createAndAppend('option', select, { html: obj['name'], class: 'rep-names' }); });
+
+
+        //info-box of the page
+        function showDetails() {
```
remarcmij: You should not normally define functions inside other functions, except when you dealing with relative short functions that you pass as a callback or event handler to some other function. 

You can define `showDetails()` at the file level, but you will need to pass the repository information that it needs to show as parameter to that function, e.g.:

```js
function showDetails(selectedRepo) {
  //...
}
```

homework/src/index.js
```diff
@@ -32,13 +32,106 @@
...
+          myTable.remove();
+
+          //and then show the details of newly selected repo
+          showDetails();
+        });
       }
     });
+
+
   }
```
remarcmij: So, the overall structure of your file could be:

```js
'use strict';
{
  function fetchJSON(...) {
    //...
  }

  function createAndAppend(...) {
    //...
  }

  function main() {
    //...
   showDetails(repositories[0]);
   //...
   select.addEventListener('change', function () {
      //...
      showDetails(repositories[select.value]);
    });
  }
 
  function showDetails(selectedRepo) {
    //...
  }

  function RepositoryDetails(repo) {
    //...
  }

  window.onload = () => main(HYF_REPOS_URL);
}
```

mehmetsaidcoskun:
thank you @remarcmij , your comments show me my mistakes and weak points I must improve. Besides, @JawharBairakdar explained a few points I could not understand well. I make some changes according to your comments and the short tutorial given by Jawhar :

first of all, I deleted 'getElementbyId' lines and combined the newly defined 'const' variables with the 'createAndAppend' functions.

secondly, I tried to give better and more explanatory names to the variables and the functions. for example, I used 'data' as a parameter of functions and 'repositories' and 'repository' as an argument.

thirdly and the most important, I took the functions in the 'main' function outside of the 'main' function. I defined three functions outside of the 'main' function and called them back inside of the 'main' function. I did not create a new file for the new functions. I hope what I did is true. 

fourthly I create 'contributions' box. And then I made some changes in 'style.css' file. My page is now responsive but I added only one media query. I suppose it works well.

Thanks to you and Jawhar my codes look more legible and understandable :)

See you next Sunday

PR #53, opened 2018-08-09
Update MAKEME.md

gijscor:
I have defined in the introduction a general concept of what the student will be doing in this assignment, so that the student can understand a bit more easily on what he/she will be doing for the homework from a more meta perspective. It took me quite some time to understand what exactly the student would be doing based on the quite technical detailed assignment guidelines.


JawharBairakdar:
@gijscor Thanks for your PR.
I see it has a very valuable addition that we are introducing what the student going to do during this assignment @remarcmij what do you say?

PR #56, opened 2018-08-11
homework week1 is done

genegr88:
homework the first part is finished

### DIFF
--- a/Week1/github.js
+++ b/Week1/github.js
@@ -1,0 +1,50 @@
+'use strict';
+{
+    const URL_Rep = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+    function renderRepository(container, repository) {
+        let someDate = new Date(repository.updated_at);
+        container.innerHTML = "";
+        const innerTable = createAndAppend("table", container);
+        const tBody = createAndAppend("tbody", innerTable);
+        const tr = createAndAppend('tr', tBody);
+        createAndAppend('td', tr, {text: "Repository:", class: "nameRow"});
+        const tdForLink = createAndAppend('td', tr,);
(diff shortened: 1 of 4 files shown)


Week1/github.js
```diff
@@ -0,0 +1,50 @@
+'use strict';
+{
+    const URL_Rep = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+    function renderRepository(container, repository) {
+        let someDate = new Date(repository.updated_at);
```
JawharBairakdar: `someDate` is a meaningful name, for you :smile_cat:, I found this thing actually that we are give a very meaningful names for us but when somebody else reads our code will face an issue understanding what is referring to.

So `someDate` of course it's some date but it would be more clear to call it `lastUpdate` or `latestCommit` or you may have a better name for it.

so the point is this variable handling a date for an event accord in the past, you can call it as the event called simply :slightly_smiling_face: .

Week1/github.js
```diff
@@ -0,0 +1,50 @@
...
+        container.innerHTML = "";
+        const innerTable = createAndAppend("table", container);
+        const tBody = createAndAppend("tbody", innerTable);
+        const tr = createAndAppend('tr', tBody);
+        createAndAppend('td', tr, {text: "Repository:", class: "nameRow"});
+        const tdForLink = createAndAppend('td', tr,);
```
JawharBairakdar: it's not a valid syntax to have an empty field after the last comma of your function call, nothing will happen but it may have some confusion, so if I came after a year or so, for the first look here I will say something was here and removed.

and the name you can Just call it `repositoyLink` in case you want to refer that it's a `<td>` tag you can just call it `td_repositoyLink` or as always you may have a better one :100:

PR #57, opened 2018-08-11
week1 finished

end0620:


Week1/assets/index.js
```diff
@@ -0,0 +1,121 @@
...
+      createAndAppend('h3', item, {html: element +': '});
+      
+      const item1 = createAndAppend('div', repoContent, {class:'item'});
+     
+      if(element === 'Repository')
+        createAndAppend('a', item1, {html: repo[repoValues[index]], href: repo.url, target: '_blank'});
+      else if(element === 'Updated')
+      {
+        let myDate = new Date(repo[repoValues[index]]);
+        createAndAppend('h4', item1, {html: myDate.toLocaleString()});
```
JawharBairakdar: Here you have two main points I am referring to:
  1. `item` && `item1` is not a very described names so you can think about what they contain and what they exactly doing, read it as story :slightly_smiling_face:.
  2. Here is a very good example of what make people start confusing !!
   - let's just read it:
      - you have `let myDate // you have to choose a better name than myThing`
      - Which you are assigning to it the returning value of `new Date();`
      - Which you you are taking it from `repo` that t's parent object of `repoValues[index]`

PR #58, opened 2018-08-11
created_first_version_homework_week1

WardMad:


homework/src/index.js
```diff
@@ -1,47 +1,91 @@
...
+            if (key === 'text') {
+                elem.innerHTML = value;
+            } else {
+                elem.setAttribute(key, value);
+            }
+        });
+        return elem;
+    }
+
+    function main(url) {
```
remarcmij: If you look at the HTML your application generates for the `table` in the Chrome inspector you can see that it has issues:

```html
<table class="table">
  <tbody>
    <tr></tr>
    <td class="label">Repository: </td><a href="https://github.com/HackYourFuture/DataStructures" target="_blank">DataStructures</a><tr>Description:
      Algorithms and Data structures course</tr>
    <tr class="label">Forks: 12</tr>
    <tr class="label">Updated: 16/04/2018, 14:43:20</tr>
    <td></td>
  </tbody>
</table>
``` 

I'll add comments at the detail level where this goes wrong.

homework/src/index.js
```diff
@@ -1,47 +1,91 @@
...
+
+
+        const root = document.getElementById('root');
+
+        const div = createAndAppend('div', root, { class: 'container' });
+        const header = createAndAppend('h3', div, { text: "HYF Repositories  ", class: 'header' });
+        const select = createAndAppend('select', header, { class: "select" });
+        const table = createAndAppend('table', root, { class: 'table' });
+        const tbody = createAndAppend('tbody', table);
+        const tr = createAndAppend('tr', tbody);
```
remarcmij: In my VSCode the variable `tr` is flagged as unused. This should always raise a red flag in your mind.

homework/src/index.js
```diff
@@ -1,47 +1,91 @@
...
+        createAndAppend('td', tbody, { text: "Repository: ", class: 'label' });
```
remarcmij: You can see in the HTML that the parent element for the `td` element should be the `tr` element that you created in the previous line and not the `tbody` element. That's why in the generated HTML your `tr` remains empty.

homework/src/index.js
```diff
@@ -1,47 +1,91 @@
...
+        const repoTitle = createAndAppend('a', tbody);
```
remarcmij: The `a` element should be a child of the `td` element, not of the `tbody` element. For applications like this it is always a good idea to use the Chrome inspector to look at the HTML your app generated. Because I'm sure that if you were to write this HTML by hand rather than through JS you would do it correctly.

homework/src/index.js
```diff
@@ -1,47 +1,91 @@
...
+                data.forEach((item, index) => {
+                    const option = createAndAppend('option', select, { text: item.name, value: index });
+                });
+
+
+                select.addEventListener('change', (event) => {
+
+                    const index = event.target.value;
+
+                    repoTitle.innerText = select[index].textContent;
```
remarcmij: `repoTitle` and `forksTitle` refer to `tr` elements. These elements should only contain `td` elements (or `th` elements in case of a header). Inside the `td` element you can define the actual content, such as text nodes and for instance `a` elements.

PR #59, opened 2018-08-16
Week1 Homework

acimanx:
*Still need to figure out how to optimize a work of functions and fetchers with template strings
*Design not finished yet
*Tired to work with synchronous requests, going to upgrade to async

### DIFF
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -1,47 +1,77 @@
-'use strict';
 
+"use strict"
 {
+  (function App() {
+    document.getElementById("root").innerHTML = `
+    <header><div class=selectfield><select id=selectmenu></select></div></header>
+    <main>
+    <section class=left>
+    <div class=scy id=info-box>
+    <div class=description>"No description"</div>
+    <div class=forks>"No forks"</div>
+    <div class=updated>"No data"</div>
+    </div>
+    <div id=size style=width:0></div>
+    </section>
+    <section class=right>
+    <div id=contributors><div class=card>
+    <a href="#">
+    <figure><img src="https://avatars1.githubusercontent.com/u/39067270?s=400&u=b8a97121f8e8623492a3fd16de13cfa780409246&v=4"></figure>
+    </a>
+    <h2>Acimanx</h2>
+    <p>contributions:53</p><div class=repos-count>123</div>
+    </div>
+    </div>
+    </section>
+    </main>`
+  })()
+
+  document.addEventListener("DOMContentLoaded", function (event) {
+
+    const url = "https://api.github.com/orgs/HackYourFuture/repos?per_page=10"
+
+    fetchJSON(url, function (err, data) {
+      if (err) {
+        document.getElementById("main").innerHTML = err
+      }
+      else {
+        document.getElementById("selectmenu").innerHTML = `<option value=""disabled id=first selected>Choose your option</option>${data.map((x, b) => `<option value=${b} class=select>${x.name}</option>`).join("")}`
+        document.getElementById("selectmenu").addEventListener("change", event => {
+          document.getElementById("info-box").innerHTML = `<div class=description>${null == data[event.target.value].description ? "No description" : data[event.target.value].description}</div><div class=forks>${null == data[event.target.value].forks ? "No forks" : data[event.target.value].forks}</div><div class=updated>${null == data[event.target.value].updated_at ? "No data" : data[event.target.value].updated_at}</div>`
+          let repositories
+          let repoData
+          fetchJSON(data[event.target.value].contributors_url, (err, contributorData) => {
+            repositories = contributorData
+            repoData = repositories.forEach(({ login }) => {
+              let usersData
+              fetchJSON(`https://api.github.com/users/${login}`, (err, fetchUsersData) => {
+                usersData = fetchUsersData.name
+                console.log(usersData)
+              })
+            })
+            console.log(repoData)
+            outputPageContent(repoData)
+          })
+          function outputPageContent() {
+            document.getElementById("contributors").innerHTML = repositories.map(a => `<div class=card><a href=${a.html_url} target=_blank><figure><img src=${a.avatar_url}></figure></a><h2>${a.login}</h2><p id=reposdata>${console.log(repoData)}<p>contributions: ${a.contributions}<div class=repos-count></div></div>`).join("")
+          }
+        }, !1)
+      }
+    })
+  })
+
   function fetchJSON(url, cb) {
-    const xhr = new XMLHttpRequest();
-    xhr.open('GET', url);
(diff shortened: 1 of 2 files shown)


homework/src/index.js
```diff
@@ -1,47 +1,119 @@
...
+        });
+        sectLinfobox = crnap("div", sectLinfobox, {
+          html: null == changeEvent.size ? "No size" : `Size: ${changeEvent.size}`,
+          class: "size"
+        });
+        fetchJSON(changeEvent.contributors_url, (err, contributors) => {
+          let fetchToDo = contributors.length;
+          contributors.forEach(contributor => {
+            fetchJSON(contributor.url, (err, user) => {
+              Object.assign(contributor, user);
```
remarcmij: This inner function is best defined as separate named function at file level. You should strive for your functions to be relatively short, say not longer than can fit on a computer screen.

homework/src/index.js
```diff
@@ -1,47 +1,119 @@
...
-    const xhr = new XMLHttpRequest();
-    xhr.open('GET', url);
-    xhr.responseType = 'json';
-    xhr.onload = () => {
-      if (xhr.status < 400) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
+  function main() {
+    fetchJSON("https://api.github.com/orgs/HackYourFuture/repos", (err, data) => {
```
remarcmij: I'm seeing a subset of HYF repositories only. For instance, the JavaScript2 and JavaScript3 repos are not listed in your `select` element. This is because you left out the `perPage` query parameter that was explicitly mentioned in the assignment:

```
https://api.github.com/orgs/HackYourFuture/repos?per_page=100
```

By leaving it out, GitHub just returns the first 30 repos.

homework/src/index.js
```diff
@@ -1,47 +1,119 @@
...
+        });
       }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+      const root = document.getElementById("root");
+      let header = crnap("header", root, {});
+      header = crnap("div", header, { class: "selectfield" });
+      const menu = crnap("select", header, { id: "selectmenu" });
+      var main = crnap("main", root, { id: "main" });
```
remarcmij: Don't reuse the name `main`. You have already used it for the function `main()`.

Also, don't use `var`. Use `let` and `const` instead.

homework/src/index.js
```diff
@@ -1,47 +1,119 @@
...
+      err
+        ? crnap("div", root, { a: err.message, b: "alert-error" })
+        : (data
+          .sort((b, a) => b.name.localeCompare(a.name))
```
remarcmij: While you can use the ternary operator like this, it is not very readable code. An `if-then-else` would more clearly express your intention:

```js
if (err) {
  //...
} else {
  //..
}

Additionally, your error `div` is not rendered visually. It falls off the screen because your `<main>` element has a 100% height. Did you check the error handling yourself?

File: homework/src/index.js [hunk 5]
```diff
@@ -1,47 +1,119 @@
...
+          .forEach((a, d) =>
+            crnap("option", menu, { value: d, b: "select", html: a.name })
+          ),
+          menu.addEventListener("change", event => {
+            event = event.target.value;
```
remarcmij: Don't reuse a parameter like this. The name `event` cannot (logically) be an event and a value at the same time. There is nothing to gained from this reuse, only confusion.

homework/src/index.js
```diff
@@ -1,47 +1,119 @@
...
+            rend(data[event]);
+          }));
+    });
   }
-
-  function createAndAppend(name, parent, options = {}) {
+  function crnap(name, prnt, opts) {
```
remarcmij: `crnap` is crappy name 😉. Also, separate function definitions with a blank line.

homework/src/index.js
```diff
@@ -1,47 +1,119 @@
...
-    xhr.open('GET', url);
-    xhr.responseType = 'json';
-    xhr.onload = () => {
-      if (xhr.status < 400) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
+  function main() {
+    fetchJSON("https://api.github.com/orgs/HackYourFuture/repos", (err, data) => {
+      function rend(changeEvent) {
```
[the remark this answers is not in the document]
acimanx: >A general comment: you have used cryptic variable and function names, such as rend which, I take it, stand for render. 
A general reply: Yes, it is render.
>Using cryptic names like that makes it hard for other people to understand your code (the JavaScript doesn't mind how you name things). You should write your code that best expresses your intentions in the clearest possible manner. 
In my opinion using large names can make a trouble for reading a code, code must be a poem, but a poem of code - is poem of logic, not the words.
>With moderns IDEs such VSCode there is really no benefit in typing effort to be gained from short names. Auto-complete will complete the name for you is you type the first few letters.
There is: large words creating a noise for an observer.
>You should as much as possible define functions at file level, rather than inside other functions. 
Before this comment, and our last lecture I never heard about it, and never got an explanation why. If its not difficult for you please explain.
>In the case of the rend() function I have a hard time to find out where it ends
Sure, I did this intentionally, because in the requirements wasn't written to make extra spaces for a better understanding.

homework/src/index.js
```diff
@@ -1,47 +1,119 @@
...
+        main.innerHTML = "";
+        let sectL = crnap("section", main, { class: "left", id: "maininfo" });
+        const sectR = crnap("section", main, { class: "right", id: "contribInfo" });
+        let sectLinfobox = crnap("div", sectL, { id: "info-box", class: "scy" });
+        sectL = crnap("div", sectLinfobox, {
```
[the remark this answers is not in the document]
acimanx: sectR means section right

>(whatever it means)

You can understand it by "section class right" described inside of this call. (it's 1 extra sentence to read). I think there is no point to name it sectionRight or rightSection or similar to it, because it's just 1 line of code.

homework/src/index.js
```diff
@@ -1,47 +1,119 @@
...
+          html:
+            null == changeEvent.full_name
```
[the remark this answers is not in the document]
acimanx: >but you will not find this style in the JS world.
Probably it's a good critics, but pity that you didn't pointed a better/right way to do that. I've never heard a ways of designing this stuff on our lectures.
So this part of the code works, program is working, money is making, client satisfied, work is done.

homework/src/index.js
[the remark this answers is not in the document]
remarcmij: Hi Andrew, you are entitled to your own opinions. If you don't agree with my advice, so be it.

PR #63, opened 2018-08-19
Week2

shadoo77:


homework/src/index.js
```diff
@@ -31,17 +33,124 @@
...
+      createAndAppend('div', root, { html: error.message, class: 'alert-error' });
+    });
+  }
+
+
+
+
+
+  // function who creates a left content section
+  function contentLeft(container, data, select) {
```
remarcmij: I would rename `data` to `repositories`. In general, use generic, meaningless names such as `data`, `info` etc only when the variable or parameter can contain anything. When you know, as in this case, exactly what the function parameter will receive, choose a names that accurately describes it contents.

homework/src/index.js
```diff
@@ -31,17 +33,124 @@
...
+  // line three
+  const tr3 = createAndAppend('tr', lTable);
+  createAndAppend('td', tr3, {html: 'Forks :', class: 'lable'});
+  const td3 = createAndAppend('td', tr3);
+  createAndAppend('p', td3, {html: repository.forks});
+  // line four
+  const tr4 = createAndAppend('tr', lTable);
+  createAndAppend('td', tr4, {html: 'Updated :', class: 'lable'});
+  const td4 = createAndAppend('td', tr4);
+  const defaultElDate = repository.updated_at;
```
remarcmij: This is a comment from week 1:

Since you are repeating code here (create a `tr` with two `td`s as children, you could create a separate function to avoid repetition (the DRY principle: Don't Repeat Yourself), for instance:

```js
  function (table, label, value) {
    createAndAppend('td', tr, { html: `${label} :`, class: 'label' });
    createAndAppend('td', tr, { html: value });
  }
```

homework/src/index.js
```diff
@@ -31,17 +33,124 @@
...
+  const defaultDate = new Date(defaultElDate);
+  const endDate = defaultDate.toLocaleString();
+  createAndAppend('p', td4, {html: endDate});
+}
+
+
+
+  // function who creates a right content section
+  function contentRight(container, data, select) {
```
remarcmij: `data` -> `contributors`

homework/src/index.js
```diff
@@ -31,17 +33,124 @@
...
+    const rightSection = createAndAppend('div', container, { id: 'rightSection', class: 'whiteframe' });
+    createAndAppend('h5', rightSection, {html: "Contributions :", class: 'contributionsLable'});
+    const contributorsURL = data[0].contributors_url;
+    const ContributorsPromise = fetchJSON(contributorsURL);
```
remarcmij: Variable, parameter and function names should be camelCase. Only constructor functions and ES6 classnames should be PascalCase. See the naming convention fundamental.

homework/src/index.js
```diff
@@ -31,17 +33,124 @@
...
+    ContributorsPromise
+      .then(contributors => {
+        showContributors(rightSection, contributors);
+      })
+      .catch(error => {
+        createAndAppend('div', rightSection, { html: error.message, class: 'alert-error' });
+      });
+
+    select.addEventListener('change', (event) => {
```
remarcmij: You have added a second event listener here.  A single event listener that handles both the repository data (left-hand) and contributor data (right-hand) is generally preferred. This simplifies understanding cause and effect in the app, especially for other developers looking at your code. They may not expect a second handler for the same event.

homework/src/index.js
```diff
@@ -31,17 +33,124 @@
...
   }
 
+
+
+
+
+function showContributors(rightSection, contributors) {
+  contributors.forEach(contributor => {
+    const contrSection = createAndAppend('section', rightSection);
+    const contrMainLink = createAndAppend('a', contrSection, {href:  contributor.html_url});
```
remarcmij: In terms of naming, all these `contr` prefixes are unnecessary, as in the context of this function everything evolves around contributors.

PR #66, opened 2018-08-25
JS3 week1

Donal86:


homework/src/index.js
```diff
@@ -31,17 +32,75 @@
   }
 
   function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
+    const root = document.getElementById("root");
+      const header = createAndAppend("header", root, { "class": "header" });
```
remarcmij: You do not need to enclose the property names such as `class` in quotes. If a property name of an object conforms to the rules of a JavaScript variables name (e.g. no spaces or special characters in the name) you can leave out the quotes. Of course, for the values you do need quotes if they are strings.

Donal86: Hi Jim, thank you so much for your feedback, I have considered your advise in the homework for week 2.

PR #68, opened 2018-08-26
Week2 homeWork

ozgurleo:


Week2/app1.js
```diff
@@ -0,0 +1,137 @@
...
+        createAppend('td', tr2, { text:"Fork: ", class:"tableData" });
+        createAppend('td',tr2, {text: JSON.stringify(repository.forks_count)});
+        
+        createAppend('td', tr3, { text:"Updated: ", class:"tableData" });
+        let updateDate = new Date (repository.updated_at);
+        createAppend('td',tr3, {text: updateDate.toLocaleString()});
+
+        
+
+    }
```
remarcmij: The preceding three blank lines are meaningless. Use blank lines conciously to separate blocks of related code like you would use blanks lines to separate paragraphs in a piece of written text. Do not use more than one blank linen at a time. Use a single blank lines between function definitions. Do no use randomly insert blank lines.

Week2/app1.js
```diff
@@ -0,0 +1,137 @@
...
+        const tr3 = createAppend('tr',tBody1);
+
+        createAppend('td', tr, { text:"Repository: ", class:"tableData" });
+
+        const nameTag= createAppend('a',tr, {text: JSON.stringify(repository.name)});
+        nameTag.setAttribute('href', repository.html_url);
+        nameTag.setAttribute('target', '_blank');
+
+        createAppend('td', tr1, { text:"Description: ", class:"tableData" });
+        createAppend('td',tr1, {text: JSON.stringify(repository.description)});
```
remarcmij: Do not use `JSON.stringify` here. There is absolutely no need for it and it causes all strings to be enclosed within quotes which you subsequently remove again in your implementation of `createAndAppend()`.

PR #69, opened 2018-08-28
Js3week2

Donal86:


homework/src/index.js
```diff
@@ -1,26 +1,29 @@
...
+      xhr.send();
+    });
+    }
+    
   function createAndAppend(name, parent, options = {}) {
     const elem = document.createElement(name);
     parent.appendChild(elem);
     Object.keys(options).forEach((key) => {
       const value = options[key];
+      //console.log(key);
```
remarcmij: Remove all commented-out code before submitting a PR.

homework/src/index.js
```diff
@@ -31,17 +34,75 @@
...
+    const root = document.getElementById("root");
+      const header = createAndAppend("header", root, { class: "header" });
+      createAndAppend("p", header, { html: "HYF Repositories" });
+      const menu = createAndAppend("select", header, { class: "select", id: "menu" });
+      const container = createAndAppend("div", root, { class: "container" });
+      const leftDiv = createAndAppend("div", container, { class: "left_div" });
+      const rightDiv = createAndAppend("div", container, { class: "right_div" });
+
+      fetchJSON(url)
+      .catch(err => {
```
remarcmij: In general the `.catch()` method should come at the end of the chain, not at the start. I explained in class that in an exceptional case you might want to have it earlier if you want to take corrective action after some error. But most of the time that is not the case and `.catch()` should come at the end.

When I force an error by fiddling the URL a bit an error message is rendered in the HTML, however the code still attempts to do the regular rendering after that. You can see an error message in the console:

```
index.js:50 Uncaught (in promise) TypeError: Cannot read property 'sort' of undefined
    at fetchJSON.catch.then.repositories (index.js:50)
```
The error is in line 50, which should not be executed in the case of an error. Check this fundamental on promise chaining:

https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/promises.md#promise-chaining

PR #70, opened 2018-08-29
Shadi's homework Week3 (try&await)

shadoo77:


homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
-        createAndAppend('pre', root, { html: JSON.stringify(data, null, 2) });
+  async function main(url) {
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root);
+    createAndAppend('p', header, {html: 'HYF Repositories', class: 'header-lable'});
+    const select = createAndAppend('select', header, { id: 'selectMenu' });
+    // use promise object
+    try {
+      const data = await fetchJSON(url);
+        data.sort((a, b) => a.name.localeCompare(b.name, 'fr', {ignorePunctuation: true}));
```
JawharBairakdar: the only Punctuation is the hyphen in github repository names as far as I know. But using French has no reason unless you speak French :slightly_smiling_face:.

homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
-      } else {
-        createAndAppend('pre', root, { html: JSON.stringify(data, null, 2) });
+  async function main(url) {
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root);
+    createAndAppend('p', header, {html: 'HYF Repositories', class: 'header-lable'});
+    const select = createAndAppend('select', header, { id: 'selectMenu' });
+    // use promise object
+    try {
+      const data = await fetchJSON(url);
```
JawharBairakdar: Data is a very nice name, but it means on something anonymous, you have to be somehow creative with your names like what does it contains or something, maybe it's boring that is everyone else using it `repositories` but it's a meaningful one, or as I told be creative :heart:.

So What is data?

homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
+        contentRight(container, data, select);
+    }
+    catch(error) {
+      createAndAppend('div', root, { html: error.message, class: 'alert-error' });
+    }
+  }
+  
+  // function who creates a left content section
+  function contentLeft(container, data, select) {
+    const leftSection = createAndAppend('div', container, { id: 'leftSection', class: 'whiteframe' });
```
JawharBairakdar: In my vsCode the spell checker is complaining about this name, which is `whiteframe`!!
Shouldn't be camelCase ?

homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
+    showTableRepo(leftSection, data[0]);
+
+    select.addEventListener('change', (event) => {
+      leftSection.innerHTML = "";
+      const theEventIndex = event.target.value;
+      showTableRepo(leftSection, data[theEventIndex]);
```
JawharBairakdar: Can't we just move the `data[theEventIndex]` to it's own variable, it's the selected repository, Right?

homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
+  // drow the table
+function showTableRepo(leftSection, repository) {
+  // create table of the contents
+  const lTable = createAndAppend('table', leftSection);
+  // line one
+  const tr1 = createAndAppend('tr', lTable);
+  createAndAppend('td', tr1, {html: 'Repository :', class: 'lable'});
+  const td1 = createAndAppend('td', tr1);
+  const nameLink = '<a href="' + repository.html_url + '" target="_blank">' + repository.name + '</a>';
+  createAndAppend('p', td1, {html: nameLink });
```
JawharBairakdar: nothing wrong with this, But if we didn't ues the `innerHTML` you couldn't hard coded this like so.
the only thing you can do is to switch:
```javascript
const nameLink = createAndAppend('p', td1);
createAndAppend('a', nameLink, {
  html: repository.name,
  href: repository.html_url,
  target: "_blank"
});
```

homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
+  async function contentRight(container, data, select) {
+    const rightSection = createAndAppend('div', container, { id: 'rightSection', class: 'whiteframe' });
+    createAndAppend('h5', rightSection, {html: "Contributions :", class: 'contributionsLable'});
+    const contributorsURL = data[0].contributors_url;
+    try {
+      const contributors = await fetchJSON(contributorsURL);
+      showContributors(rightSection, contributors);
+    }
+    catch(error) {
+      createAndAppend('div', rightSection, { html: error.message, class: 'alert-error' });
```
JawharBairakdar: I think you have to return here before you got two different kind of errors, because the interpreter will continue running till the end of the function, and when ever we have an error we need to know exactly what is it.
so if an error happen somehow the `try catch` will prevent the program crashing and with that it will just continue till the end.

homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
+    }
+
+    select.addEventListener('change', async (event) => {
+      rightSection.innerHTML = "";
```
JawharBairakdar: I saw that you have a two different kind of code structure!!
So you can choose **double or single** quotes.

homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
+      createAndAppend('h5', rightSection, {html: "Contributions :", class: 'contributionsLable'});
+      const theEventIndex = event.target.value;
+      const contributorsURL = data[theEventIndex].contributors_url;
```
JawharBairakdar: that's what the variable made for :1st_place_medal: :tada:

homework/src/index.js
```diff
@@ -30,17 +32,109 @@
...
+    });
+  }
+
+  // drow the table
+function showTableRepo(leftSection, repository) {
+  // create table of the contents
+  const lTable = createAndAppend('table', leftSection);
+  // line one
+  const tr1 = createAndAppend('tr', lTable);
+  createAndAppend('td', tr1, {html: 'Repository :', class: 'lable'});
```
JawharBairakdar: `[cSpell Checker] Unknown word: "lable"`

PR #72, opened 2018-08-30
JS WEEk1

salar01:
JS3, WEEK1 is done.

### DIFF
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -1,10 +1,10 @@
-'use strict';
+"use strict";
 
 {
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
-    xhr.open('GET', url);
-    xhr.responseType = 'json';
+    xhr.open("GET", url);
+    xhr.responseType = "json";
     xhr.onload = () => {
       if (xhr.status < 400) {
         cb(null, xhr.response);
@@ -12,16 +12,16 @@
         cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
       }
     };
-    xhr.onerror = () => cb(new Error('Network request failed'));
+    xhr.onerror = () => cb(new Error("Network request failed"));
     xhr.send();
   }
 
   function createAndAppend(name, parent, options = {}) {
     const elem = document.createElement(name);
     parent.appendChild(elem);
-    Object.keys(options).forEach((key) => {
+    Object.keys(options).forEach(key => {
       const value = options[key];
-      if (key === 'html') {
+      if (key === "text") {
         elem.innerHTML = value;
       } else {
         elem.setAttribute(key, value);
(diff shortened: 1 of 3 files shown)


homework/src/index.js
```diff
@@ -31,17 +31,66 @@
...
+      class: "header"
+    });
+    const table = createAndAppend("table", root, { class: "table" });
+    const select = createAndAppend("select", header, { class: "select" });
+
+    const tbody = createAndAppend("tbody", table);
+    const tr = createAndAppend("tr", tbody);
+
+    createAndAppend("td", tr, tbody, { text: "Repository: ", class: "label" });
+    const repository = createAndAppend("a", tbody);
```
remarcmij: The result of creating an `a` tag is not a 'repository'. That name is misleading. You should reserve the name `repository` for a single element of the `repositories` array.

A better name could be for instance `repositoryLink`.

homework/src/index.js
```diff
@@ -31,17 +31,66 @@
...
+    const description = createAndAppend("tr", tbody);
```
remarcmij: Same here, you are creating a table row. It is not a 'description'. Make sure that the names you choose accurately reflect their content.

homework/src/index.js
```diff
@@ -31,17 +31,66 @@
...
+    const forks = createAndAppend("tr", tbody, { class: "label" });
```
remarcmij: Same here.

homework/src/index.js
```diff
@@ -31,17 +31,66 @@
...
+
+        forks.innerHTML = "Forks: " + repositories[index].forks;
+        description.innerHTML =
+          "Description: " + repositories[index].description;
+        updated.innerHTML =
+          "Updated: " +
+          new Date(repositories[index].updated_at).toLocaleString();
+      }
+
+      repositoryBox();
```
remarcmij: There is no point in defining an inline function and then immediately calling as you are doing here. You could easily comment out lines 70, 82 and 84 without it changing anything.

It is a good idea however to create a function and place it at file level, so not inside the event handler function. But then, you must pass all the information that that function requires through the parameter list. If you do that your code becomes better readable, because your function `repositoryBox` has no longer any external dependencies. The resulting code could look like this (leaving aside other coding issues that still remain in this code):

```js

    repositoryBox(repositories[0], repositoryLink, forks, description, updated);

    select.addEventListener("change", () => {
      repositoryBox(repositories[select.value], repositoryLink, forks, description, updated);
    });
  });
}

function repositoryBox(repository, repositoryLink, forks, description, updated) {
  repositoryLink.innerText = repository.name;
  repositoryLink.setAttribute("href", repository.html_url);
  repositoryLink.setAttribute("target", "_blank");

  forks.innerHTML = "Forks: " + repository.forks;
  description.innerHTML =
    "Description: " + repository.description;
  updated.innerHTML =
    "Updated: " +
    new Date(repository.updated_at).toLocaleString();
}
```

Furthermore, function names should contain a verb in their names as they perform an action (do this, do that, etc.), e.g. `renderRepositoryBox`.

homework/src/index.js
```diff
@@ -31,17 +31,66 @@
...
+        createAndAppend("div", root, {
+          text: error.message,
+          class: "alert-error"
+        });
       } else {
-        createAndAppend('pre', root, { html: JSON.stringify(data, null, 2) });
+        repositories.sort((a, b) => a.name.localeCompare(b.name));
       }
+
+      repositories.forEach((element, index) => {
```
remarcmij: If the server returns an error, say a 404 "Not Found" error, your code still attempts to continue after rendering an error in the HTML page. You can see this error message in the console:

```
index.js:63 Uncaught TypeError: Cannot read property 'forEach' of undefined
    at fetchJSON (index.js:63)
    at XMLHttpRequest.xhr.onload (index.js:12)
```

A possible solution could be to simply return from your `main()` function after rendering the error. In that case, you also do not need to use `else`:

```js
fetchJSON(url, (error, repositories) => {
  if (error) {
    createAndAppend("div", root, {
      text: error.message,
      class: "alert-error"
    });
    return;
  }

  repositories.forEach((element, index) => {
  //...
```

homework/src/index.js
```diff
@@ -31,17 +31,66 @@
...
+      });
+
+      function repositoryBox() {
+        const index = select.value;
+        repository.innerText = select[index].textContent;
+        repository.setAttribute("href", repositories[index].html_url);
+        repository.setAttribute("target", "_blank");
+
+        forks.innerHTML = "Forks: " + repositories[index].forks;
+        description.innerHTML =
```
remarcmij: Since `description` refers to a `tr` element, its `innerHTML` can only be `td` or `tr` elements. You can't/shouldn't set its `innerHTML` to a text string.

salar01:
Hello Jim, thanks a lot for spend a lot of time and look through my homework. I really apperciate you and i learnd a lot till now. I changed what you have mentioned in Week2 homework. I stopped with class16, to study more on JavaScript and i follow the courses with class17 24 September, i hope that i see you there, and be our teacher again 💯  👍  :)

remarcmij:
Hi Salar, I'm looking forward to meet and teach you again in JavaScript 2 with class 17.

PR #75, opened 2018-08-30
Week2 Homework

acimanx:
Known Issues:
1. CSS is pizzled, dead code is present.
2. JS can be layered with different practices. (some code was rewritten when I got more knowledge, some are not)
3. Disorganized naming of variables in JS and CSS
4. Login is not cached
5. SVG in CSS can be unstable on non-chrome browsers

_Tested on Chrome Dev Version 70.0.3534.4_

Live preview here: 
[https://i❤code.cf/dynapp](https://xn--icode-102c.cf/dynapp)

Screencap:
Preview


remarcmij:
Hi @acimanx, I assume you closed this PR because you wanted to withdraw it as your homework submission? At present we have only homework from you for week 1. Please advise.

acimanx:
Hello Jim, I was needed to do some changes because my homework had some serious issues:
it was possible to make injection and steal a data from end-user too easy. 
I hope I fixed it now.

acimanx:
If you'll look into the code you'll see that message body went into a not formatted tag, so potential attacker could just use this with a link to any script inside. 
In the new version I'm using <pre> (preformatted HTML tag) which is a temporary solution. It will show the code as it is, and will not execute it. 
There is still possibility for a breach, if you'll simulate a  closing of this tag with sending closing tag inside of a message body.
Probably there is a possibility to close this issue: if to 
use preformatting of the symbols in a string first, in js file. But I don't know if this will guarantee a protection against injection.
Can you advice something on this?

PR #78, opened 2018-08-30
Week3

mehmetscoskun:


homework/src/index.js
```diff
@@ -30,18 +42,141 @@
...
+        const contNum = contributor.contributions;
+
+        const contributorsTr = createAndAppend('tr', contributorsTbody, {
+          id: `${contributor['id']}`,
+          class: 'contributorsTr',
+          onclick: `window.open('${contLink}', '_blank');`,
+          onkeydown: `if(event.key === 'Enter'){window.open('${contLink}', '_blank');};`,
+          role: 'link',
+          tabindex: 0
+        });
```
remarcmij: Creating a table for this list is not the best solution. For one thing, it is not ARIA-compliant. Unless you add some extra ARIA attribute a screen reader will identify this as a table and not as a list. You also need to do complicated things here to respond to mouse click and the keyboard. If instead you used a `ul` with `li` and embedded `a` tags all this would become simpler. This is what another student did in terms of HTML structure:

```html
<ul class="contributor-list">
  <li class="contributor-item">
    <a target="_blank" href="https://github.com/gijscor">
      <img class="contributor-avatar" src="https://avatars0.githubusercontent.com/u/15912395?v=4">
      <div class="contributor-data">
        <div class="contributor_name">gijscor</div>
        <div class="contributor_badge">12</div>
      </div>
    </a>
  </li>
  <li class="contributor-item">
    <a target="_blank" href="https://github.com/wouterkleijn">
      <img class="contributor-avatar" src="https://avatars1.githubusercontent.com/u/37488847?v=4">
      <div class="contributor-data">
        <div class="contributor_name">wouterkleijn</div>
        <div class="contributor_badge">4</div>
      </div>
    </a>
  </li>
</ul>
```

homework/src/index.js
```diff
@@ -1,19 +1,31 @@
...
+  function fetchJSON(url, container) {
+    return new Promise((resolve, reject) => {
+      const xhr = new XMLHttpRequest();
+
+      const loadingGif = createAndAppend('img', container, {
+        src: './P1Chi6u1.gif',
+        alt: 'loading gif',
+        class: 'loadingGif'
+      });
+      xhr.onreadystatechange = () => {
```
remarcmij: This looks like a nice gimmick but really only makes sense if, on average, you expect a response to take considerable time. In this particular case the response comes so quickly (< 200ms) that the GIF only produces an incomprehensible flicker. So definitely use this in other cases where the response takes time, but in this case I would leave it out.

Also, as a general principle, you should try and avoid mixing 'concerns' in a single function. The `fetchJSON()` that deals with fetching data should not also be concerned with manipulating the DOM.

homework/src/index.js
```diff
@@ -30,18 +42,141 @@
...
-        createAndAppend('div', root, { html: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { html: JSON.stringify(data, null, 2) });
-      }
+  //this is an object creator
+  function RepositoryInfo(sourceObj) {
+    this.Description = sourceObj.description;
+    this.Forks = sourceObj.forks;
+    const dateOfUpdate = new Date(sourceObj.updated_at);
+    this.Updated = dateOfUpdate.toLocaleString();
```
remarcmij: Object property names should be in camelCase.

homework/src/index.js
```diff
@@ -30,18 +42,141 @@
...
 
-  function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { html: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { html: JSON.stringify(data, null, 2) });
-      }
+  //this is an object creator
```
remarcmij: Technically, this is called a constructor function.

homework/src/index.js
```diff
@@ -30,18 +42,141 @@
...
+    const dateOfUpdate = new Date(sourceObj.updated_at);
+    this.Updated = dateOfUpdate.toLocaleString();
+  }
+
+  //info-box of the page
+  function createInfoBox(repoData, container) {
+
+    container.innerHTML = "";
+
+    const info = new RepositoryInfo(repoData);
```
remarcmij: I'm not sure whether creating on object here rather than using the repo data directly adds value. Perhaps this was just a finger exercise?

mehmetsaidcoskun:
Thank you @remarcmij , I will make changes on my code in the way you lead. Thank you for everything you taught. :)

homework/src/index.js
mehmetsaidcoskun: Actually using ul-li elements would be simpler than the table element. However, I added some extra codes, therefore a user can navigate through all interactive elements using the keyboard (using the Tab and Enter keys). Pressing Enter on such an element should be equivalent to clicking the mouse. 
I did not add some aria codes about screen reader. Instead of replacing the table element by ul-li elements, is it enough if I add some aria codes about screen reader?

homework/src/index.js
mehmetsaidcoskun: Actually, you are right, I named them with starting capital letters because I used them in the web page. Here is the example: `Forks: 5`

homework/src/index.js
remarcmij: I think you will find that anchor elements (`a`) are navigable with the keyboard (Enter, Tab). In general, if you use HTML elements that are appropriate for the semantic context you are almost guaranteed to be ARIA-compliant. Only when you deviate from that you need extra ARIA attributes on your elements to provide the semantic context. So, when you want to create a list, use `ul` or `ol`. If you use a `table` than you would need add ARIA roles to label the table as a list and each row as a list item.

Check section 2 of  the link below (taken from the HTML/CSS module) for some simple rules for ARIA-compliance.

https://w3c.github.io/using-aria/

PR #84, opened 2018-09-01
Js3 Week3 Barbaros

whitepiolin:


homework/src/index.js
```diff
@@ -30,18 +33,95 @@
...
+      const initialRender = repositories[0];
+      renderRepository(divInfo, initialRender);
+      renderContributors(divCont, initialRender);
+    }
+
+    catch (error) {
+      error => {
+        createAndAppend('div', root, { html: error.message, class: 'alert-error', role: "alert" });
+        return;
+      };
```
remarcmij: Your error handler doesn't work. Have you tried it?

whitepiolin: I corrected code. I had not noticed this error before i uploaded it

homework/src/index.js
```diff
@@ -1,19 +1,24 @@
...
+      xhr.responseType = 'json';
+      xhr.onload = (function () {
+        if (xhr.status < 400) {
+          resolve(xhr.response);
+        } else {
+          reject(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
+        }
+      });
+      xhr.onerror = () => {
+        reject(alert("Opppsss, Network Error"));
```
remarcmij: The `reject` method must be called with error information that is to be passed to `.catch()` handlers. Here you are passing a call to `alert()`. The `alert()` function returns `undefined`. So the value that is passed to the `.catch()` is `undefined`. This causes the `catch()` handler to produce a runtime error in line 72, because it is expecting an `Error` object with a `message` property.

```
index.js:72 Uncaught (in promise) TypeError: Cannot read property 'message' of undefined
```

Moreover, you should not do any DOM manipulation or anything to do with the UI in this `fetchJSON()` function. This is a violation of the important principle of Separation of Concerns. The `fetchJSON()` should be concerned with fetching data only. In line 12 it is done correctly.

For instance, see this Stack Overflow question:

https://stackoverflow.com/questions/98734/what-is-separation-of-concerns

homework/src/index.js
```diff
@@ -30,18 +35,97 @@
...
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { html: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { html: JSON.stringify(data, null, 2) });
-      }
-    });
+  const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+  const root = document.getElementById('root');
```
remarcmij: You should place this statement definition `root` inside function `main()` so that it is guaranteed to be executed through `window.onload`. The `root` div may not have been created by the browser yet when this statement is executed as this point. The 'load' event is guaranteed to executed when the web page and all its resources are completely loaded.

homework/src/index.js
whitepiolin: yes, i think it is corrected

homework/src/index.js
whitepiolin: It is inside the main function now

homework/src/index.js
whitepiolin: @remarcmij Sorry for my errors. I think i understood what you mean and tried to correct it. Now i can not see an error. I hope it is done now. If you think it needs more corrections, please warn me again. Thanks

whitepiolin:
i have seen that in 3 places root seems undefined. I have corrected them. I hope it is more better now. thanks again

whitepiolin:
i tried to define root in the scopes of render functions in order to cancel undefined error.

PR #85, opened 2018-09-01
Js3 week3

WardMad:


homework/src/index.js
```diff
@@ -1,47 +1,121 @@
...
+    function renderRepo(container, repository, linkName) {
+
+        container.innerHTML = "";
+        const table = createAndAppend("table", container, { class: "table" });
+        const tr = createAndAppend("tr", table);
+        const td = createAndAppend("td", tr, { text: "<b>Repository: </b> " });
+
+        createAndAppend("a", td, {
+            text: repository.html_url,
+            text: linkName,
```
remarcmij: There is an ESLint warning about a duplicate `text` property. I guess the second one takes precedence over the first one. Nevertheless, something to correct.

homework/src/index.js
```diff
@@ -1,47 +1,121 @@
...
+
+    };
+
+    async function renderContributors(container, contributorsUrl) {
+
+        const ul = createAndAppend("ul", container, { text: "<small>Contributions: ", class: "table" });
+        const li = createAndAppend("li", ul);
+
+        try {
+            const response = await fetch(contributorsUrl);
```
remarcmij: You are not dealing with HTTP status >= 400 errors here.

homework/src/style.css
```diff
@@ -1,3 +1,134 @@
+body {
```
remarcmij: While you could argue that your app is responsive as it adjusts to the screen width, I think most people would expect the contributor list to be placed below the repo info rather than next to it in the case of a portrait mobile display.

This is would your app would look like on an iPhone 6:

responsive

homework/src/index.js
```diff
@@ -1,47 +1,121 @@
...
+        });
+        return elem;
+    }
+
+    function renderRepo(container, repository, linkName) {
+
+        container.innerHTML = "";
+        const table = createAndAppend("table", container, { class: "table" });
+        const tr = createAndAppend("tr", table);
+        const td = createAndAppend("td", tr, { text: "<b>Repository: </b> " });
```
remarcmij: Your table rows each have a single column (i.e. a single `td`) only. The idea of using a table here is to line up all repo values vertically. So the first column would contain the label and the second column the data. Your approach of just concatenating the value to its label with an intervening space works too, but doesn't need a table (and by implication should not use a table).

WardMad:
Thanks for comments Jim.
 I'll take note of your advice's.

PR #86, opened 2018-09-01
Gaffari Week3 homework

Gaffari02:


JawharBairakdar:
Hi @Gaffari02, Because you submitted your all 3 homeworks in the same day and we didn't really get time for review them all.
I am going to ask for a changes that has been asked before by me or @remarcmij that we make sure you did understand the basics of what you have learned so far.

Gaffari02:
Hello@ JawharBairakdar and @remarcmij ,
I have fixed the things that you asked for in my js3 homeworks. I still have a small problem which I couldnt fix. when I open the browser with error codes the head and select parts still appears. I tried a lot but I didn't find a solution. All the other things completed.

Thanks.
Gaffari

remarcmij:
HI @Gaffari02, the solution is probably simpler than you think. You just need to postpone creating the `select` element and container divs until after the initial call to `fetchJSON()`. If this call is successful, then you know there is no error and you can continue creating elements. In case of an error you render your error message instead, e.g.:

```js
async function main(url) {

  const root = document.getElementById('root', { class: "root" });
  const div = createAndAppend('div', root, { class: "headbox" });
  const heading = createAndAppend('heading', div, { html: "HYF Repositories", class: "heading" });

  try {

    const repositories = await fetchJSON(url)

    const select = createAndAppend('select', div, { class: "select" });
    const divContainers = createAndAppend('div', root, { class: "div-containers" });
    const leftContainer = createAndAppend('div', divContainers, { class: "container" });
    const rightContainer = createAndAppend('div', divContainers, { class: "right-container" });
    //...
   }
   catch (err) {
     createAndAppend('div', root, { html: err.message, class: 'alert-error' });
   }
}
```

Gaffari02:
HI @remarcmij , as you said the solution was simpler than I think. I postponed creating the headbox, heading, select and container divs until after the initial call to fetchJSON(). 
Finally it is done. Please let me know if anything else should be completed. 
Thanks for both of your kindly helps @remarcmij @JawharBairakdar

PR #95, opened 2018-10-14
Week1

kawaraa:
Here is my homework for week1

### DIFF
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -1,4 +1,4 @@
-"use strick";
+"use strict";
 
 function main(url) {
 
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -3,6 +3,11 @@
 function main(url) {
 
   const root = document.getElementById("root");
+  const header = createEl("header", root, {
+    text: "<label>HYF Repositories</label>"
+  });
+  const article = createEl("article", root);
+  const select = createEl("select", header);
 
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
@@ -31,48 +36,36 @@ function main(url) {
     return elem;
   }
 
-  fetchJSON(url, (err, data) => {
+  fetchJSON(url, (err, repos) => {
     if (err === null) {
-      const header = createEl("header", root);
-      createEl("h3", header, { text: "HYF Repositories" });
-      const select = createEl("select", header);
-      let names = [];
-      data.forEach(el => {
-        let name = el.name.charAt(0).toUpperCase() + el.name.slice(1);
-        names.push(name);
-      });
-      names.sort();
-      for (let i = 0; i < names.length; i++) {
-        let optn = createEl("option", select, { text: names[i] });
-        optn.value = i;
-      }
-      repos(data, select);
+      repos.sort((a, b) => a.name.localeCompare(b.name))
+        .forEach((repo, i) => {
+          let nm = repo.name.charAt(0).toUpperCase() + repo.name.slice(1);
+          createEl("option", select, { text: nm }).value = i;
+        });
+      renderRepo(repos, select.value);
     } else {
+      root.innerHTML = "";
       createEl("div", root, err);
     }
   });
 
-  function repos(data, select) {
-    let article = createEl("article", root);
+  function renderRepo(repos, index) {
+    article.innerHTML = "";
+    repoDetails(repos[index]);
+    contributors(repos[index].contributors_url);
 
-    select.addEventListener("change", onSelect);
-    function onSelect() {
-      let v1 = select.options[select.selectedIndex].text.toUpperCase();
-      for (let i = 0; i < data.length; i++) {
-        let v2 = data[i].name.toUpperCase();
-        if (v1 === v2) {
-          article.innerHTML = "";
-          repoDetails(data[i], article);
-          contributors(data[i].contributors_url, article);
-        }
-      }
-    }
-    onSelect();
+    select.addEventListener("change", () => {
+      article.innerHTML = "";
+      repoDetails(repos[select.value]);
+      contributors(repos[select.value].contributors_url);
+    });
 
   }
 
-  function repoDetails(details, parent) {
-    let table = createEl("table", parent);
+  function repoDetails(details) {
+
+    const table = createEl("table", article);
     let tr = createEl("tr", table);
     createEl("th", tr, { text: "Repository:" });
     let td = createEl("td", tr);
@@ -85,10 +78,10 @@ function main(url) {
     createEl("tr", table, { text: `<th>Updated:</th><td>${details.updated_at}</td>` });
   }
 
-  function contributors(url, parent) {
+  function contributors(url) {
     fetchJSON(url, (err, info) => {
       if (err === null) {
-        let ul = createEl("ul", parent);
+        const ul = createEl("ul", article);
         for (let i = 0; i < info.length; i++) {
           let li = createEl("li", ul);
           let a = createEl("a", li, { id: "href", val: info[i].html_url });
@@ -97,8 +90,8 @@ function main(url) {
           a.innerHTML += `<p>${info[i].login}</p><span>${info[i].contributions}</span>`;
         }
       } else {
-        parent.innerHTML = "";
-        createEl("div", parent, err);
+        article.innerHTML = "";
+        createEl("div", article, err);
       }
     });
   }
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -13,7 +13,7 @@ function main(url) {
     xhr.open('GET', url);
     xhr.responseType = 'json';
     xhr.onload = () => {
-      if (xhr.status < 400) {
+      if (xhr.status === 200) {
         cb(null, xhr.response);
       } else {
         cb(new Error(`Network Error: ${xhr.status} - ${xhr.statusText}`));
@@ -104,7 +104,6 @@ function main(url) {
       }
     });
   }
-
 
 }
 
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -3,10 +3,9 @@
 function main(url) {
 
   const root = document.getElementById("root");
-  const header = createEl("header", root, {
-    text: "<label>HYF Repositories</label>"
-  });
+  const header = createEl("header", root);
   const article = createEl("article", root);
+  createEl("label", header, { txt: "HYF Repositories" });
   const select = createEl("select", header);
 
   function fetchJSON(url, cb) {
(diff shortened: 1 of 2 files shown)


homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+  const root = document.getElementById("root");
 
-{
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
     xhr.open('GET', url);
     xhr.responseType = 'json';
     xhr.onload = () => {
-      if (xhr.status < 400) {
+      if (xhr.status === 200) {
```
remarcmij: By changing the condition of the `if` statement you are now considering any status code not equal to 200 as an error. This is an error by itself. Any code equal or greater than 400 is an error. All codes less than 400 are not errors. See: https://www.restapitutorial.com/httpstatuscodes.html

If you expect a 200 code but get some other code but still less than 400 you need to find some other way of dealing with. But you can't report it to the user as an HTTP error.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
         cb(null, xhr.response);
       } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
+        cb({ text: `Error: ${xhr.status} - ${xhr.statusText}`, id: "id", val: "error" });
```
remarcmij: By convention, error information is reported through a standard JavaScript Error object (as the original code did). Why did you choose to change this?

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
       }
     };
-    xhr.onerror = () => cb(new Error('Network request failed'));
+    xhr.onerror = () => {
+      cb({ text: "Network request failed", id: "id", val: "error" });
```
remarcmij: Use an Error object (as per the original code).

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+    };
     xhr.send();
   }
 
-  function createAndAppend(name, parent, options = {}) {
+  function createEl(name, parent, obj) {
```
remarcmij: Your `createEl()` function is a less capable version than the original `createAndAppend()` function. The original `createAndAppend()` could set _any_ attribute. Your version can only deal with the `id` attribute. I have a hard time understanding why you chose to dumbed down this function.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+  fetchJSON(url, (err, data) => {
+    if (err === null) {
+      const header = createEl("header", root);
+      createEl("h3", header, { text: "HYF Repositories" });
+      const select = createEl("select", header);
+      let names = [];
```
remarcmij: ESLint warns that you could have used `const` instead of `let`. You should follow up on such warnings.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+    }
     return elem;
   }
 
-  function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+  fetchJSON(url, (err, data) => {
```
remarcmij: Do not use the name `data` if you know exactly the type of data that it will contain. In this case `data` is an array of repositories. So everywhere where this `data` variable is used you have mentally associate it with what it contains (repositories), because the word `data` is denotes generic data.

If you change this parameter name `data` to `repositories` then you need to do this mental mapping, because the name exactly describes what the parameter holds.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+    if (err === null) {
+      const header = createEl("header", root);
+      createEl("h3", header, { text: "HYF Repositories" });
+      const select = createEl("select", header);
+      let names = [];
+      data.forEach(el => {
```
remarcmij: One might ask here: what is `data` and what is `el`. It is not self-explanatory. If you had renamed `data` to `repositories` this could read as:

```js
```

Now it immediately clear what this `forEach()` is about.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+        let name = el.name.charAt(0).toUpperCase() + el.name.slice(1);
```
remarcmij: I appreciate that you are trying to make the repository names nicer by making sure that the start with an uppercase letter, but it is better to leave those names untouched because those names are the real names. For instance, there is a `database` repository on the HYF account, but not a `Database` repository.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+        names.push(name);
+      });
```
remarcmij: Irrespective of my previous comment, this `forEach()` would have better handled with ` map()`.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
-'use strict';
+"use strick";
+
+function main(url) {
```
remarcmij: You should not embed all functions inside one gigantic `main()` function. The skeleton code that was provided already had the correct structure.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+      createEl("h3", header, { text: "HYF Repositories" });
+      const select = createEl("select", header);
+      let names = [];
+      data.forEach(el => {
+        let name = el.name.charAt(0).toUpperCase() + el.name.slice(1);
+        names.push(name);
+      });
+      names.sort();
+      for (let i = 0; i < names.length; i++) {
+        let optn = createEl("option", select, { text: names[i] });
```
remarcmij: Why did you use the name `optn` instead of `option`? It is only two letter more to type and so much more readable. And note that ESLint warns that you could have used `const` instead of `let`.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+      const header = createEl("header", root);
+      createEl("h3", header, { text: "HYF Repositories" });
+      const select = createEl("select", header);
+      let names = [];
+      data.forEach(el => {
+        let name = el.name.charAt(0).toUpperCase() + el.name.slice(1);
+        names.push(name);
+      });
+      names.sort();
+      for (let i = 0; i < names.length; i++) {
```
remarcmij: As you have used `forEach()` in other places, you could have used it here too (for consistency).

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+    } else {
+      createEl("div", root, err);
+    }
+  });
+
+  function repos(data, select) {
+    let article = createEl("article", root);
+
+    select.addEventListener("change", onSelect);
+    function onSelect() {
```
remarcmij: I would have used an anonymous arrow function here:

```js
 select.addEventListener("change", () => {
```

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+        let optn = createEl("option", select, { text: names[i] });
+        optn.value = i;
+      }
+      repos(data, select);
+    } else {
+      createEl("div", root, err);
+    }
+  });
+
+  function repos(data, select) {
```
remarcmij: Functions should always indicate some **action**, i.e. have a verb in its name. For instance, `fetchAndRenderRepos()`.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+    if (err === null) {
+      const header = createEl("header", root);
+      createEl("h3", header, { text: "HYF Repositories" });
+      const select = createEl("select", header);
+      let names = [];
+      data.forEach(el => {
+        let name = el.name.charAt(0).toUpperCase() + el.name.slice(1);
+        names.push(name);
+      });
+      names.sort();
```
remarcmij: You could have saved yourself some headaches by sorting the repositories array itself rather than a new array that just contains the names. Then later in your code (lines 62-32), where you use a loop to find the repository that correspond to the name would not have been necessary, because the index value that you store as the `value` attribute of the `option` element corresponds then exactly with the selected repository.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+          repoDetails(data[i], article);
+          contributors(data[i].contributors_url, article);
+        }
+      }
+    }
+    onSelect();
+
+  }
+
+  function repoDetails(details, parent) {
```
remarcmij: This function name should contain a verb to indicate **action**.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+    let table = createEl("table", parent);
```
remarcmij: There are several variable here for which the `let` could be a `const`, as indicated by ESLint.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
-    Object.keys(options).forEach((key) => {
-      const value = options[key];
-      if (key === 'text') {
-        elem.innerText = value;
-      } else {
-        elem.setAttribute(key, value);
-      }
-    });
+    if (obj) {
+      if (obj.text) { elem.innerHTML = obj.text; }
```
remarcmij: The original function used `innerText` instead of `innerHTML`. We did that on purpose, because we want to discourage this use of `innerHTML` for creating HTML elements.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+
+  function repoDetails(details, parent) {
+    let table = createEl("table", parent);
+    let tr = createEl("tr", table);
+    createEl("th", tr, { text: "Repository:" });
+    let td = createEl("td", tr);
+    let a = createEl("a", td, { text: details.name, id: "href", val: details.html_url });
+    a.setAttribute("target", "_blank");
+    if (details.description) {
+      createEl("tr", table, { text: `<th>Description:</th><td>${details.description}</td>` });
```
remarcmij: As mentioned, we discourage the use of `innerHTML` for creating HTML elements.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+
+  }
+
+  function repoDetails(details, parent) {
+    let table = createEl("table", parent);
+    let tr = createEl("tr", table);
+    createEl("th", tr, { text: "Repository:" });
+    let td = createEl("td", tr);
+    let a = createEl("a", td, { text: details.name, id: "href", val: details.html_url });
+    a.setAttribute("target", "_blank");
```
remarcmij: The original `createAndAppend()` function could have set this `target` attribute.

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+    onSelect();
+
+  }
+
+  function repoDetails(details, parent) {
+    let table = createEl("table", parent);
+    let tr = createEl("tr", table);
+    createEl("th", tr, { text: "Repository:" });
+    let td = createEl("td", tr);
+    let a = createEl("a", td, { text: details.name, id: "href", val: details.html_url });
```
remarcmij: Your `createEl()` function doesn't know how to deal with a `val` property (what would be its purpose anyway)?

homework/src/index.js
```diff
@@ -1,47 +1,110 @@
...
+    }
+    createEl("tr", table, { text: `<th>Forks:</th><td>${details.forks_count}</td>` });
+    createEl("tr", table, { text: `<th>Updated:</th><td>${details.updated_at}</td>` });
+  }
+
+  function contributors(url, parent) {
+    fetchJSON(url, (err, info) => {
+      if (err === null) {
+        let ul = createEl("ul", parent);
+        for (let i = 0; i < info.length; i++) {
```
remarcmij: Why not a `.forEach()`?

homework/src/index.js
```diff
@@ -1,47 +1,103 @@
-'use strict';
+"use strick";
```
remarcmij: This is a silly typing error to make (and flagged by the spelling checker). Your code is now not using strict mode.

kawaraa: I can see all you have mentioned and I will fix them and come back again! thanks for your comments

homework/src/index.js
```diff
@@ -1,47 +1,111 @@
...
-  function createAndAppend(name, parent, options = {}) {
+  function createEl(name, parent, options = {}) {
     const elem = document.createElement(name);
     parent.appendChild(elem);
-    Object.keys(options).forEach((key) => {
-      const value = options[key];
-      if (key === 'text') {
-        elem.innerText = value;
+    for (let key in options) {
+      if (key === "txt") {
```
remarcmij: There is really no reason to use cryptic abbreviations. Why not just used `text` instead of `txt` or `name` instead of `nm`. Professionals write clean code, see for instance this link:

https://medium.com/mindorks/how-to-write-clean-code-lessons-learnt-from-the-clean-code-robert-c-martin-9ffc7aef870c

homework/src/index.js
remarcmij: This is still not fixed.

PR #105, opened 2018-10-24
Week2

Mohamad-AezAldin:
please check the homework of week2

### DIFF
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -30,21 +30,55 @@
     return elem;
   }
 
+  function renderRepositoryInfo(info) {
+    const li = createAndAppend('li', ul);
+    const span = createAndAppend('span', li);
+    span.innerText = 'Repository : ';
+    const a = createAndAppend('a', li, { href: info.RepositoryUrl, target: '_blank' });
+    a.innerText = `${info.Repository}`;
+    const li2 = createAndAppend('li', ul);
+    li2.innerText = `Description :  ${info.Description}`;
+    const li3 = createAndAppend('li', ul);
+    li3.innerText = `Forks :  ${info.Forks}`;
+    const li4 = createAndAppend('li', ul);
+    li4.innerText = `Updated :  ${info.Updated}`;
+  }
+
+  function renderContributorsInfo(info) {
+    fetchJSON(info.ContributorsUrl, (err, data) => {
+      if (err) {
+        createAndAppend('div', div2, { text: err.message, class: 'alert-error' });
+      } else {
+        ul2.innerText = '';
+
+        data.forEach(elem => {
+          const ul3 = createAndAppend('ul', ul2);
+          const li = createAndAppend('li', ul3);
+          createAndAppend('img', li, { src: elem.avatar_url });
+          const li2 = createAndAppend('li', ul3);
+          const a = createAndAppend('a', li2, { href: elem.html_url, target: '_blank' });
+          a.innerText = `${elem.login}`;
+          const li3 = createAndAppend('li', ul3);
+          li3.innerText = `${elem.contributions}`;
+
+        });
+      }
+    });
+  }
+
+  const root = document.getElementById('root');
+  const header = createAndAppend('header', root, { id: 'header' });
+  const h3 = createAndAppend('h3', header);
+  h3.innerText = 'HYF Repository';
+  const mainDiv = createAndAppend('div', root, { id: 'main' });
+  const div1 = createAndAppend('div', mainDiv, { id: 'div1' });
+  const div2 = createAndAppend('div', mainDiv, { id: 'div2' });
+  const ul = createAndAppend('ul', div1, { id: 'list-container' });
+  const ul2 = createAndAppend('ul', div2, { id: 'list-container' });
 
   function main(url) {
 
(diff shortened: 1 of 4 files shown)


homework/src/index.js
```diff
@@ -30,18 +32,113 @@
...
+  const header = createAndAppend('header', root, { id: 'header' });
+  const h3 = createAndAppend('h3', header);
+  h3.innerText = 'HYF Repository';
+  const mainDiv = createAndAppend('div', root, { id: 'main' });
+  const div1 = createAndAppend('div', mainDiv, { id: 'div1' });
+  const div2 = createAndAppend('div', mainDiv, { id: 'div2' });
+  const ul = createAndAppend('ul', div1, { id: 'list-container' });
+  const Contributions = createAndAppend('h3', div2)
+  Contributions.innerText = 'Contributions';
+  const ul2 = createAndAppend('ul', div2, { id: 'list-container' });
```
remarcmij: 1. Lines 35-45 are executed as soon at `index.js` is loaded. It is best to start execution when the page is fully loaded, i.e., using the `window.onload` event handler. Currently this is set up to run the function `main()`. Therefore you should move these lines to the start of function `main()` just before the `fetchJSON()` call.

2. Avoid making names unique by using a number suffix (`div1`, `div2`). Spend a bit more effort in coming up with fitting names: what are these `div`-s used for? The answer should be reflected in their names. For instance: `'leftHandContainer` and `rightHandContainer`, or `repoDiv` and `contributorDiv` etc.

3. Remember that all variable and function names should be in camelCase: `Contributions` -> `contributions`.

homework/src/index.js
```diff
@@ -30,18 +32,113 @@
...
+
+  function renderRepositoryInfo(info) {
+    const li = createAndAppend('li', ul);
+    const span1 = createAndAppend('span', li, { id: 'title' });
+    span1.innerText = 'Repository: ';
```
remarcmij: You can set the `innerText` directly using `createAndAppend()`:

```js
 createAndAppend('span', li, { id: 'title', text: 'Repository' });
```

If you do it like this you can also get rid of a lot of variables (making you code less complex), because you have many variables here which are only needed to set the `innerText`.

homework/src/index.js
```diff
@@ -30,18 +32,113 @@
...
+    const li4 = createAndAppend('li', ul);
+    const span4 = createAndAppend('span', li4, { id: 'title' });
+    span4.innerText = 'Updated: ';
+    const p4 = createAndAppend('p', li4)
+    p4.innerText = info.Updated;
+  }
+
+  function renderContributorsInfo(info) {
+    fetchJSON(info.ContributorsUrl)
+      .then(data => {
```
remarcmij: `data` -> `contributors`

homework/src/index.js
```diff
@@ -30,18 +32,113 @@
...
+        ul2.innerText = '';
+        data.forEach(elem => {
```
remarcmij:
```js
```

...please!

homework/src/index.js
```diff
@@ -30,18 +32,113 @@
...
+        createAndAppend('div', div2, { text: err.message, class: 'alert-error' });
+      });
+  }
+
   function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
+    fetchJSON(url)
+      .then(data => {
```
remarcmij: `data` -> `repositories` !

You could sort the array of the repositories at this point. There is no need to create a separate array (the one you called `queries`,  you could use the `repositories` array instead. Your `queries` array is just a subset of fields of the `repositories` array. Nothing is gained from creating this extra array, it is just adding 'noise'.

homework/src/index.js
```diff
@@ -30,18 +32,113 @@
...
+        const select = createAndAppend('select', header);
+        const notSortedQueries = [];
+        data.forEach(elem => {
```
remarcmij:
```js
```

!!!

homework/src/index.js
```diff
@@ -30,18 +32,113 @@
...
+          const item = { Repository: elem.name, Description: elem.description, Forks: elem.forks, Updated: elem.updated_at, ContributorsUrl: elem.contributors_url, RepositoryUrl: elem.html_url };
+          notSortedQueries.push(item);
+        });
+        const queries = notSortedQueries.sort(function (a, b) {
```
remarcmij: The assignment suggested to use `.localeCompare()`. It is case-insensitive and takes all language specific sorting aspects into consideration, e.g. accented letters (like in French) and letters with diacritics (like in Turkish). It uses the language setting as configured by the user in the operating system.

homework/src/index.js
```diff
@@ -30,18 +32,113 @@
...
+
+        renderRepositoryInfo(queries[0]);
+
+        renderContributorsInfo(queries[0]);
+
+        select.addEventListener('change', () => {
+          // console.log(select.value);
+
+          // create the lift div1
+          const que = queries[select.value];
```
remarcmij: The name `que` can be mistaken to mean `queue`, which is a completely different concept in computing from `query`. Avoid such confusing abbreviations. If you have an array of `queries` then a single element is called a `query`.

PR #108, opened 2018-10-26
Week2

salar01:


homework/src/index.js
```diff
@@ -22,25 +23,114 @@
...
+          })
+          .catch(error => {
+            createAndAppend('div', container, {
+              text: error.message,
+              class: 'alert-error'
+            });
+            return;
+          });
+
+        function renderRepositoryBox(repositories) {
```
remarcmij: @J05J4 made this comment in the week 1 review:

>You have to improve your code to make it clean: there are functions inside of functions and there are function declarations underneath the call to that function. You need to update this (could also be in your week two solution). If you do not understand this completely please ask questions in the next lesson, because this is really important!

I expected that you made the necessary changes in your week 2 homework, but I don't see them.

homework/src/index.js
```diff
@@ -22,25 +23,114 @@
...
+
+        renderRepositoryBox(
+          repositories[0],
+          repositoryLink,
+          description,
+          forks,
+          updated
+        );
+
+        function renderContributors(contributor, container) {
```
remarcmij: This function definition should be outside your function `main()`.

homework/src/index.js
```diff
@@ -22,25 +23,114 @@
...
+  const select = createAndAppend('select', header, { id: 'list' });
+  const table = createAndAppend('table', root, { class: 'table' });
+  const tbody = createAndAppend('tbody', table);
+  const tr = createAndAppend('tr', tbody, { text: 'Repository: ', class: 'label' });
+
+  const repositoryLink = createAndAppend('a', tr);
+  const description = createAndAppend('tr', tbody, { class: 'label' });
+  const forks = createAndAppend('tr', tbody, { class: 'label' });
+  const container = createAndAppend('div', root, { id: 'container' });
+  const updated = createAndAppend('tr', tbody, { class: 'label' });
```
remarcmij: Lines 34-46 are executed immediately when `index.js` is loaded. It is best to postpone execution until the web page is fully loaded. This happens on the `load` event handled by the `windows.onload` event handler. In our case this is function `main()`. It is therefore best to move these lines inside the function `main()`, just before the first `fetchJSON()`. This could also mean that you may need to pass some of these variables as parameters to your other functions, as they are. no longer globally accessible. Using parameters instead of global variables makes your code more maintainable as you can directly what data the function needs by looking at its parameter list.

homework/src/index.js
```diff
@@ -22,25 +23,114 @@
...
+            return;
+          });
+
+        function renderRepositoryBox(repositories) {
+          repositoryLink.innerText = repositories.name;
+          repositoryLink.setAttribute('href', repositories.html_url);
+          repositoryLink.setAttribute('target', '_blank');
+          forks.innerHTML = 'Forks: ' + repositories.forks;
+          description.innerHTML = 'Description: ' + repositories.description;
+          updated.innerHTML = 'Updated: ' + new Date(repositories.updated_at).toLocaleString();
```
remarcmij: You are creating invalid HTML here. A `tr` element can only have `td` and `th` elements as its direct children.  It should not contain text nodes. But you are creating this:

```html
<table class="table">
  <tbody>
    <tr class="label">Repository: <a href="https://github.com/HackYourFuture/alumni" target="_blank">alumni</a></tr>
    <tr class="label">Description: Alumni Guides, everything you need to land a job</tr>
    <tr class="label">Forks: 8</tr>
    <tr class="label">Updated: 16/07/2018, 11:46:31</tr>
  </tbody>
</table>
```

homework/src/style.css
```diff
@@ -1,3 +1,124 @@
...
+    .contributor .header .select {
+        width: 60%;
+        
+    }
+}
+@media (min-width: 1025px) {
+  .contributor .header .select {
+    width: 60%;
+  }
+}
```
remarcmij: Your CSS sucks. The application does not look good and is not responsive:

On my computer screen:

screenshot 2018-11-01 at 14 11 06

On an iPhone6/7/8:

screenshot 2018-11-01 at 14 11 38

homework/src/index.js
salar01: Hello Jim, Thanks for your feedback, Now i understand that functions should not ne in another function, and what should i menthioned in Main function. I have already made changes in JS-WEEK3 and splited functions from each other to be moe redable. Tomorrow i will make a pull request for JS_WEEK3.  And after that i will change week2 and week 1 as soon as possible :)

homework/src/index.js
salar01: I have changed also this part in JS-WEEK3 and  i will do that aslo in week1 and week2

homework/src/style.css
salar01: It's done :)

homework/src/index.js
```diff
@@ -22,25 +23,114 @@
...
+    fetchJSON(url)
+      .then(repositories => {
+        repositories.forEach((item, index) => {
+          createAndAppend('option', select, {
+            text: item.name,
+            value: index
+          });
+        });
 
+        repositories.sort((a, b) => a.name.localeCompare(b.name));
```
remarcmij: UPDATE: I forgot to make this comment: there is a (sorting) mismatch between what you display in the select box and the repository you display. The entries in the select element are **not* sorted.

For instance, if I select the `JavaScript1` repo, I get the `CommandLine` repo. Did you see that too?

PR #109, opened 2018-10-26
Week2

kawaraa:
I'm really sorry for late, the homework was ready from 2 weeks, I thought myself have made the pull request, but it turns out not!


kawaraa:
I'm sorry for late reply, I was so busy, so I forgot to check my Email.
I have Implemented as required and also implemented some other things to make it more readable.

PR #112, opened 2018-10-27
Week2

joudi6:
JS3 week 2 H.W


homework/src/index.js
```diff
@@ -29,19 +32,119 @@
...
+          .catch((err) => main2(err));
+
+      });
+      renderRepositories(repositories[0], leftHandContainer);
+
+      fetchJSON(repositories[0].contributors_url)
+        .then((contributors) => {
+          renderContributors(contributors, rightHandContainer);
+        })
+        .catch((err) => main2(err));
```
remarcmij: Lines 120-125 and lines 128-134 do almost the exact same thing. Can you think of a function that you could write to remove this duplication? You could pass that function two parameters:

1. The repository to render.
2. The container that acts as the parent element.

homework/src/index.js
```diff
@@ -29,19 +32,119 @@
...
+      createAndAppend('img', a, {
+        src: contributor.avatar_url
+      });
+      createAndAppend('div', a, {
+        text: contributor.contributions
+      });
+    });
+  }
+
+  function main2(err, repositories) {
```
remarcmij: Your function `main2()` (can you think of a better name? (I guess you wouldn't call your second child `child2`, would you? 🙂) is called for two almost unrelated things:

1. Normal processing of the repositories.
2. Rendering errors (even by calling itself, recursively).

It would be more clear if you separated those two things. You could create a separate function for error handling:

```js
function handleError(err) {
}

function main2(repositories) {
  // ...
  fetchJSON(...)
    .then(...) {
      //...
    })
   .catch((err) => handleError(err));
}

function main1(url) {
  fetchJSON(url)
    .then((data) => main2(data))
    .catch((err) => handleError(err));
}

//...
```

PR #113, opened 2018-10-27
Week 2

afra71:


homework/src/index.js
```diff
@@ -30,17 +33,108 @@
...
+    createAndAppend('td', tr4, {
+      text: 'Updated:'
+    });
+    createAndAppend('td', tr4, {
+      text: repository.updated_at
+    });
+  }
+  function renderContributors(contributors, container) {
+    container.innerHTML = '';
+    const h2 = createAndAppend('h2', container, {
```
remarcmij: ESLInt flags the variable `h2` as 'unused', therefore you can delete it and just call `createAndAppend()` without saving its return value:

```js
createAndAppend('h2', container, {
```

homework/src/index.js
```diff
@@ -30,17 +33,108 @@
...
   }
 
+  function main(url) {
+    const root = document.getElementById('root');
+    fetchJSON(url)
+      .then(repositories => {
+        repositories.sort((a, b) => a.name.localeCompare(b.name));
+        console.log(repositories);
+        const header = createAndAppend('header', root);
+        const h1 = createAndAppend('h2', header, {
```
remarcmij: ESLint flags the variable `h1` as 'unused', therefore you can (and should) remove it.

homework/src/index.js
```diff
@@ -30,17 +33,108 @@
...
+        });
+        select.addEventListener('change', () => {
+          const index = select.value;
+          const repository = repositories[index];
+          renderRepository(repository, leftHandContainer);
+          fetchJSON(repository.contributors_url)
+            .then(contributors => {
+              renderContributors(contributors, rightHandContainer);
+            })
+            .catch(err => err.message);
```
remarcmij: You are not rendering the error to the DOM. When an error occurs there is no way for a user to know what happened. In case of an error you should, for instance, create a `div`, set its text content to the error message string and append the `div` as a child to the `root` div. Look as line 136 where this is done.

homework/src/index.js
```diff
@@ -30,17 +33,108 @@
...
+        });
+        renderRepository(repositories[0], leftHandContainer);
+        fetchJSON(repositories[0].contributors_url)
+          .then(contributors => {
+            renderContributors(contributors, rightHandContainer);
+          })
+          .catch(err => err.message);
```
remarcmij: Lines 121-126 and lines 128-133 are almost identical. Could you think of a function that you could call that replaces this duplication of code? (The DRY principle: Don't Repeat Yourself).

homework/test/app.js
```diff
@@ -0,0 +1,130 @@
+'use strict';
```
remarcmij: You should not includes file that are not part of the homework in the PR/

homework/src/style.css
```diff
@@ -1,3 +1,74 @@
 .alert-error {
```
remarcmij: You still need to work on the CSS styling.

PR #114, opened 2018-10-27
Week2

heboshka:


homework/src/index.js
```diff
@@ -30,17 +31,88 @@
...
+    url.innerText = repositoriesdata[index].name;
+    url.setAttribute('href', repositoriesdata[index].html_url);
+    if (repositoriesdata[index].description === null) {
+      descriptionName.innerHTML = '';
+      descriptionDetails.innerText = '';
+    } else {
+      descriptionName.innerText = "Description: ";
+      descriptionDetails.innerText = repositoriesdata[index].description;
+    }
+    forkNum.innerText = repositoriesdata[index].forks;
```
remarcmij: You could directly set the text through the `options` parameter in the `createAndAppend()` call in  line 49.

homework/src/index.js
```diff
@@ -30,17 +31,88 @@
...
+    const updateRepo = new Date(repositoriesdata[index].updated_at);
+    updateNum.innerText = updateRepo.toLocaleString("en-US");
```
remarcmij: 1. You could directly set the text in the `createAndAppend()` call.
2. If you use `.toLocaleString()` without an argument the default locale that the user selected in the operating system is used. For instance, in my computer I have set the default locale to use a European date and time format. You are explicitly setting it here to use the US format (e.g. using AM/PM suffixes).

homework/src/index.js
```diff
@@ -30,17 +31,88 @@
...
+        const contributorsInfo = repositories[0].contributors_url;
+        renderRepositories(0, repositories, container);
+        select.addEventListener('change', (e) => {
+          container.innerHTML = "";
+          const index = e.target.value;
+          renderRepositories(index, repositories, container);
+          const contributorOnSelect = repositories[index].contributors_url;
+          fetchJSON(contributorOnSelect)
+            .then(contributorOnSelect => renderContributors(contributorOnSelect, container))
+            .catch(err => err.message);
```
remarcmij: You are not rendering the error to the web page. You have done it correctly in line 112.

PR #117, opened 2018-11-01
Week3

afra71:
Homework Week 3 assigment 2.1


homework/src/index.js
```diff
@@ -30,17 +33,115 @@
...
+        try {
+          const contributors = await fetchJSON(repository.contributors_url);
+          renderContributors(contributors, rightHandContainer);
+        }
+        catch (err) {
+          err.message;
+        }
+      });
+      renderRepository(repositories[0], leftHandContainer);
+      try {
```
remarcmij: Here you are nesting a `try` block inside another `try` block (the one that starts in line 97). Since you are already in a `try` block here there is no reason to start another one as the outer `try` block will already catch errors. In this case it is best to remove the inner `try` block (remove lines 132 and 135-137).

Note that you do need the try/catch block in the event handler function that starts in line 119, as this event handler is executed asynchronously, that is after function `main()` has already finished executing.

homework/src/index.js
```diff
@@ -30,17 +33,115 @@
...
 
+  async function main(url) {
+    const root = document.getElementById('root');
+    try {
+      const repositories = await fetchJSON(url);
+
+      repositories.sort((a, b) => a.name.localeCompare(b.name));
+      console.log(repositories);
+      const header = createAndAppend('header', root);
+      const h1 = createAndAppend('h2', header, {
```
remarcmij: Variable `h1` is unused (as flagged by ESLint). You can just remove it, e.g.:

```js
createAndAppend('h2', header, {
```

This also gets rid of the confusion of you naming the variable `h1` while you are creating an `<h2>`.

homework/src/index.js
```diff
@@ -30,17 +33,115 @@
...
+    createAndAppend('td', tr4, {
+      text: 'Updated:'
+    });
+    createAndAppend('td', tr4, {
+      text: repository.updated_at
+    });
+  }
+  function renderContributors(contributors, container) {
+    container.innerHTML = '';
+    const h2 = createAndAppend('h2', container, {
```
remarcmij: The variable `h2` is unused.

PR #118, opened 2018-11-01
Week3

Damla78:
Class version


Week3/homework/src/App.js
```diff
@@ -0,0 +1,131 @@
...
+
+      this.fetchContributorsAndRender(select.selectedIndex);
+
+    } catch (error) {
+      this.renderError(error);
+    }
+
+  }
+
+  async onSelectionChange(url) {
```
remarcmij: You are not getting a `url` as a parameter to this function, but the `id` property of the repository. So `url` is a misleading and confusing name. And you are not using this value anyway. The test that you do in lines 53-55 is unneeded (and confusing) as there this value will never be the empty string.

Week3/homework/src/App.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+    try {
+      this.fetchContributorsAndRender(select.selectedIndex);
+    } catch (error) {
+      this.renderError(error);
+    }
+
+  }
+
+  sortData(data, key) {
```
remarcmij: I already made a comment about not needing this function at all in a previous review.

Week3/homework/src/Contributor.js
```diff
@@ -0,0 +1,25 @@
...
+  }
+
+  /**
+   * Render the contributor info to the DOM.
+   * @param {HTMLElement} contributorList The parent element in which to render the contributor.
+  */
+  render(contributorList) {
+    // Replace this comment with your code
+
+    const itemUser = Util.createAndAppend('li', contributorList, { src: this.data.avatar_url, class: 'userOne' });
```
remarcmij: An `li` has no use for an `src` attribute.

Week3/homework/src/Repository.js
```diff
@@ -0,0 +1,56 @@
...
+
+  /**
+   * Render the repository info to the DOM.
+   * @param {HTMLElement} parent The parent element in which to render the repository.
+   */
+  async render(parent) {
+    //
+    // Replace this comment with your code
+    //
+    const index = document.getElementById('selectRepositories').selectedIndex;
```
remarcmij: You are not using this `index` variable at all.

PR #119, opened 2018-11-01
Week3

heboshka:


heboshka:
Hiba HW3 classes part

homework/src/App.js
```diff
@@ -12,18 +12,25 @@ class App {
...
+    const select = Util.createAndAppend('select', header, { 'id': 'select-list', "aria-label": "HYF Repositories" });
     try {
-      // ...
       const repos = await Util.fetchJSON(url);
-      this.repos = repos.map(repo => new Repository(repo));
-      // ...
+      this.repos = repos
+        .sort((x, y) => x.name.localeCompare(y.name))
+        .map(repo => new Repository(repo));
+      repos.forEach((repo, i) => {
```
remarcmij: A casual reader might think that only `this.repos` is sorted and that `repos` itself is unmodified. It is only because the `.sort()` array method does an _in-place_ sort that this code works, despite it being used inside a method chain as is done here. You could change this code to the snippet below to take away this confusion:

```js
this.repos.forEach((repo, i) => {
  Util.createAndAppend('option', select, { 'text': repo.name(), 'value': i });
});
```

homework/src/Contributor.js
```diff
@@ -13,6 +13,16 @@ class Contributor {
    * @param {HTMLElement} contributorList The parent element in which to render the contributor.
   */
   render(contributorList) {
-    // Replace this comment with your code
+    const li = Util.createAndAppend('li', contributorList, { 'aria-label': this.data.login, 'tabindex': '0' });
+    li.addEventListener("click", () => { window.open(this.data.html_url); });
+    li.addEventListener("keypress", () => {
```
remarcmij: See my comment in week 2 for an alternative using the `<a>` element to make this navigatable with the tab key.

PR #122, opened 2018-11-03
js3 Week2

Anashost:


homework/src/index.js
```diff
@@ -30,15 +34,96 @@
...
+    const descriptions = `${info.description} `;
+    addRow('Description: ', descriptions, tbody);
+    const forks = `${info.forks} `;
+    addRow('Forks: ', forks, tbody);
+    const updated = `${info.updated_at} `;
+    addRow('Updated: ', updated, tbody);
+
+  }
+
+  function renderContributor(info, container) {
```
remarcmij: Since this function deals with rendering the contributors (note the plural), a more accurate name would be `renderContributors` (note the plural). A better name for the parameter `info` (a rather meaningless name) would be `contributors`.

homework/src/index.js
```diff
@@ -30,15 +34,96 @@
...
+          });
+
+          select.addEventListener('change', () => {
+            const repository = data[select.value];
+            fetchJSON(repository.contributors_url)
+              .then(contributors => {
+                renderRepository(repository, leftDiv);
+                renderContributor(contributors, rightDiv);
+              })
+              .catch(err => err.message);
```
remarcmij: This does not render the error to the web page. It doesn't do anything. You need to do something like is done in line 12.

homework/src/index.js
```diff
@@ -30,15 +34,96 @@
...
+          });
+
+
+          fetchJSON(data[0].contributors_url)
+            .then(contributors => {
+              renderRepository(data[0], leftDiv);
+              renderContributor(contributors, rightDiv);
+            })
+            .catch(err => err.message);
```
remarcmij: See my previous comment on rendering the error.

homework/src/index.js
```diff
@@ -30,15 +34,96 @@
...
+
+
+        });
```
remarcmij: Lines 101-107 and 110-118 are almost identical. Can you reduce this duplication by writing a function that you can instead of duplicating this code?

homework/src/index.js
```diff
@@ -30,15 +34,96 @@
...
+      li2.innerText = `${contributor.contributions}`;
+    });
+  }
+
   function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
+    fetchJSON(url)
+      .then(data => {
```
remarcmij: `data` is a meaningless name as I have mentioned several times in class. In this case, the best name is `repositories`.

PR #124, opened 2018-11-04
Week3

salar01:


salar01:
Hello Jim, thanks for your comments. I have a question. Would you please tell me how should i test my application? Because here i do not see eny problem. If i comment out //<script src="./index.js"></script>//
in file index.html then nothing is work. So because we have already distrubuted our code through several other files i do not know how should test this app. 
Thanks in advance.

remarcmij:
Hi @salar01, because you modified the classes files, I'm assuming the homework to review is that version. So I have opened the `classess.html` file (and not `index.html`) in the browser, because that file loads all the JavaScript files related to the classes version of the homework.

salar01:
Hello Jim, i got it that how should i tets my app. Thanks a lot, I corrected some codes, but still i get this error, Uncaught (in promise) TypeError: Cannot read property 'appendChild' of null
    at Function.createAndAppend (Util.js:7)
    at App.renderError (App.js:103)
    at App.fetchContributorsAndRender (App.js:93)

would you please help me.

remarcmij:
@salar01, you do not have an element with the **id** of `container` (line 75 of App.js). You **do** have an element with the **class** of `container`.

salar01:
Hello again, it is better now, but still i can not change repository select, and also contribution part is not showing. Thanks for you guidance. 💯

PR #126, opened 2018-11-10
Week3

Mohamad-AezAldin:
please find a homework of wk3 , sorry still need work for **redistribute** the code to other files


homework/src/index.js
```diff
@@ -1,19 +1,34 @@
...
+  const header = createAndAppend('header', root, { id: 'header' });
+  const h3 = createAndAppend('h3', header);
+  h3.innerText = 'HYF Repository';
+  const mainDiv = createAndAppend('div', root, { id: 'main' });
+  const repositoryInfo = createAndAppend('div', mainDiv, { id: 'repositoryInfo' });
+  const contributionsInfo = createAndAppend('div', mainDiv, { id: 'contributionsInfo' });
+  const ul = createAndAppend('ul', repositoryInfo, { id: 'list-container' });
+  const Contributions = createAndAppend('h3', contributionsInfo);
+  Contributions.innerText = 'Contributions';
+  const ul2 = createAndAppend('ul', contributionsInfo, { id: 'list-container' });
```
remarcmij: Lines 4-14 are potentially executed before the web page is fully loaded. For instance, if you move the `<script>` tag in the HTML file from the `<body>` to the `<head>` tag, you app no longer works, because the browser had not yet created the root `<div>`.

It is best to move this code to the beginning of your `main()` function (just before the `try` statement) to make sure that this code is executed **_after_** the page has full loaded.

Because this code is then no longer global to this file, you will need to pass some variables that are now global as parameters to the function that need them. This is a good thing because functions that receive all the information that they require through the argument list are easier to understand and debug, because they are more or less **_self-contained_**.

homework/src/index.js
```diff
@@ -30,18 +45,78 @@
...
+  async function main(url) {
+    try {
+      const repositories = await fetchJSON(url);
+      const select = createAndAppend('select', header);
+      repositories.sort((a, b) => a.name.localeCompare(b.name));
+      const sortedQueries = [];
+      repositories.forEach(repository => {
+        const item = { Repository: repository.name, Description: repository.description, Forks: repository.forks, Updated: repository.updated_at, ContributorsUrl: repository.contributors_url, RepositoryUrl: repository.html_url };
+        sortedQueries.push(item);
+      });
```
remarcmij: This is a comment I already made in week 2:

>There is no need to create a separate array (here: `sortedQueries`), you could use the repositories array instead. Your `sortedQueries` array is just a subset of fields of the repositories array. Nothing is gained from creating this extra array, it is just adding 'noise'.

Note also that object property names should follow the naming conventions for JavaScript variable names: camelCase.

homework/src/index.js
```diff
@@ -30,18 +45,78 @@
...
+        createAndAppend('option', select, { value: index, text: query.Repository });
+      });
+
+      renderRepositoryInfo(sortedQueries[0]);
+
+      renderContributorsInfo(sortedQueries[0]);
+
+      select.addEventListener('change', () => {
+
+        const queire = sortedQueries[select.value];
```
remarcmij: Mind the spelling error: `queire` -> `query`.

homework/src/index.js
```diff
@@ -30,18 +45,78 @@
...
+        sortedQueries.push(item);
+      });
+
+      sortedQueries.forEach((query, index) => {
+        createAndAppend('option', select, { value: index, text: query.Repository });
+      });
+
+      renderRepositoryInfo(sortedQueries[0]);
+
+      renderContributorsInfo(sortedQueries[0]);
```
remarcmij: The `renderContributorsInfo()` function really only requires the `contributors_url` property. It is best to give a function just the information that it needs and nothing more. This help to make your code more robust and less error prone. So you could call your function like this:

```js
renderContributorsInfo(repositories[0].contributors_url);
```

homework/src/index.js
```diff
@@ -1,19 +1,34 @@
...
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+
+  const root = document.getElementById('root');
+  const header = createAndAppend('header', root, { id: 'header' });
+  const h3 = createAndAppend('h3', header);
+  h3.innerText = 'HYF Repository';
```
remarcmij: You could set the text directly though the preceding call to `createAndAppend()`.

PR #127, opened 2018-11-10
Week3

Anashost:
Here handing the first part of week3 HW also contains some fixes to week2 HW , part 2 is a little bit complex and need more work.

### DIFF
--- a/homework/src/index.js
+++ b/homework/src/index.js
@@ -1,19 +1,23 @@
 'use strict';
 
 {
-  function fetchJSON(url, cb) {
-    const xhr = new XMLHttpRequest();
-    xhr.open('GET', url);
-    xhr.responseType = 'json';
-    xhr.onload = () => {
-      if (xhr.status < 400) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
+      const xhr = new XMLHttpRequest();
+      xhr.open('GET', url);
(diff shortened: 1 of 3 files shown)


homework/src/index.js
```diff
@@ -30,17 +34,104 @@
...
+    addRow('Updated: ', updated, tbody);
+
+  }
+
+  function renderContributors(info, container) {
+    container.innerHTML = '';
+    createAndAppend('h3', container, { text: 'Contributors' });
+    const ul = createAndAppend('ul', container);
+
+    info.forEach(contributor => {
```
remarcmij: If you rename the parameter `info` to `contributors`, this line could become more readable:

```js
```

homework/src/index.js
```diff
@@ -30,17 +34,104 @@
...
+      const li = createAndAppend('li', ul);
+      createAndAppend('img', li, { src: contributor.avatar_url });
+      const a = createAndAppend('a', li, { href: contributor.html_url, target: '_blank' });
+      a.innerText = contributor.login;
+      createAndAppend('p', li, { text: contributor.contributions });
```
remarcmij: The HTML you now generate is correct but it need CSS styling to make align it horizontally. I would recommend to try and use CSS flexbox for that.

homework/src/index.js
```diff
@@ -30,17 +34,104 @@
...
+          renderContributors(contributors, rightDiv);
+        }
+        catch (err) {
+          const root = document.getElementById('root');
+          createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        }
+
+      });
+
+      try {
```
remarcmij: This inner `try` is unnecessary (and therefore unwanted) as the outer `try` block that starts in line 86 already deals with catching errors. If you remove lines 121 and 125-129 then any error that could occur in line 123 (for instance, a rejected promise) would be could by the `catch` block of lines 132-135.

PR #132, opened 2018-11-18
homwwork for js3 week1 is over

bless4:
good work


gijscor:
@bless4 this is the Repo for HYF the Netherlands, please hand it in to the github of HYF Sweden :)

PR #136, opened 2018-12-02
first commit

bless4:
JS3 week2 is an API call with promises from github


remarcmij:
@bless4 This is the repo for HYF the Netherlands, please hand it in to the GitHub of HYF Sweden 🙂.

PR #139, opened 2018-12-07
added homework week1

Wael-Alhomsi:


homework/index.js
```diff
@@ -30,13 +30,127 @@
     return elem;
   }
 
+  function buildRepositoryInfo(repositoryObject, containerDiv) {
+    const repositoryTable = createAndAppend('table', containerDiv, {
+      id: 'repository-table',
+    });
+    const repositoryTableBody = createAndAppend('tbody', repositoryTable, {});
```
remarcmij: If you don't want to set any attributes with the third parameter of `createAndAppend` then there is no need to specify an empty object for this parameter. The default for this parameter is already an empty object:

```js
function createAndAppend(name, containerDiv, options = {}) {
```

The ES6 syntax `options = {}` means that the default for the `options` parameter if not specified is (in this case) an empty object.

homework/index.js
```diff
@@ -30,13 +30,127 @@
...
+    createAndAppend('td', repositoryTableRow2, { text: repositoryObject.description });
+    const repositoryTableRow3 = createAndAppend('tr', repositoryTableBody, {});
+    createAndAppend('th', repositoryTableRow3, { text: 'Forks:' });
+    createAndAppend('td', repositoryTableRow3, { text: repositoryObject.forks_count });
+    const repositoryTableRow4 = createAndAppend('tr', repositoryTableBody, {});
+    createAndAppend('th', repositoryTableRow4, { text: 'Last updated:' });
+    createAndAppend('td', repositoryTableRow4, { text: repositoryObject.updated_at });
+  }
+
+  function buildRepositoryContributors(contributorsArray, containerDiv) {
```
remarcmij: Because you use bracket notation inside your function it is already clear the first argument is an array, so I would leave out the suffix `Array`.

homework/index.js
```diff
@@ -30,13 +30,127 @@
...
+    createAndAppend('h2', containerDiv, { text: 'Contributions' });
+    const contributorsList = createAndAppend('ul', containerDiv, { id: 'contributors-list' });
+    contributorsArray.forEach(element => {
```
remarcmij: If the function argument was called `repositories` then the `forEach` loop could be written as:

```js
```

Try and avoid generic names such as `item`, `element` if you have precise knowledge what the variable or parameter contains.

homework/index.js
```diff
@@ -30,13 +30,127 @@
...
+    createAndAppend('th', repositoryTableRow3, { text: 'Forks:' });
+    createAndAppend('td', repositoryTableRow3, { text: repositoryObject.forks_count });
+    const repositoryTableRow4 = createAndAppend('tr', repositoryTableBody, {});
+    createAndAppend('th', repositoryTableRow4, { text: 'Last updated:' });
+    createAndAppend('td', repositoryTableRow4, { text: repositoryObject.updated_at });
+  }
+
+  function buildRepositoryContributors(contributorsArray, containerDiv) {
+    createAndAppend('h2', containerDiv, { text: 'Contributions' });
+    const contributorsList = createAndAppend('ul', containerDiv, { id: 'contributors-list' });
```
remarcmij: As noted before, this function is solely concerned with contributors, so prefixing all local variables names with the word `contributor` may be a bit overkill.

homework/index.js
```diff
@@ -30,13 +30,127 @@
...
+  function buildRepositoryInfoAndFetchContributors(
+    repositoryObject,
+    infoContainer,
+    contributorsContainer,
+  ) {
+    buildRepositoryInfo(repositoryObject, infoContainer);
+    fetchRepositoryContributors(repositoryObject.contributors_url, contributorsContainer);
+  }
+
+  function startUpAndBuildSelectList(arrayOfRepositories, divContainer) {
```
remarcmij: This first argument could simply be named `repositories`.

homework/index.js
```diff
@@ -30,13 +30,127 @@
...
+        {},
+      );
+      createAndAppend('h3', contributorNumberOfContributionsContainer, {
+        text: element.contributions,
+      });
+    });
+  }
+
+  function fetchRepositoryContributors(url, containerDiv) {
+    fetchJSON(url, (err, data) => {
```
remarcmij: Since we know that we are getting a list of contributors, I would suggest to rename the parameter `data` to `contributors`.

homework/index.js
```diff
@@ -30,13 +30,127 @@
...
+      buildRepositoryInfoAndFetchContributors(
+        arrayOfRepositories[event.target.value],
+        repositoryInfoSection,
+        rightContainer,
+      );
+    });
+  }
+
   function main(url) {
     fetchJSON(url, (err, data) => {
```
remarcmij: The data we get is a list of repositories, so you could rename `data` to `repositories`.

homework/index.js
```diff
@@ -30,13 +30,127 @@
...
+    });
+    const repositoryTableRow2 = createAndAppend('tr', repositoryTableBody, {});
+    createAndAppend('th', repositoryTableRow2, { text: 'Description:' });
+    createAndAppend('td', repositoryTableRow2, { text: repositoryObject.description });
+    const repositoryTableRow3 = createAndAppend('tr', repositoryTableBody, {});
+    createAndAppend('th', repositoryTableRow3, { text: 'Forks:' });
+    createAndAppend('td', repositoryTableRow3, { text: repositoryObject.forks_count });
+    const repositoryTableRow4 = createAndAppend('tr', repositoryTableBody, {});
+    createAndAppend('th', repositoryTableRow4, { text: 'Last updated:' });
+    createAndAppend('td', repositoryTableRow4, { text: repositoryObject.updated_at });
```
remarcmij: 1. There is some code repetition here, that could be extracted into a function, say, named `addRow`.
2. The date format presented to the user could be a bit more user friendly, by using this method on new Date(repository.updated_at): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

homework/index.js
```diff
@@ -16,9 +16,9 @@
     xhr.send();
   }
 
-  function createAndAppend(name, parent, options = {}) {
+  function createAndAppend(name, containerDiv, options = {}) {
```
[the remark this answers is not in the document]
Wael-Alhomsi: Yes that's totally true. It has been changed unintentionally after a global search-and-replace operation. The next time I'll pay more attention while using it.

Wael-Alhomsi:
Dear @remarcmij ,
Thank you for your in-depth review, I'll read it thoroughly and update the code accordingly.
All the best!
Wael

homework/index.js
```diff
@@ -30,13 +30,163 @@
...
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+      hour: '2-digit',
+      minute: '2-digit',
+    };
+    return date.toLocaleDateString('en-GB', options);
+  }
+
+  function addRowToTable(parent, options = []) {
```
remarcmij: This is a good attempt at making your code DRY (_Don't Repeat Yourself_ principle), however the implementation is overly complicated because you are adding support to cover the exceptional case of the first row, which requires a link instead of a text string. During the lecture I showed how you can write a simplified version, generic version and deal with exception elsewhere.

PR #140, opened 2018-12-07
Zerisenay Solomon  JS3 Week1 HW-class 18 

ZerSolo:
sorry for the late submission.
 The application works well but I am not satisfied with the code that displays the first repository  on start-up.  I repeat similar codes.  I tried to make a function and use it  by calling it but was not successful. When I made a function.


homework/index.js
```diff
@@ -27,21 +27,163 @@
...
+    const forthRow = createAndAppend('tr', tbody, {
+      tabindex: '0',
+    });
+    createAndAppend('td', forthRow, {
+      class: 'label',
+      text: 'Updated :',
+    });
+    createAndAppend('td', forthRow, {
+      text: repository.updated_at,
+    });
```
remarcmij: You have some repeating code here: create a `tr`, add `a `td` and add another `td`. Could you think of function that you could write, called perhaps `addRow` to eliminate this code duplication?

homework/index.js
```diff
@@ -27,21 +27,163 @@
...
+    const forthRow = createAndAppend('tr', tbody, {
+      tabindex: '0',
+    });
+    createAndAppend('td', forthRow, {
+      class: 'label',
+      text: 'Updated :',
+    });
+    createAndAppend('td', forthRow, {
+      text: repository.updated_at,
```
remarcmij: The format of the displayed date could be improved, for instance by using this method on `new Date(repository.updated_at)`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

homework/index.js
```diff
@@ -27,21 +27,163 @@
...
+    });
+  }
+
+  function renderRepoContributors(contributors, rightContainer) {
+    createAndAppend('p', rightContainer, {
+      class: 'contributor-header',
+      text: 'Contributors',
+    });
+    contributors.forEach(contributor => {
+      const ul = createAndAppend('ul', rightContainer, {
```
remarcmij: You are creating one `ul` per contributor. However, a single `ul` containing the list of contributors would make more sense (also from an ARIA perspective). Then each contributor could be a single `li` and within that `li` you could have whatever elements are needed to render the contributor details.

homework/index.js
```diff
@@ -27,21 +27,163 @@
...
+      tabindex: '0',
+    });
+    createAndAppend('td', firstRow, {
+      class: 'label',
+      text: 'Repository : ',
+    });
+    const td2 = createAndAppend('td', firstRow);
+    createAndAppend('a', td2, {
+      text: repository.name,
+      href: repository.html_url,
```
remarcmij: You could add another property: `target: '_blank'` to open the link in a new browser tab.

homework/index.js
```diff
@@ -27,21 +27,163 @@
...
+    });
+    const firstRepository = repositories[0];
+    renderRepositoryInfo(firstRepository, leftContainer);
+    fetchContributors(firstRepository.contributors_url, rightContainer);
+
+    selectMenu.addEventListener('change', () => {
+      leftContainer.innerHTML = '';
+      renderRepositoryInfo(repositories[event.target.value], leftContainer);
+      rightContainer.innerHTML = '';
+      fetchContributors(repositories[event.target.value].contributors_url, rightContainer);
```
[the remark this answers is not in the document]
remarcmij: Hi @ZerSolo, you should not mark these comments as resolved if you don't fix the issue that it addresses (and you haven't done that).

PR #142, opened 2018-12-08
mohamad-js3-wk1

mohamad-monir:


mohamad-monir:
not ready yet but in progress

homework/index.js
```diff
@@ -30,18 +30,94 @@
     return elem;
   }
 
+  function render(parent, repositories) {
```
Jawhar-B: This may have the name of a singuler instead of a plural form while it's an object of a one repository, what do you say? :)

homework/index.js
```diff
@@ -30,18 +30,94 @@
...
+    const table = createAndAppend('table', parent, {});
+    const tbody = createAndAppend('tbody', table, {});
```
Jawhar-B: If no need for passing anything no need for an extra Braces while it's assumed by default on line 19

homework/index.js
```diff
@@ -30,18 +30,94 @@
...
+    const tr = createAndAppend('tr', tbody);
+    createAndAppend('td', tr, { class: 'label', text: 'Repository : ' });
+    const td = createAndAppend('td', tr);
+    createAndAppend('a', td, {
+      text: repositories.name,
+      href: 'https://github.com/HackYourFuture/alumni',
```
Jawhar-B: Imagine the situation when the repositoy name changes, the link going to be broken, right?
To fix this, I think you already did it in line 69:
```suggestion
      href: repositry.html_url, // assuming you applied the change on the singuler naming form.
```

homework/index.js
```diff
@@ -30,18 +30,94 @@
...
+    });
+    const tr3 = createAndAppend('tr', tbody);
+    createAndAppend('td', tr3, { text: 'Updated :', class: 'label' });
+    createAndAppend('td', tr3, { text: repositories.updated_at });
+  }
+
+  function listOfContributors(contributors, parent) {
+    createAndAppend('h3', parent, { class: ' contributor - header', text: 'contributions' });
+    const ul = createAndAppend('ul', parent, { class: 'contributor-list' });
+    contributors.forEach(x => {
```
Jawhar-B: `x` 🤔.
Having a good names, having a good context, as much your names are readable as much your code is maintainable.

homework/index.js
```diff
@@ -30,18 +30,94 @@
...
+      });
+    });
+  }
+
+  function fetchRepositoryContributors(url, parent) {
+    fetchJSON(url, (err, repositories) => {
+      if (err) {
+        createAndAppend('h2', parent, { text: err.message, class: 'alert-error' });
+      } else {
+        listOfContributors(repositories, parent);
```
Jawhar-B: What about: `listContributors` so it have the name of the actual action 🙌

Meno-101:
Now Naming looks great 🙂, I saw you fixed some styling in the second week homework 😃 
sorry being late on re-reviewing it.

PR #143, opened 2018-12-08
Week1 homework

FMatHYF:


homework/index.js
```diff
@@ -30,13 +30,93 @@
     return elem;
   }
 
+  function renderRepository(leftContainer, repository) {
+    const tabel = createAndAppend('tabel', leftContainer, { id: 'tabel' });
+    const tr1 = createAndAppend('tr', tabel, { class: 'descriptions' });
+
+    createAndAppend('td', tr1, { class: 'col1_data', text: 'Repository:' });
+    createAndAppend('td', tr1, { class: 'col2_data', text: repository.name });
```
alexandrazimnicaru: The renderRepository function does its job well, but you can see there's a repeating structure here. You could improve this by: 
- extracting the repeating lines (from lines 35 to 38) into a separate function that renders a row, let's call it renderRow
- looping through an array of rows you want to render and calling renderRow, like this:
```
const rows = [
  { title: 'Repository', content: repository.name },
  { title: 'Description', content: repository.description },
  { title: 'Forks', content: repository.forks },
  { title: 'Updated', content: repository.updated_at }
];

rows.forEach((row) => {
  renderTableRow(tabel, row.title, row.content);
})
```
Also, I would suggest to use a different html structure than a table to be able to make this responsive. Maybe just some divs would work, which you could arrange in a layout using flex: https://css-tricks.com/snippets/css/a-guide-to-flexbox/.

homework/index.js
```diff
@@ -30,13 +30,93 @@
...
+
+    const tr3 = createAndAppend('tr', tabel, { class: 'descriptions' });
+
+    createAndAppend('td', tr3, { class: 'col1_data', text: 'Forks:' });
+    createAndAppend('td', tr3, { class: 'col2_data', text: repository.forks });
+
+    const tr4 = createAndAppend('tr', tabel, { class: 'descriptions' });
+
+    createAndAppend('td', tr4, { class: 'col1_data', text: 'Updated:' });
+    createAndAppend('td', tr4, { class: 'col2_data', text: repository.updated_at });
```
alexandrazimnicaru: This doesn't seem to be a requirement, but as an improvement, there is a way to format the date you're getting from the API and make it more human readable, by using this method on new Date(repository.updated_at): https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleDateString

homework/index.js
```diff
@@ -30,13 +30,93 @@
...
+  }
+
+  function renderContributors(rightContainer, url) {
+    fetchJSON(url, (err, contributors) => {
+      if (err !== null) {
+        createAndAppend('div', rightContainer, { text: err.message, class: 'alert-error' });
+      } else {
+        createAndAppend('p', rightContainer, { class: 'contributions', text: 'Contributions' });
+        const ul = createAndAppend('ul', rightContainer, { class: 'listOfcontributors' });
+        contributors.forEach(contributor => {
```
alexandrazimnicaru: The renderContributors looks good, the only thing I would improve here is extract the code inside the forEach into a separate function that could be called renderContributor, just to make renderContributors shorter and easier to read.

homework/index.js
```diff
@@ -30,13 +30,93 @@
...
+    const root = document.getElementById('root');
+    fetchJSON(url, (err, repositories) => {
+      if (err !== null) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        const header = createAndAppend('header', root, { id: 'header', text: 'HYF Repositories' });
+
+        const select = createAndAppend('select', header, { id: 'repo_options' });
+        repositories.sort((a, b) => {
```
alexandrazimnicaru: The main function also looks good, but the sorting function here could also be extracted into a separate function in order to simplify the code.

PR #144, opened 2018-12-08
i'm still working on it 

soumaiamamo:


homework/index.js
```diff
@@ -21,27 +21,117 @@
...
+        class: 'label',
+      });
+      createAndAppend('td', description, { html: repository.description });
+    }
+    const forks = createAndAppend('tr', tableBody);
+    createAndAppend('td', forks, { html: 'Forks:', class: 'label' });
+    createAndAppend('td', forks, { html: repository.forks });
+    const update = createAndAppend('tr', tableBody);
+    createAndAppend('td', update, { html: 'Updated:', class: 'label' });
+    createAndAppend('td', update, { html: repository.updated_at });
```
remarcmij: The date as displayed is not very user-friendly. You could research how to convert `updated_at` from a string to a `Date` and then use `.toLocalString()` to convert it to a user-friendly format.

homework/index.js
```diff
@@ -21,27 +21,117 @@
...
+    });
+    fetchJSON(repository.contributors_url, (err, contributors) => {
+      if (err) {
+        createAndAppend('div', rightDiv, {
+          html: err.message,
+          class: 'alert-error',
+        });
+      }
+
+      contributors.map(contributor => {
```
remarcmij: You are using a `.map()` array method here but you are not 'mapping' anything. You should use the `.map()` method if you want to create a new array from an existing array, while transforming in some way each element of the existing array.

In this case you are just looping through the array, so you should just use `.forEach()` here instead of `.map()`. This will also get rid of one of the errors reported by ESLint.

homework/index.js
```diff
@@ -21,27 +21,117 @@
...
+        createAndAppend('div', contributorDetails, {
+          html: contributor.login,
+          class: 'contributor-names',
+        });
+        createAndAppend('div', contributorDetails, {
+          html: contributor.contributions,
+          class: 'contributors',
+        });
+      });
+    });
```
remarcmij: You are generating one `ul` per contributor. Instead, there should be one `ul` for the whole contributor list and an `li` per contributor:

```html
<ul>
  <li>details of first contributor</li>
  <li>details of second contributor</li>
</ul>
```

This is what you generate:

```html
<div class="right_div">
  <p class="contributor-head">Contributions</p>
  <ul class="contributor-ul">
    <li class="contributor-li">
      <a target="_blank" href="https://github.com/gijscor" >
        <img
          class="contributor-image"
        />
        <div class="contributor-details">
          <div class="contributor-names">gijscor</div>
          <div class="contributors">47</div>
        </div>
      </a>
    </li>
  </ul>
  <ul class="contributor-ul">
    <li class="contributor-li">
      <a target="_blank" href="https://github.com/wouterkleijn" >
        <img
          class="contributor-image"
        />
        <div class="contributor-details">
          <div class="contributor-names">wouterkleijn</div>
          <div class="contributors">8</div>
        </div>
      </a>
    </li>
  </ul>
</div>
```

homework/index.js
```diff
@@ -21,27 +21,117 @@
...
     fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        createAndAppend('div', root, {
+          html: err.message,
+          class: 'alert-error',
+        });
```
remarcmij: If there is an error you should not try and execute the rest of the code because there is no data to work on. You should add a `return` statement at the end of this `if` block to return 'early' from this function. 

This is what I see in the console when I and run your code and introduce an error in the url:

```
index.js:124 Uncaught TypeError: Cannot read property 'sort' of undefined
    at fetchJSON (index.js:124)
    at XMLHttpRequest.xhr.onload (index.js:12)
```

homework/index.js
```diff
@@ -21,27 +21,117 @@
...
+      html: 'Contributions',
+      class: 'contributor-head',
+    });
+    fetchJSON(repository.contributors_url, (err, contributors) => {
+      if (err) {
+        createAndAppend('div', rightDiv, {
+          html: err.message,
+          class: 'alert-error',
+        });
+      }
```
remarcmij: When there is an error you should not continue with the rest of the code because there is no data to work on. Instead, you should add a `return` statement to and the end of this `if` block to return early from this callback function.

homework/index.js
```diff
@@ -21,27 +21,117 @@
...
+    const header = createAndAppend('header', root, { class: 'header' });
+    createAndAppend('p', header, { html: 'HYF Repositories', class: 'hyf' });
+    const select = createAndAppend('select', header, {
+      class: 'select',
+      id: 'list',
+    });
+    const container = createAndAppend('div', root, { class: 'container' });
+    const left = createAndAppend('div', container, { class: 'left_div' });
+    const right = createAndAppend('div', container, { class: 'right_div' });
     fetchJSON(url, (err, data) => {
```
remarcmij: Since we know that the data we are getting is a list of repositories a more descriptive name for the `data` parameter would be `repositories`. This improves the readability of your code significantly (more so even, if you replace `val` by `repository in line 125:

```js
  if (err) {
      html: err.message,
      class: 'alert-error',
    });
    return;
  }

  repositories.map((repository, index) =>
    createAndAppend('option', select, { value: index, html: repository.name }),
  );
  select.addEventListener('change', event => {
    const index = event.target.value;
    showRepos(left, repositories[index]);
    showContributors(right, repositories[index]);
  });
  showRepos(left, repositories[0]);
  showContributors(right, repositories[0]);
});
```

homework/index.js
```diff
@@ -21,27 +21,117 @@
...
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        createAndAppend('div', root, {
+          html: err.message,
+          class: 'alert-error',
+        });
       }
+
+      data.sort((a, b) => a.name.localeCompare(b.name));
+      data.map((val, index) => createAndAppend('option', select, { value: index, html: val.name }));
+      document.getElementById('list').addEventListener('change', event => {
```
remarcmij: There is no need to use `getElementById` to get reference to the `select` element. You already have a reference in the variable `select` that you created in line 109. So you could simply write:

```js
select.addEventListener('change', event => {
```

homework/style.css
```diff
@@ -1,3 +1,112 @@
 .alert-error {
```
remarcmij: Your application is unusable on a mobile phone. This is what it looks like.

screenshot 2018-12-14 at 13 47 04

soumaiamamo:
@remarcmij 

thanks alot for your feedback , 


for innerHTML , i used in my function inner HTML then nothing show up in my page i thought it is a mistake so i tried to change it  in (createAndAppend )function from innerText to innerHTML then it works but now i changed every innerHTML to innerText and every thing works.

 

i solved those problems in week 2 homework but i still have a problem in my (npm test) but i could not solve it.
(  73:38  error  Expected to return a value in arrow function  array-callback-return) 


شكراً جزيلاً

remarcmij:
> i solved those problems in week 2 homework but i still have a problem in my (npm test) but i could not solve it.
( 73:38 error Expected to return a value in arrow function array-callback-return)

This problem is caused because you have used a `.map()` (which returns an array that ESLint expects you to use) where you should have used a  `.forEach()` as I commented here:

https://github.com/HackYourFuture/JavaScript3/pull/144#discussion_r241737663

For the other ESLint errors you should implement the changes to `.eslintrc.json` as explained here:

https://hackyourfuture.slack.com/archives/CCYC038CX/p1544375894094500

PR #146, opened 2018-12-09
Week1

Kani103:


homework/index.js
```diff
@@ -30,18 +30,96 @@
     return elem;
   }
 
+  function drawRepository(data, value, container) {
+    // Clear the details of the selected option
```
Jawhar-B: Nice!
you have comments clearifing steps 🥇

homework/index.js
```diff
@@ -30,18 +30,96 @@
...
+    drawRepositoryInfo(data, value, container);
+    drawRepositoryContributors(data, value, container);
+  }
+
+  function drawRepositoryInfo(data, value, container) {
+    const leftDiv = createAndAppend('div', container, { class: 'left-div' });
+    const repoTable = createAndAppend('table', leftDiv);
+    const repoTbody = createAndAppend('tbody', repoTable);
+
+    const tr1 = createAndAppend('tr', repoTbody);
```
Jawhar-B: This is only a styling thing, I would prefere to have the names instead of `tr1`, `tr2` ...etc, names like `repoNameRow`, `repoDescriptionRow` ...etc
it looks more clear for me, what do you say 🙂

homework/index.js
```diff
@@ -30,18 +30,96 @@
...
+        const div = createAndAppend('div', li, { class: 'contributor-data' });
+        createAndAppend('div', div, { text: entry.login });
+        createAndAppend('div', div, { text: entry.contributions });
+      })
+    });
+  }
+  // function rightContainer(url, data) {
+  //   const rightDiv = document.getElementsByClassName('right-container');
+  //   rightDiv.innerHTML = '';
+  //   // const paragraph = createAndAppend('p', rightDiv, { text: 'contributions' }, { class: 'contibutors-header' });
```
Jawhar-B: Perhaps leaving the comments is good while a code commented out will make he file looks larger and a bit hard to read.
So you may leave it out.

homework/index.js
```diff
@@ -30,18 +30,96 @@
...
+  //   // const unlist = createAndAppend(ul, paragraph, { class: 'contributors-list' });
+  // }
+
   function main(url) {
     fetchJSON(url, (err, data) => {
       const root = document.getElementById('root');
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
```
Jawhar-B: `if`, `else` statement is always important.
But the thing sometimes they gets really long in order to know where each one ends, to solve that we may have a look to the [early return](http://blog.timoxley.com/post/47041269194/avoid-else-return-early)

homework/index.js
```diff
@@ -30,18 +30,96 @@
...
+  }
+  // function rightContainer(url, data) {
+  //   const rightDiv = document.getElementsByClassName('right-container');
+  //   rightDiv.innerHTML = '';
+  //   // const paragraph = createAndAppend('p', rightDiv, { text: 'contributions' }, { class: 'contibutors-header' });
+  //   // const unlist = createAndAppend(ul, paragraph, { class: 'contributors-list' });
+  // }
+
   function main(url) {
     fetchJSON(url, (err, data) => {
```
Jawhar-B: We are receiving repositories here, so in order to make it a bit explicet we may call it as what it contains 😃 
```suggestion
    fetchJSON(url, (err, repositries) => {
```

homework/index.js
```diff
@@ -30,18 +30,96 @@
...
+          if (x > y) {
+            return 1;
+          }
+          return 0;
+        });
+
+        const header = createAndAppend('header', root, { class: 'header' });
+        const selectElem = createAndAppend('select', header, { class: 'repo-selector' });
+        // for each entry in the data, add option elements
+        data.forEach((entry, index) => {
```
Jawhar-B: Now here if you followed my last suggestion, you will find here you are just writing english, as the following:

```suggestion
        reposories.forEach((repository, index) => {
```

PR #147, opened 2018-12-14
added homework week2

Wael-Alhomsi:


homework/index.js
```diff
@@ -30,17 +32,161 @@
...
+    const date = new Date(string);
+    const options = {
+      weekday: 'long',
+      year: 'numeric',
+      month: 'long',
+      day: 'numeric',
+      hour: '2-digit',
+      minute: '2-digit',
+    };
+    return date.toLocaleDateString('en-GB', options);
```
remarcmij: Rather than specifying exactly how the date should be formatted you can also (and in many cases, preferably) rely on the computer default locale. This is what the user has set up in the computer settings. For instance, on my computer I have defined a Dutch locale, with Dutch rules for date and time. A US user might expect a US date and time format. If you use `.toLocaleDateString()` the defaults will be applied and your code is simpler too.

Wael-Alhomsi:
Dear @remarcmij !
Like always, I would like to thank you for your extensive reviews, and for the time you spend to give us these valuable in-depth notes. 
Thank you very much :)
Wael

PR #148, opened 2018-12-14
Zerisenay Solomon class 18 HW Week2

ZerSolo:
According to the previous feedback you gave me, I couldn't manage to create a function for adding a row in a table.  I will try me to do it ASAP.


Week1/example.html
```diff
@@ -0,0 +1,122 @@
+<html lang="en" class="gr__hyf-github_netlify_com">
```
remarcmij: This file should not be part of a PR. You should remove it from your local repository.

homework/index.js
```diff
@@ -27,21 +31,145 @@
...
 
-  function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
-      }
+  function addRow(parent, labelText, rowValue) {
```
remarcmij: I see you have implemented the `addRow` function I asked for. You asked for a hint and I was planning to demo it during the lecture. Good that you picked it up from the lecture.

homework/index.js
```diff
@@ -27,21 +31,145 @@
...
+      });
+    });
+  }
+
+  function fetchContributors(url, rightContainer) {
+    const fetchedContributors = fetchJSON(url);
+    fetchedContributors
+      .then(contributors => {
+        renderRepoContributors(contributors, rightContainer);
+        return fetchedContributors;
```
remarcmij: You do not need to return anything here. What you return from a `.then()` method is passed as a parameter to the next `.then()` in the chain. But there is no next `.then()`, only a `catch().`

homework/index.js
```diff
@@ -27,21 +31,145 @@
...
   }
 
+  function main(url) {
+    const promise = fetchJSON(url);
+    const root = document.getElementById('root');
+
+    promise
+      .then(parsedResponse => {
+        renderRepoInfoAndContributorsOnstartup(parsedResponse, root);
+        return promise;
```
remarcmij: No need to return anything here. See my previous comment.

ZerSolo:
thank you, Jim, for your feedback. I appreciate it a lot.  According to t your feedback, I have corrected them.

PR #154, opened 2018-12-16
Gaffari js3wk2hw

Gaffari02:


homework/index.js
```diff
@@ -30,17 +32,100 @@
...
+  }
+
+  function main(url) {
+    const root = document.getElementById('root');
+    const headBox = createAndAppend('div', root, { class: 'header' });
+    createAndAppend('p', headBox, { text: 'HYF Repositories', id: 'headTitle' });
+    const selectBox = createAndAppend('select', headBox, { class: 'selectBox' });
+
+    fetchJSON(url)
+      .catch(err => {
```
remarcmij: The `.catch()` method should normally be at the end of the chain. If you place at the beginning of the chain the error is caught, however after that, execution continues with the next `.then()` which should nor run in case of an error. Consequently, if I force an error by making the url invalid I can see this error message in the browser console:

```
Uncaught (in promise) TypeError: Cannot read property 'forEach' of undefined
    at fetchJSON.catch.then.repositories (index.js:112)
```

homework/index.js
```diff
@@ -30,17 +32,100 @@
...
+        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+      })
+      .then(repositories => {
+        repositories.forEach((repository, index) => {
```
remarcmij: I can't remember whether I already gave the comment in week 1 but the assignment asks for the repositories to be sorted.

homework/style.css
```diff
@@ -1,3 +1,103 @@
+body {
```
remarcmij: Your responsive CSS still needs work:

screenshot 2018-12-20 at 15 09 18

screenshot 2018-12-20 at 15 09 37

screenshot 2018-12-20 at 15 10 27

Gaffari02:
Hi @remarcmij I have fixed the promise chains. I will keep trying for better css in 3rd week homework. I also sort the repositories in next part of homework.
I wish you a happy new year:)

PR #156, opened 2018-12-21
Zerisenay Solomon class 18 HW JS3 Week3

ZerSolo:


homework/index.js
```diff
@@ -27,21 +12,151 @@
...
     });
   }
 
+  async function fetchContributors(url, rightContainer) {
+    try {
+      const fetchedContributors = await fetchJSON(url);
+      renderRepoContributors(fetchedContributors, rightContainer);
+    } catch (error) {
+      createAndAppend('div', rightContainer, {
+        text: `We got an error: $ {error.message}`,
```
Meno-101: Oops! 
This won't work 😅
```suggestion
        text: `We got an error: ${error.message}`,
```

homework/index.js
```diff
@@ -27,21 +12,151 @@
...
+      rightContainer.innerHTML = '';
+      fetchContributors(repositories[event.target.value].contributors_url, rightContainer);
+    });
+  }
+
+  function main(url) {
+    async function fetchAllData() {
+      const root = document.getElementById('root');
+      const parsedResponse = await fetchJSON(url);
+      renderRepoInfoAndContributorsOnstartup(parsedResponse, root);
```
Meno-101: When I made a network error, I got this back:
*(I removed the `u` from github url)*

screen shot 2018-12-21 at 23 00 34

what about `try {...} catch(error) {...}` blocks here ?

-------
As a taste: 😃
In my opinion `renderRepoInfoAndContributorsOnstartup` this is a bit long name as a one line,
so what about `initRepoInfoAndContributors` or may be while we are not using it for more than a one time `renderRepoInfoAndContributors`, still long but shorter 😃

homework/index.js
```diff
@@ -27,21 +12,151 @@
...
+    }
+    fetchAllData();
```
Meno-101: A function as you know made to be able re-use, So what about passing the url from here? and moving the function outside of the main making it more specific with it's doings ?

PR #159, opened 2018-12-22
Class18 js3 wk3

mohamad-monir:


homework/index.js
```diff
@@ -1,47 +1,136 @@
...
+function renderError(error) {
+  const h2 = document.createElement('h2');
+  document.querySelector('#root').appendChild(h2);
+  h2.innerHTML = error;
+}
+async function fetchJSON(url) {
+  try {
+    const response = await fetch(url);
+    const json = await response.json();
+    return json;
```
remarcmij: Excellent that you have replaced `XMLHttpRequest` with the newer `fetch()` function that directly works with promises. You could combine lines 11 & 12 into a single `return response.json()`. Explanation: A function marked with the `async` keyword always returns a promise. When you call `fetchJSON()`, for instance in line 121 of function `main()` you are already `await`-ing the result. There is no need to `await` it in `fetchSJON()` itself. Just return the promise from `response.json()` and the `await` done by the caller of `fetchJSON()` will `await` its result.

homework/index.js
```diff
@@ -1,47 +1,136 @@
...
-  function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
-      }
+function listContributors(contributors, parent) {
+  createAndAppend('h3', parent, { class: ' contributor - header', text: 'contributions' });
```
remarcmij: There are extraneous spaces in the `class` property value, which probably causes the class not to work. But you don't seem to use this `contributor-header` class in your `style.css` anyway, so best to remove it.

homework/index.js
```diff
@@ -1,47 +1,136 @@
...
+    );
+  });
+}
 
-  window.onload = () => main(HYF_REPOS_URL);
+async function main(url) {
+  const root = document.getElementById('root');
+  try {
+    let data = await fetchJSON(url);
+    let render = renderTableAndContributorsInfo(data, root);
```
remarcmij: These two `let` statements should be `const`. You should be getting an ESLint warning about that and a red underline in VSCode. Because the ESLint check performed when you submit your PR automatically fixes these style errors, you code still passes the test. However, you should strive to make you code correct without it needing to be fixed automatically by ESLint.

homework/index.js
```diff
@@ -1,47 +1,136 @@
...
+    return renderError(err.name);
+  }
+}
 
-  const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+function renderTableAndContributorsInfo(repositories, parent) {
+  const infoDiv = createAndAppend('div', parent, { class: 'tableInfo' });
+  const flexContainer = createAndAppend('div', parent, { class: 'flexContainer' })
+  const repoInfo = createAndAppend('div', flexContainer, { class: 'left-div whiteframe' });
+  const mainDiv = createAndAppend('div', repoInfo, { id: 'container' });
```
remarcmij: You do no seem to have a `whiteframe` CSS style defined in your `style.css`. It looks like you peeked at the HTML generated by https://hyf-github.netlify.com/ where this class is used. I have no issue with a bit of  'creative stealing' like this (as long as the bulk of the works is your own), but you should clean up unused bits.

mohamad-monir: good morning Jim 
thank you for the feedback 
I will definitely check my Vscode  ESlint setting ..
I will also check the unused classes or extra unnecessary things in my code ...
I did look at the website and  structural of the html coz i thought it is require from us to copy this website thats why there is in the homework page link to the site .... but be sure the css style it is completely mine and had diffcult time to style responsive web site coz i start with the big screen and not mobile screen like suppose to be ...
another time thanx for your feedback so i can improve myself everytime and i will for sure try the classes ...

PR #160, opened 2018-12-22
Yaacob-Js3-Week3

SignosRk:


homework/App.js
```diff
@@ -7,58 +7,50 @@ class App {
...
+    const select = Util.createAndAppend('select', header, {
+      class: 'selectBox',
+      'aria-label': 'HYF Repositories',
+    });
+    Util.createAndAppend('div', root, { id: 'container' });
     try {
-      const repos = await Util.fetchJSON(url);
-      this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      const fetch = await Util.fetchJSON(url);
```
remarcmij: This name `fetch` confused me for the following reasons:

1. `fetch` is also the name of the standard browser function `fetch()`: this is a replacement for `XMLHttpRequest` and directly works with promises.
2. The name is a verb and verbs should be reserved for functions (functions perform actions). Variables should use nouns (they reference _things_).
3. The name does not reflect the contents: an array of repository objects.

In this case even the name `data` or `repositoryData` would be better than `fetch`.

homework/App.js
```diff
@@ -7,58 +7,50 @@ class App {
...
+      select.addEventListener('change', () => this.fetchContributorsAndRender(select.value));
+      Util.createAndAppend('div', root, { id: 'container' });
+      this.repositories = fetch
+        .sort((a, b) => a.name.localeCompare(b.name))
+        .map(repo => new Repository(repo));
+      this.repositories.forEach((a, b) => {
```
remarcmij: The names `a` and `b` are meaningless. Your code would be easier to understand if you would use more descriptive names, e.g.:

```js
this.repositories.forEach((repository, index) => {
```

homework/Contributor.js
```diff
@@ -5,15 +5,28 @@
...
-   * Render the contributor info to the DOM.
-   * @param {HTMLElement} container The container element in which to render the contributor.
-   */
   render(container) {
-    // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.contributor, null, 2));
+    const contributorItemList = Util.createAndAppend('li', container, {
+      class: 'contributorItem',
+    });
+    contributorItemList.addEventListener('click', () => {
```
remarcmij: This event listener will respond to mouse clicks but you can't visit the contributors url by moving to the contributor list item and then pressing Enter, as would be required for ARIA compliance.

Alternatively, you could generate an `a` tag as a direct child of the `li`, set its `href` to the contributor url and then create the remaining elements as children of the `a` tag. In this way the whole list element can be clicked. The Enter key will work too.

homework/Contributor.js
```diff
@@ -5,15 +5,28 @@
 // eslint-disable-next-line no-unused-vars
 class Contributor {
   constructor(contributor) {
-    this.contributor = contributor;
+    this.data = contributor;
```
remarcmij: Why do you think `this.data` is an improvement over `this.contributor` as was in the original code?

SignosRk: because it allows us to attach data of any type to DOM elements in a way that is safe from circular references and therefore from memory leaks.

homework/App.js
SignosRk: yes, you are right, I have to improve in naming everything, I just used fetch because i am awaiting the fetchJSON, no other reasons, I will change it later xd... I am not so creative with naming but it will come better with time! haha

homework/Contributor.js
remarcmij: I'm afraid that the name of this property has no relation to anything you mention here.

PR #162, opened 2019-01-05
Week3 homework

soumaiamamo:


homework/style.css
```diff
@@ -1,3 +1,130 @@
 .alert-error {
```
remarcmij: Your CSS still needs work for a mobile phone:

screenshot 2019-01-08 at 19 10 12

PR #165, opened 2019-01-29
Ammar Vehbi - JS3 Homework Week1

ammarvehbi:


homework/index.js
```diff
@@ -30,13 +30,118 @@
     return elem;
   }
 
+  function setInnerText(repository, nthChild, text) {
+    document.querySelector('tbody').childNodes[nthChild].lastChild.innerText = text;
```
yash-kapila: You don't seem to be using `repository` in this function at all. Can you remove it from the parameters list please?

homework/index.js
```diff
@@ -30,13 +30,118 @@
...
+  }
+
+  function presentRepositories(leftHandColumn, repository) {
+    const table = createAndAppend('table', leftHandColumn);
+    const tBody = createAndAppend('tbody', table);
+    const headsOfDetails = ['Repository:', 'Description:', 'Forks:', 'Updated:'];
+    for (let i = 0; i < 4; i++) {
```
yash-kapila: This for-loop doesn't seem the best way to do it. Firstly because you are hard-coding the index length as 4 instead of using `headsOfDetails.length`. Or you can try using the `forEach` method. Also, you need not necessarily use the inner for loop as you can simply create two `<td>` elements inside the `forEach`.

So, after creating the `table` and `tbody` tags, you can execute a forEach where you create two `td` elements and assign the `headsOfDetails` to the first ones. After that, you can assign the second ones with the actual information. Does that make sense?

homework/index.js
```diff
@@ -30,13 +30,118 @@
...
+      const tRow = createAndAppend('tr', tBody);
+      for (let j = 0; j < 2; j++) {
+        createAndAppend('td', tRow);
+        tRow.childNodes[0].innerText = headsOfDetails[i];
+      }
+    }
+    const tData = document.querySelector('tr').lastChild;
```
yash-kapila: I don't think you need to run the `querySelector` method on the `document` object. Since we are creating the `table` element above in this function, you can try to find the `tr` element inside it. It will save you some execution time as you won't be fetching the whole document but only a subset of it for the `tr` element. 🙂

homework/index.js
```diff
@@ -30,13 +30,118 @@
...
+    setInnerText(repository, 1, repository.description);
+    setInnerText(repository, 2, repository.forks);
+    setInnerText(repository, 3, date);
+  }
+
+  function renderContributors(rightHandColumn, contributors) {
+    contributors.forEach(contributor => {
+      const ul = createAndAppend('ul', rightHandColumn);
+      const li = createAndAppend('li', ul);
+      createAndAppend('img', li, { src: contributor.avatar_url, class: 'image', height: '48' });
```
yash-kapila: Just a suggestion to have the function call the same way as you are doing in other places when you have a few options to pass in as the third argument i.e if you see a couple of lines below, you have split the last argument in multiple lines which makes it easier to see but on this line you have the entire object in a single line 🙂

homework/index.js
```diff
@@ -30,13 +30,118 @@
...
+      const contributorData = createAndAppend('div', li, { class: 'contributor-data' });
+      createAndAppend('div', contributorData, {
+        text: contributor.login,
+        href: contributor.html_url,
```
yash-kapila: Do you need to have this `href` on a `div` element?

homework/style.css
```diff
@@ -1,3 +1,178 @@
...
+}
+
+#header {
+  background: #f96363;
+  color: #ffffffc5;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  border-radius: 0.2rem;
+  padding: 8px 16px;
```
yash-kapila: Hi Ammar - I see that you are using `rem` units which is quite nice but I see that there is some consistency still needed as you are using both `px` and `rem` for margins, paddings, font-sizes and line-heights. Can you please update them? 🙂

homework/index.js
yash-kapila: Similar to what you are doing in the `renderContributors` function

homework/index.js
```diff
@@ -30,13 +30,118 @@
...
+        tRow.childNodes[0].innerText = headsOfDetails[i];
+      }
+    }
+    const tData = document.querySelector('tr').lastChild;
+    createAndAppend('a', tData, {
+      href: repository.html_url,
+      target: '_blank',
+      id: 'link-repo',
+    });
+    document.getElementById('link-repo').textContent = repository.name;
```
yash-kapila: I would suggest that you save the return value of `createAndAppend` above in a variable and assign `textContent` to it. This way you need not lookup for the element using the `document.getElementById` method and can save some execution time 🙂

yash-kapila:
```
const repoName = createAndAppend('a', tData, {
  href: repository.html_url,
  target: '_blank',
  id: 'link-repo',
});
repoName.textContent = repository.name;
```

homework/index.js
ammarvehbi: I removed the unused parameter. Thanks :pray:

homework/index.js
ammarvehbi: I applied forEach for the `<tr>` tags too. Yeah, that was bad conduct of coding. Thanks :pray:

homework/index.js
ammarvehbi: I split arguments into multiple lines. :pray:

homework/index.js
ammarvehbi: `li.addEventListener('click', () => {
        window.open(contributor.html_url, '_blank');
      });`

I wrote this code after writing href. I forgot to delete it. I deleted it now. Thanks Yash. :pray:

homework/index.js
ammarvehbi: Applied. Thanks, Yash :pray:

homework/style.css
ammarvehbi: I changed all of them into rem. :sweat_smile:  Thanks Yash :pray:

ammarvehbi:
Dear @yash-kapila , 
Thank you very much for valuable feedback. I made the changes that you suggested. You can see them in the new commit. :pray: :slightly_smiling_face:

homework/index.js
```diff
@@ -30,18 +30,132 @@
...
+      }
+    });
+    const tData = table.querySelector('tr').lastChild;
+    const a = createAndAppend('a', tData, {
+      href: repository.html_url,
+      target: '_blank',
+    });
+    a.textContent = repository.name;
+
+    const upDate = `${new Date(repository.updated_at).toLocaleDateString()}, ${new Date(
```
yash-kapila: Hi Ammar - this is just FYI but you can move this out to a function as well. Something like:
```
const getDate = date => `${new Date(date).toLocaleDateString()}, ${new Date(date).toLocaleTimeString()}`
```
and then use it:
```
const upDate = getDate(repository.updated_at);
const creation = getDate(repository.created_at)l
```

PR #166, opened 2019-01-29
JS3 Github API Homework

AhmadAbdulhaleem:


homework/index.js
```diff
@@ -36,8 +36,92 @@
...
+          const repositoryName = createAndAppend('div', leftSection, { class: 'repName' });
+          const repositoryDescription = createAndAppend('div', leftSection, {
+            class: 'repDescription',
+          });
+
+          const repositoryForks = createAndAppend('div', leftSection, { class: 'repForks' });
+          const repositoryDate = createAndAppend('div', leftSection, { class: 'repDate' });
+
+          filteredData.forEach(allData => {
+            createAndAppend('span', repositoryName, { text: 'Repository Name:' });
```
yash-kapila: When you think of splitting your function into smaller functions, moving this block to a separate function would also promote re-usability. This way, for your first requirement where you need to display data for the first dropdown item by default and displaying updated data for different repos would be easier using the new function. Please let me know if you have any questions. 🙂

PR #167, opened 2019-01-30
Bassam - JS3/week1/homeWork

BassamHajar:
Thanks & regards,

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -30,119 +30,105 @@
     return elem;
   }
 
+  // display repository information ( left side )-------------------------------/ function /
+  function showRepoBasicInfo(container, repository) {
+    const repInfoTable = createAndAppend('table', container);
+    const tr1 = createAndAppend('tr', repInfoTable);
+
+    // raw 1
+    createAndAppend('td', tr1, { class: 'heading', text: 'repository:' });
+    createAndAppend('a', tr1, {
+      href: repository.html_url,
+      target: '_blank',
+      text: repository.name,
+      class: 'value',
+    });
+
+    // raw 2
+    const tr2 = createAndAppend('tr', repInfoTable);
+    createAndAppend('td', tr2, { class: 'heading', text: 'Description:' });
+    createAndAppend('td', tr2, { text: repository.description, class: 'value' });
+
+    // raw 3
+    const tr3 = createAndAppend('tr', repInfoTable);
+    createAndAppend('td', tr3, { class: 'heading', text: 'Forks:' });
+    createAndAppend('td', tr3, { text: repository.forks, class: 'value' });
+
+    // raw 4
+    const tr4 = createAndAppend('tr', repInfoTable);
+    createAndAppend('td', tr4, { class: 'heading', text: 'Updated:' });
+    createAndAppend('td', tr4, { text: repository.updated_at, class: 'value' });
+  }
+
+  // display the contributors of the selected repository ( right side )-----------------/ function /
+  function showContributors(container, contributors) {
+    createAndAppend('h2', container, { text: 'Contributions' });
+
+    const ul = createAndAppend('ul', container);
+    contributors.forEach(contributor => {
+      const li = createAndAppend('li', ul);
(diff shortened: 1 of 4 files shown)


homework/index.js
```diff
@@ -36,11 +36,115 @@
...
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        // create the basic element tags //
+        // the main header.
+        createAndAppend('h1', root, { text: 'HYF Repositories' });
+
+        // the container
+        createAndAppend('div', root, { id: 'container' });
+        const container = document.getElementById('container');
```
yash-kapila: Hi Bassam, if you check the `createAndAppend` function you will see that it returns the new element which is created. Thus, instead of calling that function and executing `document.getElementById` in the next line, you can try replacing it with a single statement i.e

```
const container = createAndAppend('div', root, { id: 'container' });
```

This saves you one line of code but more importantly, you don't have to search for the element in document and hence can save some JavaScript execution time.

homework/index.js
```diff
@@ -36,11 +36,115 @@
...
+
+        // the basic repository information (left aside).
+        createAndAppend('aside', container, {
+          class: 'left-aside box',
+          id: 'basicInfo',
+        });
+        const basicInfo = document.getElementById('basicInfo');
```
yash-kapila: Same as above. I see that you have followed a similar approach of creating an element and then looking it up on document when you can simply save the returned value of `createAndAppend` and assign it to a variable. Can you please update all instances?

homework/index.js
```diff
@@ -36,11 +36,115 @@
...
+        };
+        showInfo();
+
+        // Changing selected: show the basic information for each repository. + show the contributors on the right aside.
+        selected.addEventListener('change', () => {
+          infoList.innerHTML = '';
+          showInfo();
+        });
       }
+      // console.log(data);
```
yash-kapila: We can get rid of any commented out code. 🙂

homework/index.js
```diff
@@ -36,11 +36,115 @@
...
+        data.sort((a, b) => a.name.localeCompare(b.name, 'en', { sensitivity: 'base' }));
+        data.forEach(repo => {
+          createAndAppend('option', select, { text: repo.name });
+        });
+
+        // show the first repository basic info on the left aside & the contributors on the right one.
+        const selected = document.getElementById('select');
+
+        const showInfo = function() {
+          data.forEach(repo => {
```
yash-kapila: Ah - I think there is a better way to do it. You can firstly move the `showInfo` function out of `main` functions scope. Then, this function can accept a repository details as an argument along with some HTML elements in which you would populate the data. This way, you need not have a `forEach` to iterate through all 47 repos returned and having `if-else` check to match it with the currently selected repo.

Thus, your function call for displaying first repo details would look like:
```
showInfo(repository, repoNameBtn, infoList); // add other elements if needed in arguments list
```
and subsequently when this function is called, you will pass the selected element's repo details.

I would suggest that you have a look into the hint provided in the assignment description once: `Add one option element per repository to the select element, where each option element has the array index of the repository as its value attribute and the name of the repository as its text content:`. This would require some changes in lines 88 - 90 of your code and you can easily then move `showInfo` out. Please let me know if you have any questions. 🙂

homework/index.js
```diff
@@ -36,11 +36,115 @@
...
+              // contributors (right aside)
+              fetchJSON(repo.contributors_url, (error, contributorsData) => {
+                if (error) {
+                  createAndAppend('div', root, {
+                    text: error.message,
+                    class: 'alert-error',
+                  });
+                } else {
+                  contributorsList.innerText = '';
+                  contributorsData.forEach(contributor => {
```
yash-kapila: There is a small JS error here. For some repos, if they aren't having any contributors list i.e `contributorsData` is `null`, `forEach` would throw an error as it can't loop over a null element. 🙂

PR #168, opened 2019-01-31
week1 homework

ahmetbibi:


homework/index.js
```diff
@@ -30,13 +30,175 @@
...
+    createAndAppend('h1', header, { text: 'HackYourFuture Github Repositories' });
+    const choice = createAndAppend('div', root, { id: 'choice' });
+    createAndAppend('h3', choice, { text: 'Please Select a Repository Below' });
+    const selection = createAndAppend('div', choice, { id: 'selection' });
+    createAndAppend('p', selection, { text: 'HYF Repositories: ' });
+    const selectElem = createAndAppend('select', selection, { id: 'select' });
+    const repos = jsonData;
+    const repoNames = repos
+      .map(repo => repo.name)
+      .sort((x, y) => (String(x).toUpperCase() > String(y).toUpperCase() ? 1 : -1));
```
joostlubach: Tip: check out the method `String#localeCompare`!

homework/index.js
```diff
@@ -30,13 +30,175 @@
...
+        }
+      });
+    });
+  }
+
   function main(url) {
     fetchJSON(url, (err, data) => {
       const root = document.getElementById('root');
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
```
joostlubach: Why not also create a function for displaying the error?

homework/index.js
```diff
@@ -30,13 +30,175 @@
     return elem;
   }
 
+  function createHtml(jsonData) {
```
joostlubach: Overall, I suggest creating more specific-purpose functions to create parts of your application.

homework/index.js
```diff
@@ -30,13 +30,175 @@
...
+
+    repoNames.forEach((repoName, index) => {
+      createAndAppend('option', selectElem, { text: repoName, value: index });
+    });
+
+    const contentDiv = createAndAppend('div', root, { id: 'content' });
+    const leftDiv = createAndAppend('div', contentDiv, { id: 'left-side' });
+    const rightDiv = createAndAppend('div', contentDiv, { id: 'right-side' });
+
+    function getSelectedOption(sel) {
```
joostlubach: It's ok to write these functions outside of `createHtml`. Now you've made them into 'nested functions', which can often make code harder to debug.

homework/index.js
```diff
@@ -30,13 +30,175 @@
...
+        }
+      }
+      return opt;
+    }
+
+    function getSelectedIndex(name, opt) {
+      let selectedIndex = 0;
+      for (let i = 0; i < name.length; i++) {
+        if (name[i].name === opt.text) {
+          selectedIndex = i;
```
joostlubach: If you write `return i;` here, and `return 0;` in line 74, you break the function early. Because when the index has been found, you can stop the loop!

homework/index.js
```diff
@@ -30,13 +30,175 @@
...
+        }
```
joostlubach: Alternatively, you could write `break;` here.

homework/index.js
```diff
@@ -30,13 +30,175 @@
...
+        createAndAppend('a', contsInfoLeft, {
+          text: `${conts[index].login}`,
+          href: `${conts[index].html_url}`,
+          target: '_blank',
+        });
+        createAndAppend('p', contsInfoRight, { text: `${conts[index].contributions}` });
+      });
+    }
+
+    const CONT_URL = repos[firstSelElemIndex].contributors_url;
```
joostlubach: Check your casing - what is the reason you're using a SHOUT_CASE here?

homework/index.js
```diff
@@ -30,13 +30,175 @@
...
+    fetchJSON(CONT_URL, (err, data) => {
```
joostlubach: You're performing a `fetchJSON` call *inside* of a function called `createHtml`. It's better to extract this call into its own method and call it on window.onload (i.e. at the bottom of the page).

homework/index.js
```diff
@@ -30,13 +30,175 @@
...
+    allOptions.addEventListener('change', () => {
+      const selOpt = getSelectedOption(allOptions);
+      const selElemIndex = getSelectedIndex(repos, selOpt);
+      document.getElementById('repo-name').innerHTML = `${selOpt.text}`;
+      document.getElementById('repo-name').setAttribute('href', `${repos[selElemIndex].html_url}`);
+      document.getElementById('repo-description').innerHTML = `${repos[selElemIndex].description}`;
+      document.getElementById('repo-fork').innerHTML = `${repos[selElemIndex].forks}`;
+      document.getElementById('repo-updated').innerHTML = `${repos[selElemIndex].updated_at}`;
+
+      function deleteRightSideInfos() {
```
joostlubach: Now you're adding another nested function with its own nested functions. It's sort of hard to understand whether this is part of createHtml, addEventListener callback, or deleteRightSideInfos. Try to extract all these things in functions in the root scope.

PR #169, opened 2019-01-31
Utku / Class 19 / Homework Week 1

u61:
Created left column. It works properly.
Due to lack of time and knowledge I had to leave right column to next week as you mentioned and gave permission on Makeme.md section 1.6.
Regards

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -30,8 +30,27 @@
     return elem;
   }
 
+  function remove() {
+    if (document.getElementById('right_column_list') !== null) {
+      document.getElementById('right_column_list').remove();
+    }
+  }
+
+  function setLink(linkID, srcArray) {
+    document.getElementById(linkID).innerText = srcArray[0].name;
+    document.getElementById(linkID).setAttribute('href', srcArray[0].html_url);
+    document.getElementById(linkID).setAttribute('target', '_blank');
+  }
+
   function main(url) {
     fetchJSON(url, (err, data) => {
+      // SORT SELECTION MENU ALPHABETICALLY
+      data.sort((a, b) => a.name.localeCompare(b.name));
+
+      const contributorsURL = data.map(item => item.contributors_url);
+      // console.log(contributorsURL);
+      // console.log(typeof contributorsURL);
+
       const root = document.getElementById('root');
 
       if (err) {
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -33,10 +33,81 @@
...
+              document.getElementById('property_3').innerText = '-';
+            } else {
+              document.getElementById('property_3').innerText = updatedValue;
+            }
+          },
+          false,
+        );
+
+        // SORT SELECTION MENU ALPHABETICALLY
+        const dataSorted = data.sort((a, b) => a.name.localeCompare(b.name));
```
joostlubach: Watch out here: `sort` is a *mutating* function - it means that after this call, the contents of `data` itself are sorted. What it returns (and what you assign to `dataSorted` is literally the same array as `data`).

homework/index.js
```diff
@@ -33,10 +33,81 @@
...
+
+        // CREATE EVENT LISTENER
+        select.addEventListener(
+          'change',
+          () => {
+            const selected = select.selectedIndex;
+            const repositoryValue = data[selected].name;
+            const repoLink = data[selected].html_url;
+            if (repositoryValue === null) {
+              document.getElementById('property_0').innerText = '-';
```
joostlubach: Instead of calling it `property_0`, why not call it `property_repository`? That's much easier to read.

homework/index.js
```diff
@@ -33,10 +33,81 @@
...
+        const leftColumn = createAndAppend('div', root, { id: 'info_panel' });
+        const leftColumnTable = createAndAppend('table', leftColumn);
+        const leftColumnTableProperties = ['Repository :', 'Description :', 'Forks :', 'Updated :'];
+        const leftColumnTableFirstTr = createAndAppend('tr', leftColumnTable);
+        createAndAppend('td', leftColumnTableFirstTr, { text: leftColumnTableProperties[0] });
+        const linkRepo = createAndAppend('td', leftColumnTableFirstTr);
+        createAndAppend('a', linkRepo, { id: `property_${0}` });
+
+        // CREATE TABLE ELEMENTS OF LEFT COLUMN
+        for (let j = 1; j < 4; j++) {
```
joostlubach: What if I want to add a property? Let's say, I modify the array `leftColumnTableProperties`, and I add the property `"Maintainer: "`. Will it appear in the table?

homework/index.js
```diff
@@ -33,10 +33,81 @@
...
+          const leftColumnTableTr = createAndAppend('tr', leftColumnTable);
+          createAndAppend('td', leftColumnTableTr, { text: leftColumnTableProperties[j] });
+          createAndAppend('td', leftColumnTableTr, { id: `property_${j}` });
+        }
+
+        // GET DEFAULT VALUES OF LEFT COLUMN AT PAGE OPENING
+        document.getElementById('property_0').innerText = data[0].name;
```
joostlubach: Small suggestion: perhaps write a function that can assign this value, so you don't have to copy setting text, href and target. E.g. `setLinkAttributes(data[0].name, data[0].html_url, '_blank');`

homework/index.js
u61: Got it. So, there is no need to create a new variable because my intention is already to mutate the variable.

homework/index.js
u61: IDs are changed with more semantic ones.

homework/index.js
u61: I understand. Instead of using specific values, I could use dynamic values like array.length in for loop's definition. I changed it.

homework/index.js
u61: Written a function for this section.

u61:
Good Evening Joost,

Appreciate for your review.
I applied your corrections on my project.
I have put the right column also.
My code became so complicated, I am uncomfortable with that.
Furthermore, page is not stylish yet.
This is the week1 continue part.
I will apply week2 assignments on it and try to tidy and shorten it before deadline.

Regards

PR #170, opened 2019-01-31
homework week1

soolaf:


homework/index.js
```diff
@@ -29,18 +29,78 @@
...
+        const rightHand = createAndAppend('div', root, { id: 'right-hand' });
+
+        repositories.forEach((repository, index) => {
+          createAndAppend('option', select, { text: repository.name, value: index });
+        });
+        select.addEventListener('change', () => {
+          const index = select.value;
+          const repository = repositories[index];
+          renderRepositories(repository, leftHand);
+          fetchJSON(repository.contributors_url, (_err, contributors) => {
```
joostlubach: Small thing here - these lines (92 - 94) are almost identical to lines 98 - 100. Perhaps extract into a function.

homework/index.js
```diff
@@ -29,18 +29,78 @@
     });
     return elem;
   }
+  function renderRepositories(repository, container) {
```
joostlubach: The naming of these functions is 👍! It's very "React" to call this `render...`.

However a small thing here: this function renders *one* repository, so it should also be named `renderRepository`.

PR #171, opened 2019-01-31
Homework of Week1

ufukcalimli:


homework/index.js
```diff
@@ -30,13 +30,188 @@
...
+      id: 'forks',
+      text: `Forks: ${selectedData.forks}`,
+    });
+    createAndAppend('p', div, {
+      id: 'updated_at',
+      text: `Updated: ${selectedData.updated_at}`,
+    });
+  }
+
+  function generateContSection(selected) {
```
joostlubach: I would suggest typing out `contributor`. Right now I'm like: "Hmmm is this 'content'? Or 'continue'? 🤔"

homework/index.js
```diff
@@ -30,13 +30,188 @@
...
+    const contsDiv = createAndAppend('div', container, { id: 'contsDiv' });
```
joostlubach: Same here, I'd call it `'contributorsDiv'` or something.

homework/index.js
```diff
@@ -30,13 +30,188 @@
...
   function main(url) {
     fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        const repoName = data
+          .map(repo => repo.name)
+          .sort((a, b) => a.toLowerCase().localeCompare(b.toLowerCase()));
```
joostlubach: When you use `localeCompare`, you don't also have to use `toLowerCase`. `localeCompare` has its own means to compare casing, based on the language.

homework/index.js
```diff
@@ -30,13 +30,188 @@
...
+        id: id++,
+        value: repo,
+        text: repo,
+      });
+    });
+  }
+
+  function generateInfoSection(selected, data) {
+    const div = createAndAppend('div', container, { id: 'infoDiv' });
+    const selectedData = getSelectedData(selected, data);
```
joostlubach: Tip: look into the function `Array#find`. It does sort of what you want here :D.

PR #174, opened 2019-02-01
JavaScript3 HW1 - fatih

kfth:


kfth:
I had some problems about  eslint-plugin-prettier and i managed to fix it. Thank you

homework/index.js
```diff
@@ -30,13 +30,100 @@
...
+        showContributors(contributors);
+      }
+    });
+  }
+
+  function createHeader(data) {
+    const header = createAndAppend('header', root, { id: 'header' });
+    createAndAppend('h1', header, { text: 'Hack Your Future Repositories' });
+    const nav = createAndAppend('nav', root, { id: 'navigation' });
+    const select = createAndAppend('select', nav, { id: 'repositorySelect' });
```
yash-kapila: Hi Fatih - if we go back to our first JavaScript lessons, you would remember that it is a good practice to use kebab-casing rather than camelCasing for class names and ID attribute values in HTML elements. So instead of having `{ id: 'repositorySelect' }`, you should try for `{ id: 'repository-select' }`.

homework/index.js
```diff
@@ -30,13 +30,100 @@
...
+    const repositoryContainer = document.getElementById('repository');
+    removeChildElements(repositoryContainer);
+    const selected = document.getElementById('repositorySelect').selectedIndex;
+    const repository = data[selected];
+    createAndAppend('img', repositoryContainer, {
+      src: 'hyf.png',
+      alt: 'Hack Your Future Logo',
+      class: 'logo',
+    });
+    const h2 = createAndAppend('h2', repositoryContainer, { id: 'repositoryName' });
```
yash-kapila: `kebab-casing` here for ID please 🙂

homework/index.js
```diff
@@ -30,13 +30,100 @@
...
+    createAndAppend('p', repositoryContainer, { text: repository.description });
+
+    const contributorsContainer = document.getElementById('contributors');
+
+    fetchJSON(repository.contributors_url, (error, contributors) => {
+      removeChildElements(contributorsContainer);
+      if (error) {
+        displayError(error, contributorsContainer);
+      } else {
+        showContributors(contributors);
```
yash-kapila: You know you can pass the `contributorsContainer` here as the second argument to `showContributors` method and then you can avoid getting the element as you do on line 43(`const contributorsContainer = document.getElementById('contributors');`). You save one operation. 😉

PR #177, opened 2019-02-05
Week2

ahmetbibi:


homework/index.js
```diff
@@ -30,17 +15,204 @@
...
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+  function sortElements(repositories) {
+    return repositories.sort((a, b) => a.name.localeCompare(b.name));
+  }
+
+  function getSelectedOption(sel) {
```
yash-kapila: Hi Ahmet - this operation looks like a filtering operation where you are trying to filter out an element from the options list based on a certain condition i.e `opt.selected === true`. Although, this function with for-loop works well but I would suggest that you go back to the Week2 course and see the array operation methods like `filter`, `map` and `reduce`.

The operation here can be easily written using a filter function and then we need not create extra variables like `opt` or `i` anymore. Can you please check? 🙂

yash-kapila: Also, take care of one thing here. `sel.options` is not exactly an array but an HTMLCollection list so you might want to first convert it to an array using either the spread operator or `Array.from` method.

homework/index.js
```diff
@@ -30,17 +15,204 @@
...
+      opt = sel.options[i];
+      if (opt.selected === true) {
+        break;
+      }
+    }
+    return opt;
+  }
+
+  function getSelectedIndex(name, opt) {
+    let selectedIndex = 0;
```
yash-kapila: Hi Ahmet, just like the previous function, there is a possibility to do away with the for loops as JavaScript now provides an inbuilt method to search for a particular element's index. It's called `findIndex`. I think this method provides the same functionality that you wish to achieve with the for loop. Please have a look here for its documentation - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/findIndex

PR #179, opened 2019-02-06
homework week1

mohamdmahdi:


homework/index.js
```diff
@@ -16,27 +53,79 @@
...
+    repositories.sort((a, b) => a.name.localeCompare(b.name, 'en', { sensitivity: 'base' }));
+    repositories.forEach(repository => {
+      createAndAppend('option', selectRepositoryMenu, {
+        value: repository.id,
+        text: repository.name,
+      });
+    });
+    const repositorySection = createAndAppend('div', body, { id: 'repositoryInfo' });
+    selectRepositoryMenu.addEventListener('change', event => {
+      const selectedRepos = repositories.find(option => {
```
yash-kapila: The way how `find` method works is that it requires a testing function and it returns the value as soon as the testing matches(or returns true). In this case, rather than having to return something manually when there is a match, you can simply do:

```
return Number(option.id) === Number(event.target.value)
```

So this will return the repo as soon as the ID matches the event's target value. 🙂

homework/index.js
```diff
@@ -16,27 +53,79 @@
...
+    createAndAppend('span', dateRow, { text: RepositoryElement.updated_at });
+  }
+
+  function buildRepositoryList(repositories, parent) {
+    const top = createAndAppend('div', parent, { id: 'top' });
+    const body = createAndAppend('div', parent, { id: 'bodyTag' });
+    const rightSide = createAndAppend('div', body, { id: 'right' });
+    createAndAppend('label', top, { text: 'Select a Repository: ' });
+    const selectRepositoryMenu = createAndAppend('select', top, {
+      id: 'selectMenu',
```
yash-kapila: Please try to use `kebab-casing` rather than `camelCasing` when trying to enter values for HTML element's class and ID attributes. Refer the `alert-error` class for example 🙂

PR #180, opened 2019-02-06
Home Work Week 2

YaseirAlkhwalda:


homework/index.js
```diff
@@ -29,18 +14,99 @@
...
     });
   }
 
+  function main(url) {
+    const root = document.getElementById('root');
+    const header = createAndAppend('div', root, { class: 'header' });
+    createAndAppend('p', header, { text: 'HYF Repositories' });
+
+    fetch(url)
+      .then(Response => Response.json())
```
yash-kapila: Hi Yaseir, you should try to keep your variable names in lowercase or camelCase. Pascal casing names like `Response` in this case are usually used for naming classes or constructor functions. 🙂

homework/index.js
```diff
@@ -29,18 +14,99 @@
...
+            createAndAppend('div', root, {
+              text: Error.message,
+              class: 'alert-error',
+            }),
+          );
+
+        select.addEventListener('change', () => {
+          const index = select.selectedIndex;
+          createRepositoryDescription(repoContainer, repositories[index]);
+          fetch(repositories[index].contributors_url)
```
yash-kapila: Do you notice how this code snippet is similar to one above where you are fetching contributors data for the first repo.

Event listener:
```
          fetch(repositories[index].contributors_url)
            .then(contributors => contributors.json())
            .then(contributors => {
              createContributorsSide(contributorContainer, contributors);
            })
            .catch(changeError =>
                text: changeError.message,
                class: 'alert-error',
              }),
            );
```
and fetching contributors data for first repo:
```
        fetch(repositories[0].contributors_url)
          .then(response => response.json())
          .then(contributors => createContributorsSide(contributorContainer, contributors))
          .catch(Error =>
              text: Error.message,
              class: 'alert-error',
            }),
          );
```

In such cases, it is better to move the repeated logic in its own function say `fetchContributorsData` and call this function wherever you want to use the functionality. Can you try to do it? 🙂

homework/index.js
```diff
@@ -29,18 +14,99 @@
...
+        const contributorContainer = createAndAppend('div', container, {
+          class: 'right-div',
+        });
+        const select = createAndAppend('select', header, { class: 'repo-selector' });
+        repositories.sort((a, b) => a.name.localeCompare(b.name));
+        repositories.forEach(repository => {
+          createAndAppend('option', select, { text: repository.name });
+        });
+        createRepositoryDescription(repoContainer, repositories[0]);
+        fetch(repositories[0].contributors_url)
```
yash-kapila: In addition to my comment regarding moving this logic to a function, if you remember from our class last Sunday, we talked about promise chaining instead of having nested promises. The way you are trying to fetch the contributors details for the first repository, it is being done using nested promises. Can you try to make use of promise chaining here instead of nested promises?

Something like this inside your `main` function:
```
// (fetch ALL DATA)
fetch(url)
  .then(resp => resp.json())
  .then(data => {
    // create HTML as you are doing already
   // add the select event listener here
  // possibly return something here for the next .then block to use ;)
  })
  .then(/* fetch contributors details here by calling function created after above comment */)
  .catch(/* this need not be touched as it is being done correctly */);
```

yash-kapila:
Approving and closing the PR as review comments were addressed. Thanks @YaseirAlkhwalda.

PR #181, opened 2019-02-07
Week2

AhmadAbdulhaleem:


homework/index.js
```diff
@@ -1,20 +1,20 @@
...
-      if (xhr.status < 400) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
-  }
+  // function fetchJSON(url, cb) {
```
yash-kapila: Can remove any commented code 🙂

homework/index.js
```diff
@@ -30,15 +30,106 @@
...
+              });
+          });
+        });
+    }
+    selectRepo.addEventListener('change', fetchSingleData);
+    fetchSingleData();
+  }
+
+  function main(url) {
+    fetchAllRepoData(url).then(() => getRepoData(url));
```
yash-kapila: Hi Ahmat - upon page load, I see that there are two network requests going for the `/repos` endpoint. This happens because you call the `fetch` endpoint twice, once from the `main()` and then again inside `getRepoData` by calling an internal function `fetchSingleData`. Can you please try to correct this behaviour?

homework/index.js
```diff
@@ -30,15 +30,106 @@
...
+
+  function getRepoData(url) {
+    const selectRepo = document.getElementById('getRepoData');
+    const container = createAndAppend('div', root, { class: 'container' });
+    function fetchSingleData() {
+      return fetch(url)
+        .then(res => res.json())
+        .then(repData => {
+          container.innerHTML = '';
+          const filteredData = repData.filter(el => el.id === Number(selectRepo.value));
```
yash-kapila: I think you try to re-fetch the repos data so that you can apply this filter operation to get data of the selected repo(which is the first repo by default at page load). This can be avoided by simply accessing the 0th element of the repositories array as it represents the first element of an array. Thus, when you are creating the repositories container elements, you can use the 0th index instead of applying a filter operation first. This would allow you to do away with the second fetch operation. You might have to use a parameter to this function to get access to the repos data. 😉

homework/index.js
```diff
@@ -30,15 +30,106 @@
...
+              class: 'ContInfo-side',
+            });
+            createAndAppend('h4', contributorSection, {
+              text: 'Contributions',
+              class: 'contributorHeader',
+            });
+            const contributorWrapper = createAndAppend('div', contributorSection, {
+              class: 'contributorWrapper',
+            });
+            return fetch(allData.contributors_url)
```
yash-kapila: I think you should try to move fetching contributors data to its own separate function and then call that function from here and from the event listener. Please have a look into my comment on the Slack class19 channel and try to refactor this code.

homework/index.js
```diff
@@ -30,18 +15,107 @@
...
+        });
+        const selectRepo = document.getElementById('getRepoData');
+        const container = createAndAppend('div', root, { class: 'container' });
 
-  const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+        const fetchData = () => {
+          fetchSingleRepoData(fetchedData, selectEl.value, container);
+        };
+
+        fetchSingleRepoData(fetchedData, Number('82560415'), container);
```
yash-kapila: Hi Ahmat - having the ID value hardcoded here is quite risky. What happens if the ID changes of the repo in the BE. The code would break in such a case, wouldn't it? 🙂 

What you can do here is pass `fetchedData[0].id` instead of the hardcoded value.

homework/index.js
```diff
@@ -30,18 +15,107 @@
...
+      createAndAppend('p', repositoryDate, { text: allData.updated_at });
+      /* Fetch Contributor Data */
+      fetchContributorsData(allData, wrapper);
     });
   }
+  function fetchAllRepoData(url) {
+    return fetch(url)
+      .then(res => res.json())
+      .then(fetchedData => {
+        const navDiv = createAndAppend('div', root, { class: 'nav' });
```
yash-kapila: The way native `fetch` method works is that if a network response fails, it would still resolve the promise. That's why we need to have an extra check here for a failure scenario. Please refer this documentation - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch (**Checking that the fetch was successful**)

```
  if(response.ok) {
    return response.json();
  }
  throw new Error('Network response was not ok.');
```

PR #185, opened 2019-02-07
homework Week2

soolaf:


homework/index.js
```diff
@@ -1,19 +1,21 @@
...
-      if (xhr.status < 400) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
```
yash-kapila: Hi @soolaf - in addition to comments from @eidosam, the assignment for the week was to use the browser `fetch` API rather than converting the XHR to a promise. Although, this pattern is something which you might use later as you get deeper into web development(where callbacks are sometimes wrapped with promises) but it isn't needed for this assignment. Can you please correct this as well? 🙂 

Please let me know if you have any questions

soolaf: Hi @yash-kapila  @eidosam - thanks for your feedback,  I tried to correct my homework and I hope I did it correctly, I update my homework many times because there was a problem in my VS
sorry for that 🙂

yash-kapila:
Approving and closing the PR as all the comments have been addressed. Thank you @soolaf.

PR #186, opened 2019-02-07
Week2

ufukcalimli:


homework/index.js
```diff
@@ -1,19 +1,22 @@
...
-      if (xhr.status < 400) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
```
yash-kapila: Hi Ufuk, I think you were supposed to use the native `fetch` method available in the browser rather than doing this. Having said that, I really liked that you learnt about wrapping callbacks in promises. You would definitely use this sort of a pattern in the future while developing real applications.

homework/index.js
```diff
@@ -30,15 +33,189 @@
...
+    updatedAt.innerText = `Updated: ${selectedData.updated_at}`;
+  }
+
+  function main(url) {
+    fetchJSON(url).then(
+      data => {
+        const repoName = data.map(repo => repo.name).sort((a, b) => a.localeCompare(b));
+        generateSelections(repoName);
+
+        const selected = document.getElementById('repositories');
```
yash-kapila: This might have been missed earlier but I think you can combine the two statements together in a single line i.e instead of calling `generateSelections()` and then trying to fetch `repositories` element created by it, you can simply update your function to return the newly created element and store that in the return value.

```
function generateSelections(repos) {
  // code remains the same
  return selectionElem;
}

// Inside main()
const selected = generateSelections(repoName);
```

homework/index.js
```diff
@@ -30,15 +33,189 @@
...
+
+    const updatedAt = document.getElementById('updated_at');
+    updatedAt.innerText = '';
+    updatedAt.innerText = `Updated: ${selectedData.updated_at}`;
+  }
+
+  function main(url) {
+    fetchJSON(url).then(
+      data => {
+        const repoName = data.map(repo => repo.name).sort((a, b) => a.localeCompare(b));
```
yash-kapila: Umm.. maybe you can add the ID being returned in the `data` by the API here and send it to `generateSelections` method so that you don't have to write the `id` logic there.

Something like:
```
const repoName = data
  .map(repo => ({
    name: repo.name,
    id: repo.id
  })
  .sort((a,b) => a.name.localeCompare(b.name));
```

homework/index.js
```diff
@@ -30,15 +33,189 @@
...
     });
+    return selected;
+  }
+
+  function generateSelections(repos) {
+    const header = createAndAppend('header', root, {});
+    const selectionElem = createAndAppend('select', header, {
+      id: 'repositories',
+    });
+    let id = 0;
```
yash-kapila: Try removing the `id` logic here as per my comment below in the main() 🙂

homework/index.js
```diff
@@ -30,15 +33,189 @@
...
+  function generateContributorsSection(selected) {
+    const contributorsDiv = createAndAppend('div', container, { id: 'contsDiv' });
+    createAndAppend('p', contributorsDiv, { id: 'cont-header', text: 'Contributors' });
+    const contributorsLists = createAndAppend('ul', contributorsDiv, { id: 'conts-list' });
+
+    fetchJSON(`https://api.github.com/repos/HackYourFuture/${selected}/contributors`).then(
+      data => {
+        data.forEach(item => {
+          const contributorListItem = createAndAppend('li', contributorsLists, {
+            id: `${item.login}ListItem`,
```
yash-kapila: The attributes in HTML elements should ideally be kept `kebab-case` unlike `camelCase` 🙂

homework/index.js
```diff
@@ -30,15 +33,189 @@
...
+        });
+      },
+      err => {
+        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+      },
+    );
+  }
+
+  function updateContributorSection(selected) {
+    const contributorsDiv = document.getElementById('contsDiv');
```
yash-kapila: This is interesting. If you compare both `generateContributorsSection` and `updateContributorSection` methods, you will see that except `contributorsDiv.innerHTML = '';` this statement, all other logic is exactly same.

If such is the case, I would suggest that you keep your code DRY(don't repeat yourself). You can inside `updateContributorSection`, after clearing out the innerHTML, call the `generateContributorsSection` function. This way, you are saving about 40 lines of code and that's a lot. 🙂

homework/index.js
```diff
@@ -30,15 +33,189 @@
...
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+  const root = document.getElementById('root');
+  const container = createAndAppend('div', root, { id: 'container' });
+
+  function getSelectedData(value, data) {
+    let selected;
```
yash-kapila: If you look carefully, the operation which you are trying to do here is a `filter` operation i.e out of `data`, you want to filter out the entry where the name matches `value` being passed. Can you try to use the filter operation here instead? 😉

yash-kapila:
Hi @ufukcalimli - it seems that some of your code pushes from today aren't passing Travis checks. Can you please have a look? 🙂

ufukcalimli:
The code has passed from test. The only problem is that I couldn't apply your last review that using filter function on line 39. When I applied filter function I couldn't display repositories names on the list and on the "infoDiv" I get undefined.

homework/index.js
yash-kapila: Hi Ufuk - this is another way how you can write the `getSelectedData` method using the Array.prototype.filter method(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/filter).

```
function getSelectedData(value, data) {
  return data.filter(repo => repo.name === value)[0];
}
```

homework/index.js
yash-kapila: I think you missed addressing this point in your updates. I am approving the PR right now but please do have a look. It saves you one operation of adding an element and then finding it using `document.getElementById` in the next step. 🙂

homework/index.js
yash-kapila: Even better, you can try to use `Array.prototype.find` method here. More information can be found here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/find

This way, you need not return the first element from the filter operation and can re-write the function as:
```
function getSelectedData(value, data) {
  return data.find(repo => repo.name === value);
}
```

PR #187, opened 2019-02-07
Utku / Class 19 / Homework Week 2

u61:


homework/index.js
```diff
@@ -30,15 +15,162 @@
     return elem;
   }
 
+  function remove() {
+    document.getElementById('right_column_list').remove();
+  }
+
+  function setLink(linkID, srcArray) {
+    document.getElementById(linkID).innerText = srcArray.name;
```
yash-kapila: If you look closely, you are trying to fetch the element here in all three steps before you assign its `innerText`, `href` and `target` attributes. You can instead have the calling function i.e from where we call `setLink` to provide the element itself or simply retrieve it once in this function and use it in all three assignments. This way, you wouldn't have to fetch the same element thrice and will save yourself two operations. 

```
function setLink(linkId, srcArray) {
  const elem = document.getElementById(linkID);
  elem.innerText = srcArray.name;
  elem.setAttribute('href', srcArray.html_url);
  elem.setAttribute('target', '_blank');
}
```

homework/index.js
```diff
@@ -30,15 +15,162 @@
...
+        select.addEventListener('change', () => {
+          const selected = select.selectedIndex;
+          const fetchedDataContributors = fetch(contributorsURL[selected]).then(response =>
+            response.json(),
+          );
+          const rigthColumnList = createAndAppend('ul', rightPanel, {
+            id: 'right_column_list',
+          });
+
+          fetchedDataContributors
```
yash-kapila: Lines 131 to 166 can hence be removed when we create a common function for fetching contributors. 🙂

homework/index.js
```diff
@@ -30,15 +15,162 @@
...
+                text: contributor.contributions,
+              });
+            });
+          });
+        }
+
+        createRightPanel();
+
+        select.addEventListener('change', () => {
+          remove();
```
yash-kapila: You can also put the `remove()` inside the listener you are having below.

homework/index.js
```diff
@@ -30,15 +15,162 @@
...
+    document.getElementById('right_column_list').remove();
+  }
+
+  function setLink(linkID, srcArray) {
+    document.getElementById(linkID).innerText = srcArray.name;
+    document.getElementById(linkID).setAttribute('href', srcArray.html_url);
+    document.getElementById(linkID).setAttribute('target', '_blank');
+  }
+
+  const root = document.getElementById('root');
```
yash-kapila: I would suggest you move all these statements to its own function which you can call from the `main()`. And if you are using some of these elements in other places, you can then try return those elements or do a `document.getElementById` in areas where you need them.

homework/index.js
```diff
@@ -30,15 +15,162 @@
...
+          document.getElementById('property_updated').innerText = new Date(dataIndex.updated_at);
+        }
+
+        assignLeftPanelValues(data[0]);
+
+        function createRightPanel() {
+          const rigthColumnListDefault = createAndAppend('ul', rightPanel, {
+            id: 'right_column_list',
+          });
+          fetchedDataContributorsDefault.then(data2 => {
```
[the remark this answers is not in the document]
u61: Created a new function and implemented it required places as you mentioned.

homework/index.js
u61: Replaced your code but I understand what you meant.
Surely, this should be like this.

homework/index.js
u61: Yes, I removed that huge block.

homework/index.js
u61: Absolutely. I failed to noticed that as well.

homework/index.js
u61: Actually, I was trying to empty main function in order to get more pure code. But, in the end there is no need for this. Included them again.

u61:
Hello Yash,

I have tried to refactor my code as per your suggestions.

Thank you for your corrections and your time.

PR #188, opened 2019-02-07
Ammar Vehbi - JS3 Homework Week2

ammarvehbi:


homework/index.js
```diff
@@ -30,18 +17,144 @@
...
+      const index = select.selectedIndex;
+      presentRepositories(repositoriesContainer, repositories[index]);
+      presentContributors(contributorsContainer, repositories[index].contributors_url);
+    });
+  }
+
+  function main(url) {
+    fetch(url)
+      .then(response => {
+        if (response.ok) {
```
yash-kapila: Hi Ammar - for this assignment, you need not have this check for `response.ok`. In case there is any issue in fetching response from GitHub, it is automatically going to go inside the catch block. 🙂

yash-kapila: Hi Ammor - my apologies. I forgot that `fetch` doesn't rejects a promise in case there is an error while fetching a resource. In that case, the check that you have added is necessary. Please ignore my above comment. 🙂 

Thank you @remarcmij for correcting me here.

ammarvehbi: :pray: :slightly_smiling_face:

PR #190, opened 2019-02-08
Week2

mohamdmahdi:
I had to catch up all the learning and then doing it.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,5 +1,3 @@
-/* eslint-disable consistent-return */
-
 'use strict';
 
 {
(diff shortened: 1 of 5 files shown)


homework/index.js
```diff
@@ -27,18 +12,130 @@
...
+    buildRepositoryInfo(repositories[0], repositorySection);
+    rightSide.innerHTML = '';
+    fetchRepositoryContributors(repositories[0].contributors_url, rightSide);
+  }
+
+  function main(url) {
+    const promise = fetchJSON(url);
+    promise
+      .then(response => {
+        const root = document.getElementById('root');
```
yash-kapila: Hi Mohamd - you can also move out the retrieval of `root` element at the beginning of function and need not repeat it in both `.then` and `.catch` blocks.

homework/index.js
```diff
@@ -27,18 +12,130 @@
...
+    createAndAppend('span', dateRow, { text: RepositoryElement.updated_at });
+  }
+
+  function buildRepositoryList(repositories, parent) {
+    const top = createAndAppend('div', parent, { id: 'top' });
+    const body = createAndAppend('div', parent, { id: 'bodyTag' });
+    const rightSide = createAndAppend('div', body, { id: 'right' });
+    createAndAppend('label', top, { text: 'Select a Repository: ' });
+    const selectRepositoryMenu = createAndAppend('select', top, {
+      id: 'selectMenu',
```
yash-kapila: Please refer to my week1's review comments about using `kebab-casing` 🙂

homework/index.js
```diff
@@ -27,18 +12,130 @@
...
+    repositories.forEach(repository => {
+      createAndAppend('option', selectRepositoryMenu, {
+        value: repository.id,
+        text: repository.name,
+      });
+    });
+    const repositorySection = createAndAppend('div', body, { id: 'repositoryInfo' });
+    selectRepositoryMenu.addEventListener('change', event => {
+      const selectedRepos = repositories.find(option => {
+        if (Number(option.id) === Number(event.target.value)) {
```
yash-kapila: Please refer to my week1's homework comment about using the `Array.prototype.find` method.

homework/index.js
```diff
@@ -27,18 +12,130 @@
...
+      createAndAppend('span', contributorData, { text: contributor.login, class: 'name' });
+      createAndAppend('span', contributorData, {
+        text: contributor.contributions,
+        class: 'numberOfContribution',
+      });
+    });
+    return parent;
+  }
+  const fetchJSON = url => {
+    const promise = new Promise((resolve, reject) => {
```
yash-kapila: Hi Mohamd, even though this is a pattern of wrapping callbacks to promises which you probably would use later in the future but as far as this week's homework is concerned, you are to use the browser native `fetch` API. Please find the documentation around it here - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

The native API works quite similar to what you are doing here explicitly. Thus, you need not use this function in your assignment. But please be mindful of this statement in the documentation above: `An accurate check for a successful fetch() would include checking that the promise resolved, then checking that the Response.ok property has a value of true.`

This means that even if a promise is resolved, it doesn't mean that the request was successful. It could have failed as well. Which is why, we have an extra check for `response.ok` to make sure that our response was successful. The documentation provides a code snippet about how you can use it. Please refer the same.

PR #191, opened 2019-02-09
Homework Week2

alquinex:


homework/index.js
```diff
@@ -1,19 +1,21 @@
...
-      if (xhr.status < 400) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
```
yash-kapila: Hi Alusine, even though this is a pattern which you might encounter as you dive deep to programming where it is sometimes useful to wrap callbacks into promises but this homework doesn't require you to do so. Instead, you had to use the native browser `fetch` method which kind of does the same thing implicitly as your are doing here. Can you please try to use the same? More information on the `fetch` method can be found here - https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch

Please be mindful of this statement which you can also find in the documentation - `An accurate check for a successful fetch() would include checking that the promise resolved, then checking that the Response.ok property has a value of true. `

This means, even if the promise has resolved, it is not sure if the response was successful. You will have to do another check of `response.ok` to make sure that you have got the repositories and contributors data.

homework/index.js
```diff
@@ -30,15 +32,73 @@
...
+      });
     });
+
+    const container = createAndAppend('div', root, { id: 'container' });
+    const repoContainer = createAndAppend('div', container, { id: 'repo-container' });
+    const contributorContainer = createAndAppend('div', container, { id: 'contributor-container' });
+
+    select.addEventListener('change', () => {
+      const selectedRepo = repos[select.value];
+      console.log('name', selectedRepo);
```
yash-kapila: Could you please remove the console.log statements from here?

homework/index.js
```diff
@@ -30,15 +32,73 @@
...
+      repoContainer.innerHTML = '';
+      contributorContainer.innerHTML = '';
+
+      const updatedAt = new Date(selectedRepo.updated_at);
+
+      createAndAppend('p', repoContainer, { text: `Repository: ${selectedRepo.name}` });
```
yash-kapila: I wonder why you moved this logic in the event listener. In week1's homework, you had separate function for this logic which was better than doing it like this. Could you please explain the reasoning for doing it this way?

homework/index.js
```diff
@@ -30,15 +32,73 @@
...
+    repos.forEach((repo, index) => {
+      createAndAppend('option', select, {
+        text: repo.name,
+        value: index,
+      });
     });
+
+    const container = createAndAppend('div', root, { id: 'container' });
+    const repoContainer = createAndAppend('div', container, { id: 'repo-container' });
+    const contributorContainer = createAndAppend('div', container, { id: 'contributor-container' });
```
yash-kapila: There is one thing missing from this week's homework which was working in Week1 which is fetching the first repositories data by default on page load. You did it using this earlier:

```
fetchAndRender(repos[0], repoContainer, contributorContainer, root);
```

but it isn't there in this week's homework. Could you please look into it?

homework/index.js
alquinex: Hi Yash,
Thanks for taking the time to review my homework. I am currently loaded with other homework and exercises. I will soon complete the required tasks you mentioned and will let you know.

yash-kapila: Thank you. I realise that you must be stuffed with other home-works.

PR #194, opened 2019-02-13
Utku / Class 19 / Homework Week 3

u61:


homework/Repository.js
```diff
@@ -12,9 +12,19 @@ class Repository {
    * Render the repository info to the DOM.
    * @param {HTMLElement} container The container element in which to render the repository.
    */
-  render(container) {
+
+  static assignLeftPanelValues(dataIndex) {
```
yash-kapila: Hi Utku - this method shouldn't necessarily be a static method. It should instead be a method on the class which can be accessed from an object. I think you tried doing the same thing we are doing in Utils.js but please know that it serves a different purpose there.

A rule of thumb while trying to identify when to make a method `static` or not is: ```does it make sense to call this method, even if no Obj has been constructed yet?```

If we go with the above rule of thumb, calling `assignLeftPanelValues` without a repository object doesn't make sense which is why we need it as a class method. 🙂

homework/Contributor.js
```diff
@@ -12,8 +12,31 @@ class Contributor {
    * Render the contributor info to the DOM.
    * @param {HTMLElement} container The container element in which to render the contributor.
    */
-  render(container) {
+  static render(data, container) {
```
yash-kapila: Please make this an `instance` method of the class rather than `static`. 🙂

homework/Contributor.js
```diff
@@ -12,8 +12,31 @@ class Contributor {
...
     // TODO: replace the next line with your code.
```
yash-kapila: Can remove the comments as well

homework/App.js
```diff
@@ -15,14 +15,70 @@ class App {
     // Add code here to initialize your app
```
yash-kapila: Please remove the comments. 🙂

homework/App.js
```diff
@@ -15,14 +15,70 @@ class App {
...
+    });
 
     try {
-      const repos = await Util.fetchJSON(url);
-      this.repos = repos.map(repo => new Repository(repo));
+      const data = await Util.fetchJSON(url);
+      this.data = data.map(repo => new Repository(repo));
+      this.data = data.sort((a, b) => a.name.localeCompare(b.name));
+      const contributorsURL = this.data.map(item => item.contributors_url);
+      const fetchedDataContributorsDefaultRaw = await fetch(contributorsURL[0]);
```
yash-kapila: Why not use `Utils.fetchJSON` here as well. 😉

homework/App.js
```diff
@@ -44,24 +100,9 @@ class App {
    * repo and its contributors as HTML elements in the DOM.
    * @param {number} index The array index of the repository.
    */
-  async fetchContributorsAndRender(index) {
+  async fetchContributorsAndRender(data, container) {
     try {
```
yash-kapila: I don't think you need to put this in a try-catch block

homework/index.js
```diff
@@ -30,17 +15,143 @@
...
+        class: 'badge',
+        text: contributor.contributions,
+      });
+    });
+  }
+
+  async function showData(data) {
+    data.sort((a, b) => a.name.localeCompare(b.name));
+    const contributorsURL = data.map(item => item.contributors_url);
+    const fetchedDataContributorsDefaultRaw = await fetch(contributorsURL[0]);
```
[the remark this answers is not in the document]
u61: Done.

homework/Repository.js
u61: Absolutely. I have replaced them with instance methods.

homework/Contributor.js
u61: Done.

homework/Contributor.js
u61: Replaced.

homework/App.js
u61: Replaced.

homework/App.js
u61: Yes, no longer need.

u61:
Hi Yash,

I've editted my code as per your recommendations.

Thank You

homework/App.js
```diff
@@ -7,72 +7,86 @@ class App {
...
+      id: 'right_column_list',
+    });
 
     try {
-      const repos = await Util.fetchJSON(url);
-      this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      const data = await Util.fetchJSON(url);
+      this.data = data.map(repo => new Repository(repo));
+      this.data = data.sort((a, b) => a.name.localeCompare(b.name));
```
yash-kapila: Hi Utku - there is something not right here. On line 34, you are creating a new repository object for each repository and assigning it to `this.data`. However, on the next line itself, you are trying to sort the fetched repositories data and assign it again to `this.data`.

This causes `this.data` to lose the repository objects you had created before. Due to this, you are having to use the `assignLeftPanelValues` method incorrectly at line 54 where you are using `.prototype` property. This method would have been already present with your object created in line 34 and you could have called it simply as `this.data[0].assignLeftPanelValues()`.

Could you please look into it? Drop me a message if you need any further information. 🙂

u61: Good evening Yash,

I stucked at this point.

I have tried to not touch skeleton code as much as I can when I have been writing this code.
But at this point, only solution that I can see is deleting `line 34`(it is skeleton). Because, I really bored with this homework. Struggled too much while trying to change code from one homework's objective to another. I felt that if i wrote whole code again.

I didn't understand classes(syntactic sugar/'salt') very well, I have tried to recover but for now I couldn't.

The problem is (from my side); there were no enough blueprints on JS3 HW1 & HW2 to achieve a code that can be easily(without major alteration) implemented one to another. In HW3, I faced a skeleton code that is totally conflicting with my code's layout. I have changed it as I can and it worked.

So, I don't have any idea about how I can transform existing code but i will try to focus again.

u61: I know that ignorance is not an option to become better but I believe that this homework does not have any benefit for me at this point, just wasting my time and confusing my brain. But as I said I will try.

u61:
Hello Yash,

I've corrected last part thanks to your today's recap.

Appreciate for your help.

PR #197, opened 2019-02-14
Week3_Bassam = JS3/week3/homework.

BassamHajar:
Thanks & regards,

### DIFF
--- a/homework/App.js
+++ b/homework/App.js
@@ -14,16 +14,62 @@ class App {
   async initialize(url) {
     // Add code here to initialize your app
     // 1. Create the fixed HTML elements of your page
-    // 2. Make an initial XMLHttpRequest using Util.fetchJSON() to populate your <select> element
+    this.root = document.getElementById('root');
 
-    const root = document.getElementById('root');
+    // header:
+    const header = Util.createAndAppend('header', this.root, { class: 'header' });
 
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
+    // title:
+    Util.createAndAppend('h1', header, { text: 'HYF Repositories' });
+
+    // select element:
+    const select = Util.createAndAppend('select', header);
+
+    // the container:
+    this.theContainer = Util.createAndAppend('div', this.root, { id: 'container' });
+
+    // the basic repository information (left aside).
+    this.basicRepoInfoAside = Util.createAndAppend('aside', this.theContainer, {
+      class: 'box',
+    });
+
+    // the repository contributors (right aside).
+    this.contributorsAside = Util.createAndAppend('aside', this.theContainer, {
+      class: 'box',
+    });
 
     try {
-      const repos = await Util.fetchJSON(url);
-      this.repos = repos.map(repo => new Repository(repo));
       // TODO: add your own code here
+      // 2. Make an initial XMLHttpRequest using Util.fetchJSON() to populate your <select> element
+      const reposData = await Util.fetchJSON(url);
+
+      // sorting the repositories (case-insensitive) on repository name. && creating instances of class Repository:
+      this.repositories = reposData
+        .sort((a, b) => a.name.localeCompare(b.name, 'en', { sensitivity: 'base' }))
+        .map(repo => new Repository(repo));
+
+      // populating repositories as options to the select element:
+      this.repositories.forEach((repo, index) => {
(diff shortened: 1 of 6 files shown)


homework/index.js
```diff
@@ -30,17 +15,114 @@
...
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
-      }
+  // Function: display repository information ( left side )
+  function showRepoBasicInfo(container, repository) {
+    const repInfoTable = createAndAppend('table', container);
+    const tr1 = createAndAppend('tr', repInfoTable);
+
+    // raw 1
+    createAndAppend('td', tr1, { class: 'heading', text: 'repository:' });
+    createAndAppend('a', tr1, {
```
Meno-101: an `a` tag is not a valid child element for the `tr` so, it may cuase some sort of bugs into your code, Imagine the following:

you have an unorderedlist element which holds a list element, if you changed the `ul` to be an `ol` the list element will be changed to number or letters instead of bullets, but if you have an `h1` inside where the bullets will go?

homework/index.js
```diff
@@ -30,17 +15,114 @@
...
+      data.forEach((repo, index) => {
+        // appending repositories as options to the select element.
+        basicRepoInfoAside.innerHTML = '';
+        contributorsAside.innerHTML = '';
+        createAndAppend('option', select, { text: repo.name, value: index });
+        showRepoBasicInfo(basicRepoInfoAside, data[0]);
+        showContributors(data[0].contributors_url, contributorsAside);
+
+        // change select:
+        select.addEventListener('change', () => {
```
Meno-101: you'r having your `eventListener` inside your `forEach`, which is okay in your case as a very small application, but in a real worl you need to have it seperated, you will know why in a later advanced classes.

homework/Repository.js
```diff
@@ -12,16 +12,46 @@ class Repository {
...
-    // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.repository, null, 2));
+  render(container, theRepo) {
+    // creating a table:
+    const repInfoTable = Util.createAndAppend('table', container);
+    const tr1 = Util.createAndAppend('tr', repInfoTable);
+
+    // theRepo is an object of object. (get an array of the values to iterate through):
+    const theRepoValues = Object.values(theRepo);
+    theRepoValues.forEach(repo => {
```
Meno-101: what about:
```
theRepo.repository.forEach(...)
```

homework/Repository.js
```diff
@@ -12,16 +12,46 @@ class Repository {
...
+      // raw 1
+      Util.createAndAppend('td', tr1, { class: 'heading', text: 'repository:' });
+      Util.createAndAppend('a', tr1, {
```
Meno-101: Same comment as the previous about validation tags.

homework/Contributor.js
```diff
@@ -12,8 +12,23 @@ class Contributor {
    * Render the contributor info to the DOM.
    * @param {HTMLElement} container The container element in which to render the contributor.
    */
-  render(container) {
-    // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.contributor, null, 2));
+  render(table, contributor) {
+    // appending the table with the list items:
+    const li = Util.createAndAppend('li', table);
```
Meno-101: being more specific, `table` is not a table 🙄.

homework/App.js
```diff
@@ -14,16 +14,62 @@ class App {
...
 
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
+    // title:
+    Util.createAndAppend('h1', header, { text: 'HYF Repositories' });
+
+    // select element:
+    const select = Util.createAndAppend('select', header);
+
+    // the container:
+    this.theContainer = Util.createAndAppend('div', this.root, { id: 'container' });
```
Meno-101: you don't need to set everything on the this keyword, unless you have to use them in your class again somewhere.
so you may need to have this as a normal variable, while some of it's fellows can be on the `this`. 🙂

Meno-101: you don't need to set everything on the this keyword, unless you have to use them in your class again somewhere.
so you may need to have this as a normal variable, while some of it's fellows can be on the `this`. 🙂

PR #198, opened 2019-02-14
Week3

ufukcalimli:


homework/index.js
```diff
@@ -30,15 +26,137 @@
...
+  }
+
+  function main(url) {
+    try {
+      const json = fetchJSON(url);
+      json.then(data => {
+        const repoNames = data
+          .map(repo => ({ name: repo.name, id: repo.id }))
+          .sort((a, b) => a.name.localeCompare(b.name));
+        generateSelections(repoNames);
```
Meno-101: if you are returning the `selectionElem`, then probably you're having an extra getElementById in the next line.
and using it, what about:
```suggestion
        const selected = generateSelections(repoNames); // you may change the name of `selected` _(pastense)_ to `select` _(noun)_
```

homework/index.js
```diff
@@ -30,15 +26,137 @@
...
+        selected = item;
       }
     });
+    return selected;
+  }
+
+  function generateSelections(repos) {
+    const header = createAndAppend('header', root, {});
+    const selectionElem = createAndAppend('select', header, {
+      id: 'repositories',
```
Meno-101: And because you are returning and using the retruned value, _(if you followed the previous comment)_ this is an extra `id` in your element

homework/index.js
```diff
@@ -30,15 +26,137 @@
...
+    });
+
+    repos.forEach(repo => {
+      createAndAppend('option', selectionElem, {
+        id: repo.id++,
```
Meno-101: You are recieving a ready `id` making the item identified without any kind of repetition, here you have two main things to have an attention:
  1. you are changing the original id, for future when you recieve a one don't change it on the client, if you don't have you may create a one, but not when you have it already.
  
  2. you are creating `id`s attribute for each option you have recieved as a repo name, means if you have 100 option then you have 100 different `id` you will never use, here or in your Css.

homework/index.js
```diff
@@ -30,15 +26,137 @@
...
+        value: repo.name,
```
Meno-101: Smart move 😎

homework/index.js
```diff
@@ -30,15 +26,137 @@
...
 
-  function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+  const root = document.getElementById('root');
+  const container = createAndAppend('div', root, { id: 'container' });
```
Meno-101: I liked this container 🙌

homework/index.js
```diff
@@ -30,15 +26,137 @@
...
+
+  function getSelectedData(value, data) {
```
Meno-101: 💯

homework/index.js
```diff
@@ -30,15 +26,137 @@
...
+    }
+  }
+
+  function updateContributorSection(selected) {
+    const contributorsDiv = document.getElementById('contsDiv');
+    contributorsDiv.parentNode.removeChild(contributorsDiv);
+    generateContributorsSection(selected);
+  }
+
+  function updateInfoSection(selected, data) {
```
Meno-101: I liked these two functions 👆🏻

Meno-101:
Hey @ufukcalimli, Thanks for submitting the second part of the homework, I'll be reviewing it asap 😃.

homework/App.js
```diff
@@ -17,18 +17,40 @@ class App {
     // 2. Make an initial XMLHttpRequest using Util.fetchJSON() to populate your <select> element
 
     const root = document.getElementById('root');
+    Util.createAndAppend('div', root, { id: 'container' });
 
     Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
```
Meno-101: 🤔

homework/Contributor.js
```diff
@@ -13,7 +13,37 @@ class Contributor {
    * @param {HTMLElement} container The container element in which to render the contributor.
    */
   render(container) {
-    // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.contributor, null, 2));
+    // Util.createAndAppend('pre', container, JSON.stringify(this.contributor, null, 2));
```
Meno-101: I saw many commented code in your files, and it's fine for now, but in the near future you will see thousends of lines, and you will probably think about leaving it out.

homework/App.js
```diff
@@ -17,18 +17,40 @@ class App {
...
 
     try {
       const repos = await Util.fetchJSON(url);
       this.repos = repos.map(repo => new Repository(repo));
```
Meno-101: Oops! you've forgotting the sort.

homework/App.js
```diff
@@ -17,18 +17,40 @@ class App {
...
-      // TODO: add your own code here
+
+      const select = this.generateSelections(repos, root);
+      this.fetchContributorsAndRender(repos.indexOf(this.repos[0].repository));
```
Meno-101: All this line is just about select the 0 index from the repos array, so istead you may do the following:
```suggestion
      this.fetchContributorsAndRender(0); // it's same result as yours 👆
```
because you have:
  - `this.repos[0].repository` is the repositories array
  - `repos.indexOf` with it call will give you the index of your item which is as you spicified the index number `[0]`.

So all together going to give you as a result:
```javascript
      console.log(repos.indexOf(this.repos[0].repository)) // 0
```

homework/App.js
```diff
@@ -52,16 +74,14 @@ class App {
...
+      repo.render(container);
+      const rightDiv = Util.createAndAppend('div', container, { id: 'contsDiv' });
 
-      repo.render(leftDiv);
+      Util.createAndAppend('p', rightDiv, { id: 'cont-header', text: 'Contributors' });
 
       contributors
         .map(contributor => new Contributor(contributor))
-        .forEach(contributor => contributor.render(contributorList));
+        .forEach(contributor => contributor.render(rightDiv));
```
Meno-101: Follow to the next comment, but keep in mind, this line is executed repeatedly as long as we have a contributors.

homework/Contributor.js
```diff
@@ -13,7 +13,37 @@ class Contributor {
...
+    const contributorsDiv = Util.createAndAppend('div', container);
+
+    const contributorsLists = Util.createAndAppend('ul', contributorsDiv, { id: 'conts-list' });
+
+    const contributorListItem = Util.createAndAppend('li', contributorsLists, {
```
Meno-101: on line 17 of this file you are creating a `div` as a container, then line 19 you are creating a `ul` as a contributors list**s**, then here on line 21 you are creating an `li` element to have at the end the following from the previous comment:
```HTML
<div>
  <ul>
    <li>contributor #1</li>
  </ul>
  <ul>
    <li>contributor #2</li>
  </ul>
  <!-- till the end of our contributors count -->
</div>
```

A `ul` is for declaring lists on the web page.
when you use it we used to tell the browser and the user `Hey I have a list over here` making the browser knows how to style it, and the user can connect it as it's relations for legibility.

homework/Contributor.js
```diff
@@ -13,7 +13,37 @@ class Contributor {
...
+
+    Util.createAndAppend('a', contributorDiv, {
+      id: `${this.contributor.login}Name`,
+      class: 'cont-name',
+      text: this.contributor.login,
+      href: this.contributor.html_url,
+      target: '_blank',
+    });
+    Util.createAndAppend('div', contributorDiv, {
+      id: `${this.contributor.login}Badge`,
```
Meno-101: you are almost using an `id` on every element you have, `id` we used to use them mostly in Css.

but even in Css we had a note about overusing it, it's good when we have a use case for it.
Also when we don't have a one, we are having an extra `id` I may use it's same value in other place so the browser gets confused over which he have to handle.

ufukcalimli: Actually I have no idea why I used id for every element. So I removed them.

PR #199, opened 2019-02-15
homework week-3

soolaf:


homework/index.js
```diff
@@ -30,18 +15,141 @@
...
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+  function renderRepositories(repository, container) {
+    container.innerHTML = '';
+    const table = createAndAppend('table', container);
+    const tbody = createAndAppend('tbody', table);
+    const tr1 = createAndAppend('tr', tbody);
+    createAndAppend('td', tr1, {
```
Meno-101: An improvment you may apply, is changing the `td` on the hard coded text fields to be `th` as a header for what it represents.

homework/index.js
```diff
@@ -30,18 +15,141 @@
...
+  }
+
+  async function handleError(container, error) {
+    createAndAppend('div', container, {
+      text: error.message,
+      class: 'alert-error',
+    });
+  }
+
+  async function mainPromise(root, repositories) {
```
Meno-101: Hmm, What about `render` as a function name?

homework/App.js
```diff
@@ -16,14 +16,26 @@ class App {
...
     // 2. Make an initial XMLHttpRequest using Util.fetchJSON() to populate your <select> element
 
-    const root = document.getElementById('root');
-
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
-
+    this.root = document.getElementById('root');
+    this.header = Util.createAndAppend('header', this.root);
+    Util.createAndAppend('label', this.header, { text: 'HYF Repositories' });
+    this.select = Util.createAndAppend('select', this.header);
```
Meno-101: `this.select` and `this.header` have no use for them outside of this method, So you may use a plain normal variable as `const` and it will work fine.

homework/App.js
```diff
@@ -16,14 +16,26 @@ class App {
...
+    Util.createAndAppend('div', this.root, {
+      id: 'container',
+    });
     try {
       const repos = await Util.fetchJSON(url);
       this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      repos.forEach((repository, index) => {
```
Meno-101: you have to choose either using `this.repos` or `repos` variable, I think it would be more usefull using the `this.repos` in case of modifing it somehow to keep track on it in the whole program, like sorting or deleting a field or...etc

homework/App.js
```diff
@@ -16,14 +16,26 @@ class App {
...
+    this.root = document.getElementById('root');
+    this.header = Util.createAndAppend('header', this.root);
+    Util.createAndAppend('label', this.header, { text: 'HYF Repositories' });
+    this.select = Util.createAndAppend('select', this.header);
+    Util.createAndAppend('div', this.root, {
+      id: 'container',
+    });
     try {
       const repos = await Util.fetchJSON(url);
       this.repos = repos.map(repo => new Repository(repo));
```
Meno-101: Oops! you are not sorting.

PR #201, opened 2019-02-16
homework Week3

alquinex:


homework/index.js
```diff
@@ -30,15 +32,79 @@
...
   function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+    const root = document.getElementById('root');
+
+    async function refactoring(url1) {
```
Meno-101: A function inside of a function!
this may happen but in a different circumstances,

So what you are trying todo is to refactor something, so from the word refactoring:
What ever you are refactoring doesn't to appear anywhere/anytime else while you are refactoring, means, If am I refactoring a factory in a real life I will have to remove the old one when I'm done refactoring.

From my understanding and our calls, I'm saying you wanted to refactor the ( main ) function, I guess you have only to move this outside of the main function and removing the ( main ) to be this ( refactoring ) -> ( main ) as a names.

While you have done it in front of me, we may have a call about that for you correcting it and telling me why is that... 😇, 
And I asked, you have to clean it up a bit.

homework/index.js
```diff
@@ -30,15 +32,79 @@
...
+      contributorContainer.innerHTML = '';
+
+      const updatedAt = new Date(selectedRepo.updated_at);
+
+      createAndAppend('p', repoContainer, { text: `Repository: ${selectedRepo.name}` });
+      createAndAppend('p', repoContainer, { text: `Description: ${selectedRepo.description}` });
+      createAndAppend('p', repoContainer, { text: `Forks: ${selectedRepo.forks}` });
+      createAndAppend('p', repoContainer, { text: `Update: ${updatedAt.toLocaleString()}` });
+
+      fetchJSON(selectedRepo.contributors_url)
```
Meno-101: This is still using the `.then` chain methedology, what about async/await ?

homework/index.js
```diff
@@ -30,15 +32,79 @@
...
+        text: contributor.login,
+        href: contributor.html_url,
+        target: '_blank',
+      });
+      createAndAppend('p', contributorDetail, { text: contributor.contributions });
+    });
+  }
+
+  function dropDown(root, repos) {
+    const header = createAndAppend('header', root, { class: 'header' });
```
Meno-101: Wohoo! a header tag 😃, I liked it.

homework/index.js
```diff
@@ -30,15 +32,79 @@
...
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+    const root = document.getElementById('root');
+
+    async function refactoring(url1) {
+      try {
+        const repositories = await fetchJSON(url1);
+        dropDown(root, repositories);
+        repositories.forEach(repo => {
```
Meno-101: I'm not sure for what is this `.forEach` for 🙄🤔

PR #209, opened 2019-04-12
Javascript3 week1

Muhannad1989:


homework/index.js
```diff
@@ -27,21 +27,159 @@
...
+      } else {
+        for (let j = 0; j < contributor.length; j++) {
+          const nameOfContributor = contributor[j].login;
+          const numberOfContributor = contributor[j].contributions;
+          const avatarUrl = contributor[j].avatar_url;
+          const cont = document.getElementById('cont');
+          const avatarContainer1 = createAndAppend('div', cont, {
+            class: 'avatar_div',
+          });
+          const avatarContainer2 = createAndAppend('div', cont, { class: 'avatar_div data' });
```
eidosam: Avoid variables numbering by using different and descriptive names, like: `avatarContainer1` -> `avatarContainer` and `avatarContainer2` -> `contributorNameContainer`?

homework/index.js
```diff
@@ -27,21 +27,159 @@
...
+            text: numberOfContributor,
+            class: 'result label',
+          });
+          createAndAppend('br', avatarContainer2, {});
+        }
+      }
+    });
+  }
+
+  function reloadSelectedRepository(valueOfSelectedRepository, selectedData) {
```
eidosam: You could pass the data only, instead of passing the array and index.

homework/index.js
```diff
@@ -27,21 +27,159 @@
...
+    createAndAppend('div', parent, { id: 'info', class: 'info' });
+    createAndAppend('div', parent, { id: 'cont', class: 'cont' });
+
+    const selectRepository = createAndAppend('select', pre, { id: 'select', class: 'select' });
+
+    for (let i = 0; i < data.length; i++) {
+      createAndAppend('option', selectRepository, { text: data[i].name, value: i });
+    }
+    // sort list of repository's
+    sortList();
```
eidosam: As mentioned above, you could sort the data before rendering, it is simpler.

    const reposSorted = data.sort((r1, r2) => r1.name.localeCompare(r2.name));
    for (let i = 0; i < reposSorted.length; i++) {
      createAndAppend('option', selectRepository, { text: reposSorted[i].name, value: i });
    }

homework/index.js
```diff
@@ -27,21 +27,159 @@
...
       const root = document.getElementById('root');
+      root.className = 'root';
+      const pre = createAndAppend('pre', root, { class: 'pre', id: 'pre' });
+      createAndAppend('span', pre, { text: 'HYF Repositories', class: 'logoName' });
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        // we can use below code to sort list of repository instead of sort function above
+        // data.sort((x, y) => x.name.localeCompare(y.name));
```
eidosam: Just saw this comment after commenting on the same idea above. Yes, better to sort before rendering.

homework/index.js
```diff
@@ -27,21 +27,159 @@
...
+          break;
+        }
+      }
+      if (shouldSwitch) {
+        b[i].parentNode.insertBefore(b[i + 1], b[i]);
+        switching = true;
+      }
+    }
+  }
+
```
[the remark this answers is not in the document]
Muhannad1989: Hi Samir, 
Thank you for checking my homework, i see there are a lot of mistakes here. But i will try fix them all and avoid to do that again. This is my big problem. I am not used to checking my common mistake before i hand my homework, Its a bad practice. But i will try to avoid that in futures homework.

Best regards,

homework/index.js
Muhannad1989: Which data do you mean here? I used data by many functions to have access to scope a another which means don't need to use closure functions. Does that make sense?
1- scoops 
2-callbacks: call function inside another instead of using closure functions.
I am actually not sure about those cases but i am trying with my code to get the whole image. If this explanation not the answer to you question, I hope to hear from you about that.

Thanks

eidosam: What I meant by data is the repo JSON data, where you change the function signature to 

    function reloadSelectedRepository(repo)

So `selectedData[valueOfSelectedRepository]` will be replaced with `repo`

And the call in https://github.com/Muhannad1989/JavaScript3/blob/d9aba0940f5dd5bdddc82c3e2e5fb4b3a167df0d/homework/index.js#L159 will be changed to `reloadSelectedRepository(data[0]);` and the same for https://github.com/Muhannad1989/JavaScript3/blob/d9aba0940f5dd5bdddc82c3e2e5fb4b3a167df0d/homework/index.js#L164 the function call will be changed to `reloadSelectedRepository(data[indexOfSelectedRepo]);`

homework/index.js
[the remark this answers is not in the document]
eidosam: Hi Muhannad, no problem at all, we are here to learn from each other.

homework/index.js
Muhannad1989: I did it , it has now only one parameter. Thanks

PR #211, opened 2019-04-12
 Single Page Application (SPA)

ghifo:


homework/index.js
```diff
@@ -29,19 +29,149 @@
     });
     return elem;
   }
+  // ==========start===========
+  /* cSpell: disable */
+  const repoInfo = (index, repoName) => {
```
eidosam: You could send the repo data instead of passing both array and index.

homework/index.js
```diff
@@ -29,19 +29,149 @@
...
+    const container = createAndAppend('div', document.getElementById('root'), {
+      class: 'container',
+      id: 'container',
+    });
+
+    // left div
+    const leftDiv = createAndAppend('div', container, {
```
eidosam: You could name it `repoContainer` instead?

homework/index.js
```diff
@@ -29,19 +29,149 @@
     });
     return elem;
   }
+  // ==========start===========
+  /* cSpell: disable */
```
eidosam: Please enable **Spell Checker** and fix spelling errors if there is any.

wouterkleijn:
Dear @eidosam Thanks for your comments. @ghifo would be great if you could reply to some of the comments and make the necessary improvements - thanks!

ghifo:
Sure!

ghifo:
I know I ought to ask that before,


@eidosam <https://github.com/eidosam>


I just wanted to discuss the comments today if possible!

homework/index.js
ghifo: I tried to do so but I got 'Network request failed'

homework/index.js
ghifo: OK

homework/index.js
ghifo: I did it at the end of the code, I used naming which are considered as spell error. I will avoid that next time.

PR #212, opened 2019-04-12
Akif_JS3W1_Homework

Akif2709:


homework/index.js
```diff
@@ -29,19 +29,95 @@
...
+    createAndAppend('p', rightdiv, { id: 'rightp', text: 'Contributors' });
+    const ul = createAndAppend('ul', rightdiv);
+    function contributors() {
+      fetchJSON(data[select.value].contributors_url, (err, dt) => {
+        if (err) {
+          createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        } else {
+          removeChildren(ul);
+          for (let i = 0; i < dt.length; i++) {
+            const href = createAndAppend('a', ul, { href: dt[i].html_url, target: '_blank' });
```
paulbremer: You are appending a `a` tag to the `ul` and inside the `a` you put the `li` tag, should be the other way around!

homework/index.js
```diff
@@ -29,19 +29,95 @@
...
+    contributors();
+    select.onchange = contributors;
+  }
 
   function main(url) {
     fetchJSON(url, (err, data) => {
       const root = document.getElementById('root');
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        renderError(err);
```
paulbremer: Because you are making new functions here you are not using the `root` const you made, which is giving you an error on the linter check. You should run `npm run lint` before you make a pull request to find these little mistakes.

Akif2709: i am using root inside the 'if' for appending 'div'. But anyway i deleted it in second week homework.

homework/index.js
Akif2709: i changed it and put the 'a' tag inside the 'li' tag. I hope you meant that.

homework/index.js
Akif2709: it is done 👍

homework/index.js
paulbremer: Yes!

wouterkleijn:
Super, thanks for the comments @paulbremer. Great job on implementing the feedback @Akif2709 !

PR #216, opened 2019-04-14
Ramazan Ince JS3 HW1 

RamazanInce:


homework/index.js
```diff
@@ -3,7 +3,9 @@
 {
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
+    // "Authorization: token f55126d198f3c6e9e7ef74e04919e059b82a231b"
```
yash-kapila: Hi Ramazan - you don't need to have this comment here 🙂

homework/index.js
```diff
@@ -30,17 +32,94 @@
...
+        cellContent.textContent = repo[content[i].attribute];
+      }
+    }
+    fetchJSON(repo.contributors_url, (error, contributors) => renderContributors(contributors));
+  }
+
   function main(url) {
     fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
+      const root1 = document.getElementById('root');
```
yash-kapila: Why did you have to rename the variable here as `root1` instead of `root`?

homework/index.js
```diff
@@ -30,17 +32,94 @@
...
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        createAndAppend('div', root1, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        let optionArr = data;
+        optionArr = optionArr.sort((a, b) => a.name.localeCompare(b.name));
+
+        option(optionArr);
+
+        const e = document.getElementById('selector');
```
yash-kapila: Name of the variable isn't the best here. It's hard to understand what `e` represents. 🙂

homework/index.js
```diff
@@ -30,17 +32,94 @@
...
     fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
+      const root1 = document.getElementById('root');
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        createAndAppend('div', root1, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        let optionArr = data;
+        optionArr = optionArr.sort((a, b) => a.name.localeCompare(b.name));
```
yash-kapila: Just FYI - `sort` method here is a mutable method i.e it mutates the array on which it was called. What it means is that if you don't reassign `optionArr` again on line 108, it would still contain the sorted array.

```
const optionArr = data;
optionArr.sort((a,b) => a.name.localeCompare(b.name));

// try console logging optionArr next and verify
console.log(optionArr);
```

homework/index.js
```diff
@@ -30,17 +32,94 @@
     return elem;
   }
 
+  const root = document.getElementById('root');
```
yash-kapila: Hi Ramazan, ideally, we should put all this dangling logic(line 35 - 57) inside a function like main and then if another function like `renderRepo` or `renderContributors` require any of these variables, they should be passed as arguments to it. For example,

`listContributor` is a variable which is being passed from `main` to `renderRepo` function. Thus, the declaration of that variable on line 49 can be moved inside the main function along with other declarations which it needs.

homework/index.js
```diff
@@ -30,17 +32,94 @@
...
+  createAndAppend('p', header, { text: 'HYF Repositories' });
+
+  const select = createAndAppend('select', header, { id: 'selector' });
+  const option = x => {
+    for (let i = 0; i < x.length; i++) {
+      createAndAppend('option', select, { text: x[i].name, value: i });
+    }
+  };
+
+  const div = createAndAppend('div', root, { id: 'container' });
```
yash-kapila: Hi Ramazan, the name here isn't the best again. If you remember our class from JS1 about naming conventions, variable names should be nouns while function names as verbs describing an action. Please take care of it next time onwards. 🙂

yash-kapila:
Another thing please - I see that you force pushed a commit to this branch. What was the need for it?

wouterkleijn:
Thanks @yash-kapila for your super detailed feedback! @RamazanInce Can you have a look at the comments and address them? Thanks!

homework/index.js
RamazanInce: Let me delete and see what will happen.

RamazanInce: We added that line for us to  browse when many users try to enter the web site

homework/index.js
RamazanInce: It did not work that time I thought that was the reason, But I also changed some other things may be that changed effect it.
Let me update it end see what will happen.

RamazanInce: instead of root1 just I rote as root .that is what you mean?

homework/index.js
RamazanInce: I forget to change it, when I was trying it for it to be quick I just named it as e, let me change it.

homework/index.js
M3kH: 🚨Please @RamazanInce DELETE this token from your authorisation list. This is a security concern.

homework/index.js
RamazanInce: Ok I delete it ...

homework/index.js
RamazanInce: I tried it but after that the code did not work in a good way.

homework/index.js
RamazanInce: I rename it as mainContainer

RamazanInce:
> Another thing please - I see that you force pushed a commit to this branch. What was the need for it?

I try several times to push the homework but always gave an error, then google search advice me to push it by force then I did it in that way letter on with the help of Jim I could able to push it.

RamazanInce:
> Thanks @yash-kapila for your super detailed feedback! @RamazanInce Can you have a look at the comments and address them? Thanks!

Thanks for your kind remainder, I finished it now, sorry for the late reply.

homework/index.js
RamazanInce: I changed it add push it again.
that will be okey?
thanks

RamazanInce:
Dear @yash-kapila, @wouterkleijn and @M3kH thanks a lot for your contribution.  I am happy to work such a wonderful people like you.

homework/index.js
yash-kapila: Yes. That's correct. You can rename it back to `root`.

homework/index.js
yash-kapila: Maybe you can put it as a task for your homework separately. When you think you are done with your future homeworks, just go through the code you have written and see if there is any room for improvement. I use this approach regularly at work. 🙂

homework/index.js
eidosam: Not only in the code, it has to be deactivated from your GitHub account, Go to https://github.com/settings/tokens and delete it.

RamazanInce: Let me do it in that way

wouterkleijn:
Thanks for the follow-up @RamazanInce

PR #218, opened 2019-04-18
Javascript3 week2

Muhannad1989:


homework/index.js
```diff
@@ -27,21 +30,138 @@
...
+      createAndAppend('span', liDescription, { text: 'Description: ' });
+      createAndAppend('span', liDescription, { text: getDescription, class: 'result' });
+    }
+
+    const link = createAndAppend('li', info, {});
+    createAndAppend('span', link, { text: 'link: ' });
+    createAndAppend('a', link, {
+      text: dataName,
+      target: '_blank',
+      href: `https://github.com/HackYourFuture/${info}`,
```
eidosam: The link to repo on GitHub is in `selectedData[valueOfSelectedRepository].html_url`, but if you still want to compose it; you need to replace `info` with `dataName`.

homework/index.js
```diff
@@ -27,21 +30,138 @@
...
+    }
+    // reload default repository
+    reloadSelectedRepository(0, data);
+
+    // reload selected repository way (1)
+    // selectRepository.onchange = function reloadOnceSelect() {
+    // const indexOfSelectedRepo = this.value;
+    // reloadSelectedRepository(indexOfSelectedRepo, data); // comment this line to stop getting information once change select list
+    // };
+    // reload selected repository way (2)
```
eidosam: Unused code should be removed

homework/index.js
```diff
@@ -27,21 +30,138 @@
...
+    function reloadOnceSelect() {
```
eidosam: Try to avoid nested functions.

homework/index.js
```diff
@@ -27,21 +30,138 @@
...
-      }
+  // fetch url contributor
+  function fetchContributorsUrl(contributor) {
+    const cont = document.getElementById('cont');
+    cont.innerHTML = '';
+    fetchJSON(contributor)
+      .then(myContributor => {
+        for (let j = 0; j < myContributor.length; j++) {
+          const nameOfContributor = myContributor[j].login;
+          const numberOfContributor = myContributor[j].contributions;
```
eidosam: Should be `numberOfContributions`
Also the link can be found in `myContributor[j].html_url`, no need to compose it manually.

homework/index.js
```diff
@@ -27,21 +30,138 @@
...
+
+  function reloadSelectedRepository(valueOfSelectedRepository, selectedData) {
+    const info = document.getElementById('info');
+    info.innerHTML = '';
+    const getDescription = selectedData[valueOfSelectedRepository].description;
+    const getForks = selectedData[valueOfSelectedRepository].forks;
+    const getUpdatedAt = selectedData[valueOfSelectedRepository].updated_at;
+    const dataName = selectedData[valueOfSelectedRepository].name;
+    const getContributorsUrl = selectedData[valueOfSelectedRepository].contributors_url;
+
```
eidosam: The `get` prefix in variable name indicates this variable is function, but it is not the case here, variables should be nouns in general.

homework/index.js
```diff
@@ -1,19 +1,22 @@
...
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
+      const xhr = new XMLHttpRequest();
+      xhr.open('GET', url);
+      xhr.responseType = 'json';
+      xhr.onreadystatechange = () => {
```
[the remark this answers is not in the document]
Muhannad1989: Hey Samir, thanks for your time you give to add this encouraging comments which helps me to improve my skills. 
I changed all bugs and mistakes that you had mentioned and hope they would be as expected. 

Best Regards.

homework/index.js
```diff
@@ -1,19 +1,22 @@
...
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
+      const xhr = new XMLHttpRequest();
+      xhr.open('GET', url);
+      xhr.responseType = 'json';
+      xhr.onload = () => {
+        if (xhr.readyState === 4) {
```
eidosam: Thanks for addressing the comment, one another remark here; that you don't need to check `readyState` code because this event will be triggered on complete, which is the same as 4, see https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/readyState and https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/load_event
Another thing is you need to handle the network error as well, by listening to event `error`, use [onerror](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequestEventTarget/onerror)

PR #221, opened 2019-04-19
javascript3-week2

alaashhadat:


homework/index.js
```diff
@@ -29,16 +32,62 @@
     });
     return elem;
   }
+  function app(x) {
```
Meno-101: `x` could refer to something specific but it's a short hand when you don't want to use the parameter, other wise you may use a different name describing it's content.

homework/index.js
```diff
@@ -29,16 +32,62 @@
...
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+    const list2 = createAndAppend('ul', div1, { text: '' });
+    function renderContributer(cont1) {
+      for (let i = 0; i < cont1.length; i++) {
+        const link = createAndAppend('a', list2, {
```
Meno-101: with this your HTML structure is as the following which makes it invalid HTML structure, the only allowed children for a `ul` are `li`, `ul` or `ol`.
reference: 
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul
look for permitted content below the example.

homework/index.js
```diff
@@ -29,16 +32,62 @@
...
+        });
+        createAndAppend('br', list3);
+        const img = cont1[i].avatar_url;
+        createAndAppend('img', list3, { src: img });
       }
-    });
+    }
+    function innerText() {
+      des.innerText = `Description: ${x[select.value].description}`;
+      if (x[select.value].description === null) {
```
Meno-101: you could have these four lines with the `createAndAppend` line making them five, in only three lines as the following:
```javascript
createAndAppend('li', list, {
  text: `Description: ${x[select.value].description}`,
  style: x[select.value].description ? 'display: block' : 'display: none'
})
```

Same as some of the next few lines.

homework/index.js
```diff
@@ -29,16 +32,62 @@
...
+    if (Number(select.value) === 0) {
+      innerText();
+    }
+    select.onchange = innerText;
+  }
+
+  function main(url) {
+    fetchJSON(url)
+      .then(app)
+      .catch(err => new Error(err.message));
```
Meno-101: When I get an error I don't see any kind of error rendering on the screen telling the user about the current status of the data he is waiting for.

Meno-101:
Thanks for update, I think you have done the best you could 😃, anything else will be coming from experience. :)

alaashhadat:
Thank you for your support.

PR #223, opened 2019-04-23
Enes homework week 2

eneserdek:


eneserdek:
No more eslint problems anymore. Do you think that is there any other parts that needs to be better?

homework/index.js
```diff
@@ -29,19 +31,120 @@
...
+    const tHeads = ['Repository: ', 'Description: ', 'Forks: ', 'Update: '];
+    const table = createAndAppend('table', rightDiv, {
+      id: 'tableOfInformation',
+    });
+    for (let i = 0; i < tHeads.length; i++) {
+      const tr = createAndAppend('tr', table, {});
+      createAndAppend('th', tr, {
+        text: tHeads[i],
+      });
+      if (i === 0) {
```
Meno-101: Why 0? 
every hard coded number like this is called a magic number, and magic numbers aren't good,
let's say it's Friday and you get back home after a long work day you won't remember why it was a 0 not may be 3 or 5 on Monday!

what about: `tHeads[i] === 'Repository: '` ?
or you may have a variable with the name of that magic number! But it have to be all `CAPS` because it's constant.

homework/index.js
```diff
@@ -29,19 +31,120 @@
...
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        createAndAppend('td', tr, {
+          text: arr[0][firstTime[i]],
+          id: `tableData.${i}`,
+        });
+      }
+    }
+    selectMenu.addEventListener('change', event => {
+      for (let i = 0; i < 4; i++) {
```
Meno-101: this 4 is a magic number too.

homework/index.js
```diff
@@ -29,19 +31,120 @@
...
+        const changed = document.getElementById(`tableData.${i}`);
+        if (i === 0) {
+          changed.innerHTML = `<a target = "_blank" href = ${arr[event.target.value].html_url}> ${
```
Meno-101: No need for spaces in between,
```suggestion
          changed.innerHTML = `<a target="_blank" href=${arr[event.target.value].html_url}> ${
```

and you may re-consider and try to avoid `.innerHTML` for a cleaner code and some other reasons as well, otherwise there is no need for any `.appendChild` any more :)

homework/index.js
```diff
@@ -29,19 +31,120 @@
...
+      id: 'left-div',
+      class: 'contained',
+    });
+    createAndAppend('h2', leftDiv, {
+      text: 'Contributors',
     });
+    const ul = createAndAppend('ul', leftDiv, {
+      id: 'contributorList',
+    });
+    function contributions(data) {
```
Meno-101: perhaps moving this function to the file level would make it a lot cleaner.
```javascript
// you have similar to:
function a() {
  // code
  function b() {
    // more code.
  }
  // and more...
}

// my suggestion would be as:
function b(/* what this function need you can pass from here. */) {
  // more code.
}

function a() {
  // code
  // and more....
}

```

homework/index.js
```diff
@@ -29,19 +31,120 @@
...
+      while (ul.firstChild) {
```
Meno-101: Nice use of while loop 🔥

homework/index.js
```diff
@@ -29,19 +31,120 @@
...
+        const listDiv = createAndAppend('div', li, {
+          id: 'contributor-left',
+          class: 'contained',
+        });
+        createAndAppend('img', listDiv, {
+          src: data[i].avatar_url,
+          class: 'avatar',
+        });
+        const p = createAndAppend('p', listDiv, {});
+        p.innerHTML = `<a target="_blank" href=${data[i].html_url}>${data[i].login}</a>`;
```
Meno-101: same comment as line 91, about `.innerHTML`.

PR #226, opened 2019-04-27
JS3 Week1 HW

saidcetinkaya:


Meno-101:
Hello @saidcetinkaya, Thanks for your Homework, before review I would suggest you go over the errors after running the command `npm test` in your terminal.

It's important to understand each, you can skip the `prettier/prettier` errors and warnings.
Please reply to this comment if any isn't clear, either ways when you done you can just mention me to have a complete review over it ;)

homework/index.js
```diff
@@ -30,13 +30,75 @@
     return elem;
   }
 
+  function pageConstructor(data) {
```
Meno-101: `data` or `dt` is a name for a general use, for me if you will make me a choice I will use `data` over `dt`, because it refers to an unknown data somehow.

for a general naming what I think about a name usually is what kind of information it contains.
in this case it's an information about repositories.

homework/index.js
```diff
@@ -30,13 +30,75 @@
...
+    data.sort((a, b) => a.name.localeCompare(b.name));
+    const root = document.getElementById('root');
```
Meno-101: this line has been used in multiple places so what about using it in the top level of your file?
```javascript
  function pageConstructor(data) {
    // Code...
  }
  function main(url) {
    // Code...
  }
```

and in all functions will be the usage of `root` available, also new once will be created will grant that access according the top level scope. ;)

homework/index.js
```diff
@@ -30,13 +30,75 @@
...
+      target: '_blank',
+      id: 'repository',
+    });
+    createAndAppend('li', ul, { text: `Description :${data[0].description}`, id: 'description' });
+    createAndAppend('li', ul, { text: `Forks :${data[0].forks}`, id: 'forks' });
+    createAndAppend('li', ul, { text: `Update :${data[0].updated_at}`, id: 'update' });
+    const contributorDiv = createAndAppend('div', container, { class: 'contributorDiv' });
+    createAndAppend('p', contributorDiv, { text: 'Contributions' });
+    const contributorUl = createAndAppend('ul', contributorDiv, { class: 'contributorUl' });
+    function contributor() {
```
Meno-101: I see this function just in-collapses it's content to be executed the very next line to it
But it became handy on line [88] when we wish to have it called in the event by it self

If I where you I would recommend moving this function to the file level then having some arguments to passing it's mainly used `variables/values/etc...` as a parameters and that is the main reason of functions 
> declering somewhere and calling in other place

making code look shorter for a human read and making `pageConstructor` function less 18 lines long then it currently is, and `contributor` is a small function so what to say about functions around a 100 lines long [which you may see], and that will make the parent function a huge one for one or two things..

homework/index.js
```diff
@@ -30,13 +30,75 @@
...
+      fetchJSON(data[headerSelect.value].contributors_url, (err, dt) => {
+        if (err) {
+          createAndAppend('div', root, { text: err.message, class: 'alert-error' });
```
Meno-101: At the end of your if statement, if you got an error it will give another error!!

**Why is that?**
because the javascript interpreter won't stop on just creating a new element on the page here in line 61, but instead javascript will continue reading until it finishes this `index.js` document.

and the other error speaking about, is because there is no `dt` content but it will be `undefined` _(no response received)_ and that means it will have no content to loop through

**solution?** just extend your `else` close to the end of the function to contain every appending and it will prevent the error 😃

homework/index.js
```diff
@@ -30,13 +30,75 @@
...
+          const aForLi = createAndAppend('a', li, { href: dt[i].html_url, target: '_blank' });
+          createAndAppend('img', aForLi, { src: dt[i].avatar_url });
+          createAndAppend('p', aForLi, { text: dt[i].login });
+          createAndAppend('p', aForLi, { text: dt[i].contributions, class: 'contNumber' });
+        }
+      });
+    }
+    contributor();
+    headerSelect.addEventListener('change', event => {
+      const change = data[event.target.value];
```
Meno-101: isn't this the repository? 🤔💭

Never mind but just variable naming have to be meaningful to it's content, you did your best but more on that with experience.

homework/index.js
```diff
@@ -30,13 +30,75 @@
...
+      id: 'repository',
+    });
+    createAndAppend('li', ul, { text: `Description :${data[0].description}`, id: 'description' });
+    createAndAppend('li', ul, { text: `Forks :${data[0].forks}`, id: 'forks' });
+    createAndAppend('li', ul, { text: `Update :${data[0].updated_at}`, id: 'update' });
+    const contributorDiv = createAndAppend('div', container, { class: 'contributorDiv' });
+    createAndAppend('p', contributorDiv, { text: 'Contributions' });
+    const contributorUl = createAndAppend('ul', contributorDiv, { class: 'contributorUl' });
+    function contributor() {
+      fetchJSON(data[headerSelect.value].contributors_url, (err, dt) => {
```
Meno-101: `dt`? 🤔
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dt

PR #229, opened 2019-05-02
Week3

alaashhadat:


homework/index.js
```diff
@@ -29,16 +32,69 @@
     });
     return elem;
   }
+  function app(repositories) {
+    repositories.sort((a, b) => a.name.localeCompare(b.name));
```
yash-kapila: Although this logic works well and fulfils all requirements, I think there is still some room for improvement. Ideally, we should write functions in such a way that it is responsible for only one piece of action. For example, the already available functions `createAndAppend` and `fetchJSON` are such functions which are doing only one piece of work.

However, the `app` function here is doing multiple things. It is responsible for:
1. Creating the app layout by calling `createAndAppend` for different tags.
2. Defining logic for putting text into the layout created
3. Fetching contributors information
4. Displaying the fetched contributors information.

I think we can split all these operations in their own functions(which you are already doing) and they need not be contained within the `app` function. Thus, I would suggest that you give it a try and see if you can move them out of `app` function.

For example, `renderContributors` and `innerText` can be moved out of `app` function and if they are using any variable from defined inside `app`, then you can pass those variables as function parameters to them. `innerText` uses `des`, `fork`, `up`, `list2` and `repositories` variables which can be set as function parameters when we move out `innerText` out of the `app` function.

Drop me a message on Slack if you have any questions around it 🙂

homework/Contributor.js
```diff
@@ -14,6 +14,22 @@ class Contributor {
    */
   render(container) {
     // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.contributor, null, 2));
+    const list2 = Util.createAndAppend('li', container);
+    Util.createAndAppend('a', list2, {
+      target: ' _blank',
+      href: `${this.contributor.html_url}`,
+    });
+    const list3 = Util.createAndAppend('p', list2, {
```
yash-kapila: The variable names here could be descriptive 🙂

homework/Contributor.js
```diff
@@ -14,6 +14,22 @@ class Contributor {
...
+      text: `${this.contributor.login}  ${this.contributor.contributions} `,
+    });
+    Util.createAndAppend('br', list2);
+    const img = this.contributor.avatar_url;
+    Util.createAndAppend('img', list3, { src: img });
+    list2.addEventListener('keyup', contributor => {
```
yash-kapila: Why did you have to add the `keyup` event listener here? 🙂

homework/Contributor.js
```diff
@@ -14,6 +14,22 @@ class Contributor {
    */
   render(container) {
     // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.contributor, null, 2));
+    const list2 = Util.createAndAppend('li', container);
+    Util.createAndAppend('a', list2, {
```
yash-kapila: I don't think the way you are adding HTML elements here is the best way to do. Because if you inspect the app on browser, the anchor(`a`) tag is added as a sibling to the paragraph element(`p`) and has got no height or width of its own. And hence it is not clickable from mouse. You can only get to the anchor element through keyboard. Could you please check this once? 🙂

alaashhadat: @yash-kapila  thank you for commenting on my PR. 
i have made the requested changes.

wouterkleijn:
Thanks @alaashhadat for following up, much appreciated. @yash-kapila Can you have another quick look and see if we are done or more changes are needed?

PR #230, opened 2019-05-02
Javascript3 week3

Muhannad1989:


homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+      });
+    }
+  }
 
+  const HYF_REPOS_URL = `https://api.github.com/orgs/HackYourFuture/repos?per_page=100`;
+
+  function main(url) {
+    const root = document.getElementById('root');
+    const pre = createAndAppend('pre', root, { class: 'pre', id: 'pre' });
+    createAndAppend('span', pre, { text: 'HYF Repositories', class: 'logoName' });
```
yash-kapila: Ah. The `class` attribute value should preferably be in `kebab-case` rather than `camelCase` i.e it should be `logo-name` instead of `logoName` 🙂

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+    const selector = createAndAppend('select', pre, { id: 'select', class: 'select' });
+    allData.forEach((data, index) => {
+      createAndAppend('option', selector, { text: data.name, value: index });
+    });
+    // reload default repository while reloading page.
+    reloadAndRenderAllData(allData[0]);
+
+    // reload repository once selected a repository.
+    selector.addEventListener('change', function changeMe() {
+      reloadAndRenderAllData(allData[this.value]);
```
yash-kapila: 👍

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+    const parent = createAndAppend('div', parenElement, { class: 'bodyInformation' });
+    createAndAppend('div', parent, { id: 'info', class: 'info' });
+    createAndAppend('div', parent, { id: 'cont', class: 'cont' });
+
+    const selector = createAndAppend('select', pre, { id: 'select', class: 'select' });
+    allData.forEach((data, index) => {
+      createAndAppend('option', selector, { text: data.name, value: index });
+    });
+    // reload default repository while reloading page.
+    reloadAndRenderAllData(allData[0]);
```
yash-kapila: 👍

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+    createAndAppend('span', liOfForks, { text: fork, class: 'result' });
+    const liUpdateAt = createAndAppend('li', info, {});
+    createAndAppend('span', liUpdateAt, { text: 'updated_at: ' });
+    createAndAppend('span', liUpdateAt, { text: updated, class: 'result' });
+
+    // call all contributors for this selected repository
+    fetchContributorsAndRender(contributorsUrl);
+  }
+
+  function reloadAllRepositories(parenElement, allData) {
```
yash-kapila: I am not sure if this is the best name for this function. We are not reloading any repositories data here if I am not wrong. Instead, we are creating the dropdown and creating layout for displaying `bodyInformation`, right?

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+    const pre = document.getElementById('pre');
+    const parent = createAndAppend('div', parenElement, { class: 'bodyInformation' });
```
yash-kapila: `body-information` instead of `bodyInformation` 🙂

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+      text: dataName,
+      target: '_blank',
+      href: `https://github.com/HackYourFuture/${dataName}`,
+      class: 'result',
+    });
+    const liOfForks = createAndAppend('li', info, {});
+    createAndAppend('span', liOfForks, { text: 'Forks: ' });
+    createAndAppend('span', liOfForks, { text: fork, class: 'result' });
+    const liUpdateAt = createAndAppend('li', info, {});
+    createAndAppend('span', liUpdateAt, { text: 'updated_at: ' });
```
yash-kapila: The `updated_at` date which is retrieved from the GitHub API comes in ISO 8601 format i.e 2018-04-06T17:18:06Z. This can be shown in a more readable format like `dd/mm/yyyy`.

For this, I would suggest that you explore the Date object of JavaScript and see if you can make this appear on UI in `dd/mm/yyyy` format - https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Date

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+  function reloadAndRenderAllData(repository) {
+    const info = document.getElementById('info');
+    info.innerHTML = '';
+    const repositoryDescription = repository.description;
+    const fork = repository.forks;
+    const updated = repository.updated_at;
+    const dataName = repository.name;
+    const contributorsUrl = repository.contributors_url;
+
+    if (repositoryDescription == null) {
```
yash-kapila: == or ===? 😉

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+
+    const link = createAndAppend('li', info, {});
+    createAndAppend('span', link, { text: 'link: ' });
+    createAndAppend('a', link, {
+      text: dataName,
+      target: '_blank',
+      href: `https://github.com/HackYourFuture/${dataName}`,
+      class: 'result',
+    });
+    const liOfForks = createAndAppend('li', info, {});
```
yash-kapila: A better name would be `forksList` here I think. If you remember JavaScript1 naming conventions(https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md), variable names should ideally be nouns while function names should be verbs representing the action being performed inside it.

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
-      }
+  // fetch contributor
+  async function fetchContributorsAndRender(contributorUrl) {
+    try {
+      const contributors = await fetchJSON(contributorUrl);
```
yash-kapila: I liked how you have split logic of your program in separate functions and each function doing one thing only. 🙂

homework/style.css
```diff
@@ -1,3 +1,147 @@
...
+  background-color: blueviolet;
+  padding-bottom: 15px;
+  width: 100%;
+  display: flex;
+  flex-wrap: wrap;
+  justify-content: space-around;
+  background: rgba(15, 15, 15, 0.8);
+  margin-bottom: 2vw;
+}
+.select {
```
yash-kapila: The dropdown responsiveness breaks on mobile resolutions 🙁

homework/App.js
```diff
@@ -11,21 +11,40 @@ class App {
...
+  async initialize(url) {
     const root = document.getElementById('root');
-
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
-
+    const pre = Util.createAndAppend('pre', root, { id: 'pre', class: 'pre' });
+    Util.createAndAppend('span', pre, { text: 'HYF Repositories', class: 'logoName' });
+    Util.createAndAppend('div', root, { id: 'container', class: 'bodyInformation' });
     try {
+      // Fetch all repository's
```
yash-kapila: 👍 for comments

homework/App.js
```diff
@@ -11,21 +11,40 @@ class App {
...
+      // Create instance's for each repository
+      this.repos = repos.map(repo => new Repository(repo)); // to render Repository
+
+      // Start value for each repository inside HTML selector
+      let value = 0;
+
+      // Create and render the Selector
+      const select = Util.createAndAppend('select', pre, { id: 'select', class: 'select' });
+      this.repos.forEach(repo =>
+        Util.createAndAppend('option', select, { text: repo.name(), value: value++ }),
```
yash-kapila: I don't think you need a new variable `value` for the select options. `forEach` callback function has three arguments(element, index and the original array). You can make use `index` instead of `value` here then 🙂

homework/index.js
```diff
@@ -27,21 +31,136 @@
...
+        class: 'alert-error',
+      });
+    }
+  }
 
+  const HYF_REPOS_URL = `https://api.github.com/orgs/HackYourFuture/repos?per_page=100`;
+
+  function main(url) {
+    const root = document.getElementById('root');
+    const pre = createAndAppend('pre', root, { class: 'pre', id: 'pre' });
```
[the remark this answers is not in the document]
Muhannad1989: No, There in no any reason to write it in the code but It was already exist there and i continue creating my elements under it ;)

homework/index.js
Muhannad1989: Yes, you are right here, thank you Yash 🙂

homework/App.js
Muhannad1989: Yes you are right. when i was writhing this code i tried to use the index but it didn't work, then i switched to try using any simple way to skip my issues and look to another. I have forgot to go back to fix it. 
But now thanks your comments, i would try it out and fix it 🙂.

wouterkleijn:
Well done @Muhannad1989 Thanks for reviewing @yash-kapila

PR #231, opened 2019-05-02
Wadeea js3 homework week3

WadeeaKiwan:


homework/index.js
```diff
@@ -30,15 +32,138 @@
...
+  }
+
+  async function main(url) {
+    const root = document.getElementById('root');
+
+    // Creating the header and the select menu
+    const header = createAndAppend('header', root, {
+      class: 'header',
+    });
+    createAndAppend('h2', header, {
```
yash-kapila: 👍 for using the `header` tag for the heading section. However, I would suggest that you use `h1` tag instead of `h2` for the header as it is more semantic and is also important for accessibility. Usually, we shouldn't have `h2` tags without any `h1` tags in the document and similarly `h3` without `h2` and so on.

homework/index.js
```diff
@@ -30,15 +32,138 @@
...
+      text: 'HYF Repositories ',
+      class: 'nav-title',
     });
+    const selectMenu = createAndAppend('select', header, {
+      id: 'repo-selector',
+    });
+    try {
```
yash-kapila: Although this logic works just fine, I would suggest that you split your `main` function a bit further as well. Currently, it is doing multiple things like creating the layout for header, fetching repository details and creating the dropdown. I would suggest that you can also move fetching the repository details in a separate function outside and call that from `main`.

homework/index.js
```diff
@@ -30,15 +32,138 @@
...
+      const hyfRepos = await fetchJSON(url);
+      // Sorting the list of repositories alphabetically on repository name.
+      hyfRepos.sort((a, b) => a.name.localeCompare(b.name));
+
+      // Pushing the repository name to the option element.
+      hyfRepos.forEach((hyfRepo, index) => {
+        createAndAppend('option', selectMenu, {
+          text: hyfRepo.name,
+          value: index,
```
yash-kapila: 👍

homework/index.js
```diff
@@ -30,15 +32,138 @@
...
+      const leftDiv = createAndAppend('div', container, { class: 'left-div' });
+      const rightDiv = createAndAppend('div', container, { class: 'right-div' });
+      createHyfRepoDetails(leftDiv, hyfRepos[0]);
+      createContributors(rightDiv, hyfRepos[0].contributors_url);
+
+      // Adding event listener
+      selectMenu.addEventListener('change', () => {
+        leftDiv.innerText = '';
+        rightDiv.innerText = '';
+        const index = selectMenu.value; // or it could be used with a method of select .selectedIndex
```
yash-kapila: 👍

homework/index.js
```diff
@@ -30,15 +32,138 @@
...
+    });
+    if (hyfRepo.description === null) {
+      document.querySelector('.description-row').setAttribute('class', 'hide-description-row');
+    } else {
+      document.querySelector('.description-data').innerText = hyfRepo.description;
+    }
+    document.querySelector('.forks-data').innerText = hyfRepo.forks;
+    document.querySelector('.updated-data').innerText = new Date(
+      hyfRepo.updated_at,
+    ).toLocaleString();
```
yash-kapila: 👍

homework/index.js
```diff
@@ -30,15 +32,138 @@
...
+        });
+        contributors.forEach(contributor => {
+          const contributorItem = createAndAppend('li', ul, {
+            class: 'contributor-item',
+          });
+          const contributorLink = createAndAppend('a', contributorItem, {
+            href: contributor.html_url,
+            target: '_blank',
+          });
+          const contributorDiv = createAndAppend('div', contributorLink, {
```
yash-kapila: I would say that this isn't the best way to do it. Because an anchor element(`a`) is an inline element while a `div` element is a block level element. Here, you are trying to add a block level element inside an inline element which isn't semantically correct. You can perhaps think of changing `div` to `span` which are inline elements. 🙂

homework/App.js
```diff
@@ -12,18 +12,45 @@ class App {
...
+      id: 'repo-selector',
+    });
+    Util.createAndAppend('main', root, { id: 'container' });
     try {
       const repos = await Util.fetchJSON(url);
       this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+
+      /**
+       * Sort the list of repositories alphabetically on repository name.
```
yash-kapila: 👍 for comments

homework/index.js
WadeeaKiwan: Many thanks @yash-kapila for your time, feedback and remarks 🙂.

Yes, you are right! It's corrected 👍

homework/index.js
WadeeaKiwan: 🙂

homework/index.js
WadeeaKiwan: 🙂

homework/index.js
WadeeaKiwan: 🙂

homework/index.js
WadeeaKiwan: Thanks for this advise. It's done 👍

homework/App.js
WadeeaKiwan: Thanks 🙂

wouterkleijn:
Dear @yash-kapila Thanks for the detailed feedback. @WadeeaKiwan Great job on addressing all the comments. Very professional!

PR #232, opened 2019-05-02
Week3

ghifo:


homework/index.js
```diff
@@ -29,19 +31,162 @@
...
+    const contributors = data => {
+      data.forEach(user => {
+        const li = createAndAppend('li', ul, {
+          class: 'contributor-item',
+        });
+        li.addEventListener('click', () => {
+          window.open(user.html_url, '_blank');
+        });
+
+        createAndAppend('img', li, { src: user.avatar_url, class: 'contributor-avatar' });
```
yash-kapila: There is one thing missing here which is mentioned as part of homework - `Make your app ARIA-compliant (see below).`. If you notice your app, it isn't accessible using the keyboard. Only the repo name(anchor link) is accessible using keyboard. One thing is that the contributors information isn't added as an anchor link but an event listener is being added for capturing any click events. Could you try to add an anchor link using which the browser would automatically add it as part of keyboard navigation.

homework/index.js
```diff
@@ -29,19 +31,162 @@
...
+
+      // Create container
+      createAndAppend('div', root, {
+        class: 'container',
+        id: 'container',
+      });
+      createAndAppend('p', header, {
+        text: 'HYF Repositories',
+      });
+      const list = () => {
```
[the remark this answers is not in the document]
ghifo: Thanks for the tip!

homework/index.js
ghifo: I hope it works now

wouterkleijn:
Great job @ghifo on addressing some of the comments from Yash. Is this PR done or are there still changes to be made? Perhaps you have already discussed this on Slack or last Sunday.

ghifo:
Thanks @Wouter Kleijn <wouter@hackyourfuture.net> ,
There is actually still a huge function that should be broken into some
smaller functions.
I tried that but it seams to me like starting my homework from the very
beginning.
It takes a lot of time and I have to recap the first node (week1 readMe)
homework.
We have already discussed this on Slack, whenever I have extra time, I will
try that.

wouterkleijn:
Ok, great, thanks for your reply @ghifo Let me know when you have updated this.

PR #233, opened 2019-05-02
homework week3

Nadoshka:


homework/promise.js
```diff
@@ -0,0 +1,118 @@
...
+      element.removeChild(element.firstChild);
+    }
+  }
+
+  function renderError(error) {
+    const root = document.getElementById('root');
+    createAndAppend('div', root, { text: error.message, class: 'alert-error', role: 'alert' });
+  }
+
+  function app(repos) {
```
yash-kapila: Hi Nadin - although the logic here works just fine, I think there is still some room for improvement in `app` function. Firstly, if you remember from JavaScript1 about naming conventions(https://github.com/HackYourFuture/fundamentals/blob/master/fundamentals/naming_conventions.md), the function names should describe the logic inside it with a verb. For example, you already have function names like `renderError`, `removeChildren` which are quite descriptive about the function's internals.

Secondly, the `app` function is doing multiple things in it:
1. Creating layout of the application
2. Creating the select dropdown
3. Having a nested function `displayRepoInfo` which in itself is also doing multiple things like displaying repository details and fetching and displaying contributors information.

I think you can split them in separate functions. It is always good to have multiple functions doing one thing only rather than a single function doing too many things.

If you find yourself in a situation where a variable inside a new function isn't available as it was declared in `app`, try passing that as function arguments to the new function. Drop me a message on Slack if you have any questions. 🙂

homework/promise.js
```diff
@@ -0,0 +1,118 @@
...
+    createAndAppend('p', header, {
+      text: 'HYF Repositories',
+    });
+    const select = createAndAppend('select', header, {
+      id: 'list',
+    });
+    for (let i = 0; i < repos.length; i++) {
+      createAndAppend('option', select, { text: repos[i].name, value: i });
+    }
+    const repoContainer = createAndAppend('div', container, { class: 'container-repo' });
```
yash-kapila: I like the names which you have chosen for variables.

homework/promise.js
```diff
@@ -0,0 +1,118 @@
...
+      const repoName = createAndAppend('li', listInfo);
+      repoName.innerHTML = `Repository: <a target="_blank" href= ${repos[select.value].html_url}>${
+        repos[select.value].name
+      }</a>`;
+      createAndAppend('li', listInfo, {
+        text: `Description: ${repos[select.value].description}`,
+      });
+      createAndAppend('li', listInfo, { text: `Forks: ${repos[select.value].forks}` });
+      createAndAppend('li', listInfo, {
+        text: `Updated: ${repos[select.value].updated_at}`,
```
yash-kapila: The `updated_at` date which is retrieved from the GitHub API comes in ISO 8601 format i.e 2018-04-06T17:18:06Z. This can be shown in a more readable format like `dd/mm/yyyy`.

For this, I would suggest that you explore the Date object of JavaScript and see if you can make this appear on UI in `dd/mm/yyyy` format - https://developer.mozilla.org/nl/docs/Web/JavaScript/Reference/Global_Objects/Date

homework/index.html
```diff
@@ -1,23 +1,24 @@
...
-<body>
-  <div id="root"></div>
-  <script src="./index.js"></script>
-</body>
-
-</html>
\ No newline at end of file
+  <body>
+    <div id="root"></div>
+    <script src="./promise.js"></script>
```
yash-kapila: Any specific reason why you decided to go with a new file `promise.js` instead of refactoring `index.js`?

homework/App.js
```diff
@@ -18,17 +18,42 @@ class App {
 
     const root = document.getElementById('root');
 
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
+    const container = Util.createAndAppend('div', root, {
```
yash-kapila: There are still some unneeded comments above this declaration. Could you please remove them? 🙂

homework/App.js
```diff
@@ -18,17 +18,42 @@ class App {
...
+    const select = Util.createAndAppend('select', header, {
+      id: 'list',
+    });
+    select.addEventListener('change', () => {
+      this.fetchContributorsAndRender(select.value);
+    });
+
+    Util.createAndAppend('div', container, { id: 'container-repo' });
 
     try {
```
yash-kapila: You can perhaps think of extracting this logic out in a function as well. Similar to how you have `fetchContributorsAndRender`.

homework/style.css
```diff
@@ -1,3 +1,70 @@
+/* cSpell:disable */
+
 .alert-error {
-    color: red;
-}
\ No newline at end of file
+  color: red;
```
yash-kapila: Hi Nadin - I think the styling of the page can be a bit better as currently, contributors list is appended right after the repository information and it doesn't look nice. Could you please have a look?

homework/App.js
```diff
@@ -18,17 +18,42 @@ class App {
...
       const repos = await Util.fetchJSON(url);
-      this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      this.repos = repos
+        .sort((a, b) => a.name.localeCompare(b.name))
+        .map(repo => new Repository(repo));
+
+      repos.forEach((repo, i) => {
+        Util.createAndAppend('option', select, { text: repo.name, value: i });
```
yash-kapila: 👍

homework/Contributor.js
```diff
@@ -14,6 +14,11 @@ class Contributor {
    */
   render(container) {
     // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.contributor, null, 2));
+    const contributorName = Util.createAndAppend('li', container);
+    contributorName.innerHTML += `<a target ="_blank" href= ${
```
yash-kapila: Is there any reason why you didn't use `createAndAppend` method from Utils class here to create this innerHTML?

homework/Repository.js
```diff
@@ -14,7 +14,17 @@ class Repository {
    */
   render(container) {
     // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.repository, null, 2));
+    const repoName = Util.createAndAppend('li', container);
+    repoName.innerHTML = `Repository: <a target="_blank" href= ${this.repository.html_url}>${
```
yash-kapila: Can you not use `Utils.createAndAppend` here instead?

homework/index.html
Nadoshka: No particular reason! I will change it!
Thanks! :) 👍

homework/promise.js
Nadoshka: Thanks for the link! change is made! :) 👍

homework/promise.js
Nadoshka: Thanks :)

homework/promise.js
Nadoshka: Thanks, I refactored 👍

homework/App.js
Nadoshka: It's done! 👍 :)
Thanks for your feedback!

wouterkleijn:
@Nadoshka thanks for addressing the comments from @yash-kapila ! Is this PR done Yash?

PR #236, opened 2019-05-07
Akif_JS3W3_Homework

Akif2709:


Akif2709:
> Thanks Akif for sending the PR!
> According to assignment specs, the [Classes part](https://github.com/HackYourFuture/JavaScript3/blob/master/Week3/MAKEME.md#222-es6-classes)
> 
> > Repository.js: The Repository class holds code and data for a single repository.
> 
> I your current implementation the `Repository` class represents an array of repos. Can you please follow the requirements, and I will revisit the PR once it is updated.

homework/Repository.js
```diff
@@ -8,25 +8,34 @@ class Repository {
...
 
-  /**
-   * Render the repository info to the DOM.
-   * @param {HTMLElement} container The container element in which to render the repository.
-   */
   render(container) {
-    // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.repository, null, 2));
+    const theads = ['Repository:', 'Description:', 'Forks:', 'Updated:'];
+    for (let i = 0; i < theads.length; i++) {
```
eidosam: You can use `forEach` here instead of regular `for`

homework/Repository.js
```diff
@@ -8,25 +8,34 @@ class Repository {
...
+      const tr = Util.createAndAppend('tr', container);
+      Util.createAndAppend('th', tr, { text: theads[i] });
+      const tds = ['name', 'description', 'forks', 'updated_at'];
```
eidosam: This array can be moved out of the loop, here it is created on each iteration.

homework/Repository.js
```diff
@@ -8,25 +8,34 @@ class Repository {
...
+      if (tds[i] === 'name') {
+        const linked = Util.createAndAppend('td', tr);
+        Util.createAndAppend('a', linked, {
+          id: 'linked',
+          text: this.repository[tds[i]],
+          href: this.repository.html_url,
+          target: '_blank',
+          role: `Repository name is ${this.repository.name}`,
```
eidosam: ARIA Role represents the type of the element, see the list of roles supported to choose the proper one to your case
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques#Roles

wouterkleijn:
Great job @Akif2709 !

homework/Repository.js
Akif2709: This 'for loop' here is looping two different arrays, so I suppose it is difficult to use forEach here.

homework/Repository.js
Akif2709: I changed it as you mentioned. Thanks for the feedback.

PR #237, opened 2019-05-09
Saidcetinkaya js3week2

saidcetinkaya:


homework/index.js
```diff
@@ -30,18 +32,82 @@
...
+    const repository = createAndAppend('li', ul, { text: `Repository :` });
+    createAndAppend('a', repository, {
+      text: `${data[0].name}`,
+      href: data[0].html_url,
+      target: '_blank',
+      id: 'repository',
+    });
+    createAndAppend('li', ul, { text: `Description :${data[0].description}`, id: 'description' });
+    createAndAppend('li', ul, { text: `Forks :${data[0].forks}`, id: 'forks' });
+    createAndAppend('li', ul, { text: `Update :${data[0].updated_at}`, id: 'update' });
```
eidosam: The _updated_at_ date in repo info is in ISO 8601 format, see https://developer.github.com/v3/#schema
Have a look at [JavaScript Date](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date) to find a method that shows the date in readable format.

homework/index.js
```diff
@@ -30,18 +32,82 @@
...
+    const contributorDiv = createAndAppend('div', container, { class: 'contributorDiv' });
+    createAndAppend('p', contributorDiv, { text: 'Contributions' });
+    const contributorUl = createAndAppend('ul', contributorDiv, { class: 'contributorUl' });
+    function contributor(contributorData) {
```
eidosam: Shouldn't be plural? Because this is an array of contributors.

homework/index.js
```diff
@@ -30,18 +32,82 @@
...
+      for (let i = 0; i < contributorData.length; i++) {
```
eidosam: `forEach` is your friend, try to use it instead of regular `for` loop.

homework/index.js
```diff
@@ -30,18 +32,82 @@
...
+      const newA = document.getElementById('repository');
+      newA.textContent = change.name;
+      newA.href = change.html_url;
+      const newDescription = document.getElementById('description');
+      newDescription.textContent = `Description :${change.description}`;
+      const newForks = document.getElementById('forks');
+      newForks.textContent = `Forks :${change.forks}`;
+      const newUpdate = document.getElementById('update');
+      newUpdate.textContent = `Update :${change.updated_at}`;
+
```
eidosam: Smart way to use IDs to quickly select the elements to update, but even better if you create a function to render the repository and use it in the initial state and on change.

homework/index.js
```diff
@@ -30,18 +32,82 @@
...
+      fetchJSON(contributorUrls).then(contributorData => contributor(contributorData));
     });
   }
 
   const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
 
-  window.onload = () => main(HYF_REPOS_URL);
+  fetchJSON(HYF_REPOS_URL)
+    .then(data => pageConstructor(data))
+    .catch(error => handleError(error));
```
eidosam: Better to keep `window.onload` to properly populate the data after the page skeleton is created.
It can be like

      window.onload = () => fetchJSON(HYF_REPOS_URL)
        .then(data => pageConstructor(data))
        .catch(error => handleError(error));

Even better to build on top of existing code, so you call `pageConstructor` from `main`

wouterkleijn:
Thanks @eidosam @saidcetinkaya can you make the necessary changes please? :)

PR #243, opened 2019-06-09
Murad - Week1

Hdvndgr:


homework/index.js
```diff
@@ -30,13 +30,108 @@
...
+    const rightDiv = createAndAppend('div', container, { class: 'right-div whiteframe' });
+    createAndAppend('p', rightDiv, {
+      text: 'Contributions',
+      class: 'contributor-header',
+    });
+    const ul = createAndAppend('ul', rightDiv, { class: 'contributor-list' });
+    const urlCol = rep.contributors_url;
+    fetchJSON(urlCol, (e, dt) => {
+      if (e) {
+        // Error message will be shown
```
remarcmij: Not without some code to render it to the page.

homework/index.js
```diff
@@ -30,13 +30,108 @@
     return elem;
   }
 
+  function update(val, dBase) {
```
remarcmij: What is the meaning of these names: `val` and `dBase`? I need to deduce from looking into your code. I see in line 38 that you use these params to get the repository info to update. And this line 38 is the only place where you use these parameters. It would be better to have just one parameter, `repository`, and call this function with:

```js
update(repos[selector.value]);
```

The `update` function has no need to have access (and therefore should not have access) to the array with **all** repositories. The only one it needs to know about is the one it needs to update. Supplying data on a 'need to know' basis will help to make your code more robust.

homework/index.js
```diff
@@ -30,13 +30,108 @@
...
+          createAndAppend('div', contData, {
+            text: cont.contributions,
+            class: 'contributor-badge',
+          });
+        });
+      }
+    });
+  }
+
+  function initialize(rt, repos) {
```
remarcmij: I had to examine other parts of your code to find out what you mean with `rt`. Avoid cryptic names like this. How much more effort does it take to type the full word `root`. VSCode will auto-complete it for you anyway.

homework/index.js
```diff
@@ -30,13 +30,108 @@
...
+    const header = createAndAppend('header', rt, { class: 'header' });
+    const container = createAndAppend('div', rt, { id: 'container' });
+    createAndAppend('p', header, { text: 'HYF Repositories' });
+    const selector = createAndAppend('select', header, {
+      class: 'repo-selector',
+      'aria-label': 'HYF Repositories',
+    });
+    selector.addEventListener('change', () => {
+      // first remove than update
```
remarcmij: Excellent explanatory comment here.

homework/index.js
```diff
@@ -30,13 +30,108 @@
...
+      }
+    }
+
+    const rightDiv = createAndAppend('div', container, { class: 'right-div whiteframe' });
+    createAndAppend('p', rightDiv, {
+      text: 'Contributions',
+      class: 'contributor-header',
+    });
+    const ul = createAndAppend('ul', rightDiv, { class: 'contributor-list' });
+    const urlCol = rep.contributors_url;
```
remarcmij: What does `Col` mean in `urlCol`? I don't get it. Why do you need this variable anyway? Why not just do:

```js
fetchJSON(rep.contributors_url, (error, contributors) => {
```

Avoid cryptic names such as `e` and `dt` unless it is absolutely clear (not just to yourself) what they stand for.

homework/index.js
```diff
@@ -30,13 +30,108 @@
...
+    fetchJSON(urlCol, (e, dt) => {
+      if (e) {
+        // Error message will be shown
+      } else {
+        dt.forEach(cont => {
```
remarcmij: How much more readable this would be if it said:

```js
```

homework/index.js
```diff
@@ -30,13 +30,108 @@
...
+    const leftDiv = createAndAppend('div', container, { class: 'left-div whiteframe' });
+    const table = createAndAppend('table', leftDiv);
+    const tbody = createAndAppend('tbody', table);
+    const rep = dBase[val];
+
+    for (let ix = 0; ix < 4; ix++) {
+      const row = createAndAppend('tr', tbody);
+      const tdFirst = createAndAppend('td', row, { class: 'label' });
+      const tdSec = createAndAppend('td', row);
+      const updateDate = new Date(rep.updated_at).toLocaleString();
```
remarcmij: We don't need to do this for every loop iteration.

homework/index.js
Hdvndgr: Sorry, I have forgotten to render errors in that line.

homework/index.js
Hdvndgr: val : represents value of select element. 
dBase : represent database (all repos).
I understood it is better if i pass just one repo to update function. 
First, I wrote all codes in initialize then i realized I should create a function to use in case of changing select element. Then I create this function. But it makes me blind to realize these things. 
Thanks for advices.

homework/index.js
Hdvndgr: I used rt instead of root. Because i thought it will throw error if i use the same name in different functions. But i guess there will be no error. Did I understand right?

homework/index.js
Hdvndgr: Thanks.

homework/index.js
Hdvndgr: I was thinking "Collaborators" while i write urlCol. I have forgotten "Contributors". 
Got it I am going to avoid cryptic names.

homework/index.js
Hdvndgr: You are right it is quite clear in this way.

homework/index.js
remarcmij: The 'error' you may get is a style error (reported by ESLint). Your code will still run. But you only get this 'error' if reusing a name that is already available in a higher level scope, for instance:

```js
function foo() {
   const root = ...
   function baz(root) {
                 ^ESLint error : shadowed name
   }
}
```

But this is okay:

```js
function baz(root) {
   ...
}

function foo() {
  const root = ...
  baz(root);
}
```

homework/index.js
Hdvndgr: Right.

homework/index.js
Hdvndgr: I understood. Thank you.

PR #245, opened 2019-06-12
hamza js3-week1 homework

lizozo:


homework/index.js
```diff
@@ -29,14 +29,116 @@
...
+          id: 'updateSpan',
+          text: 'Updated:',
+          class: 'spanClass',
+        });
+        createAndAppend('span', repoDiv, { id: 'repoSpan', text: 'Repository:' });
+        createAndAppend('p', descriptionDiv, {
+          id: 'descP',
+          text: getDivContent('description', data, selectedOption),
+          class: 'pClass',
+        });
```
remarcmij: Rather than setting an `id` attribute on a DOM element just for the purpose that you can reference this element later in your code through `getElementById` it is simpler to just use a JavaScript variable like this:

```js
const descP = createAndAppend('p', descriptionDiv, {
  text: repos[0].description,
  class: 'pClass',
});
```

I'm using `repos[0]` here because that is the initial repo that is selected in the `select` element.

homework/index.js
```diff
@@ -29,14 +29,116 @@
...
+          id: 'forkP',
+          text: getDivContent('forks', data, selectedOption),
+          class: 'pClass',
+        });
+        createAndAppend('p', updatedDiv, {
+          id: 'updateP',
+          text: getDivContent('updated_at', data, selectedOption),
+          class: 'pClass',
+        });
+        createAndAppend('a', repoDiv, {
```
remarcmij: Save this in a JavaScript variable rather than using an `id`:

```js
const repoA =  createAndAppend('a', repoDiv, {
```

homework/index.js
```diff
@@ -29,14 +29,116 @@
...
+          id: 'repoA',
+          href: getDivContent('html_url', data, selectedOption),
+          target: '_blank',
+        });
+        const selectA = document.getElementById('selectId');
+        selectA.onchange = function func() {
+          selectedOption = document.getElementById('selectId').options[
+            document.getElementById('selectId').selectedIndex
+          ].text;
+          document.getElementById('repoA').href = getDivContent('html_url', data, selectedOption);
```
remarcmij: Because you now have a variable `repoA` you no longer need this `getElementById`:

```js
repoA.href = selectedRepo.html_url;
repoA.innerText = selectedRepo.name;
descP.innerText = selectedRepo.description;
```

etc.

homework/index.js
```diff
@@ -29,14 +29,116 @@
...
+          changeInnerText('descP', getDivContent('description', data, selectedOption));
+          changeInnerText('forkP', getDivContent('forks', data, selectedOption));
+          changeInnerText('updateP', getDivContent('updated_at', data, selectedOption));
+          if (getDivContent('description', data, selectedOption) === null) {
+            document.getElementById('descDiv').style.display = 'none';
+          } else document.getElementById('descDiv').style.display = 'flex';
+        };
       }
     });
   }
```
remarcmij: As a general remark for this function, you are first creating some `p` elements and setting their initial value for the first repo when the application starts. Then, whenever the user changes the selection you update those elements with the new value. While that is a valid approach it can quickly become unwieldy because of the many places that you need to keep track of, by assigning `id` attributes or creating JavaScript variables. Another approach would be just to first clear out any `innerHTML` from the `repoDetailsDiv` (i.e. 'wipe the board clean') and then recreate any necessary inner HTML from scratch. This can result in simpler code.

homework/index.js
lizozo: That's so true. I was adding an extra 'id' attribute to reach that element. I just created a new variable for each element and named it same as its 'id'. Now its much more handy. I dont have to call getElementById everytime.

lizozo:
> 
> 
> Hi @lizozo, please go through my comments below and see what you can improve. I take you will add the contributor part next week.

Hello @remarcmij , I have just corrected all suggested change requests. They are appeared in 'commits' section.
1. addOptions function is updated as you suggested.
2. Instead of creating and sorting an array of repo names, first I directly sorted my object and called necessary information (forks, description, html_url...) throught this sorted array just by using attributes(repoNames[selectedIndex].attribute). This is much more easy and reduce codes to type()some unnecessary functions are eliminated such as: getDivContent). also added localeCompare for sorting my object.
3.I set variable names for the elements which I reach by 'Id'.
4, I will definately add contrubutors part in week 2 homework

Lastly I again would like to thank you, because when I go back to my code to correct them, It took some time to understand my own code. I cant even think how you manage to read all of these code snippets from different students and evaluate them one by one. I admire your patience and contribution. We are lucky to have in HFY. Have a nice week!!!

homework/index.js
```diff
@@ -29,14 +29,86 @@
...
       const root = document.getElementById('root');
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        const repoNames = data.sort((a, b) =>
+          a.name.toLowerCase().localeCompare(b.name.toLowerCase()),
+        );
+
+        const header = createAndAppend('header', root, { id: 'header' });
```
remarcmij: The `id` attributes that you set in several places in your code are unneeded. Initially, you had them in there because you needed them for use with `getElementById()`. Now that you have changed your code as per my feedback to use JavaScript variables for accessing the elements created with `createAndAppend` you no longer need them (and therefore you should remove them).

You only need to add an `id` if:

1. You still need, for some reason, to use `getElementById` to get a reference to the element.
2. You need the `id` for CSS styling:

   ```css
   #header {
     ...

If none of these cases apply (and I don't think they do in your case) you should remove the redundant `id` attributes. Leaving them in will leave people looking at your code puzzled why they are there.

PR #246, opened 2019-06-13
Suheyb - Week1

suh3yb:


homework/index.js
```diff
@@ -30,13 +32,126 @@
...
+        const selectMenu = createAndAppend('select', header, {
+          class: 'select-menu',
+          id: 'select-menu',
+        });
+
+        const sortedData = data.sort((a, b) =>
+          a.name.toLowerCase().localeCompare(b.name.toLowerCase()),
+        );
+
+        const elementsToUseInPage = sortedData.map(element => ({
```
remarcmij: Since you haven't reserved to use of the name `repos` or `repositories` elsewhere, this seems a good place to use instead of  `elementsToUseInPage`. You could also chain the `sort` and `map` operations so that you can get rid of the `sortedData` intermediate variable:

```js
const repos = data
  .sort((a, b) => a.name.toLowerCase().localeCompare(b.name.toLowerCase()))
  .map(element => ({
    name: element.name,
    description: element.description,
    forks: element.forks,
    updated: element.updated_at,
    repoUrl: element.html_url,
    contributorsUrl: element.contributors_url,
  }));
```

Later on (line 139), your `forEach` can then become:

```js
repos.forEach((repo, index) => {
  ...
});
```

homework/index.js
```diff
@@ -30,13 +32,126 @@
...
+      }
+    });
+  }
+
   function main(url) {
     fetchJSON(url, (err, data) => {
       const root = document.getElementById('root');
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
```
remarcmij: In case of an error it is best to 'exit early' and leave out an `else`.  In the rest of the code you can now do without the extra indentation, which makes the code more readable.

```js
    if (err) {
      return;
    }
   ...
```

homework/index.js
```diff
@@ -30,13 +32,126 @@
...
+      parentTable.innerHTML = '';
+      showRepoDetailsInTable(obj, parentTable);
+    }
+  }
+
+  function showContributorsAsList(obj, parent) {
+    fetchJSON(obj.contributorsUrl, (error, data) => {
+      if (error) {
+        createAndAppend('div', parent, { text: error.message, class: 'alert-error' });
+      } else {
```
remarcmij: Exit early (see similar comment below).

homework/index.js
```diff
@@ -30,13 +32,126 @@
...
+        parent.innerHTML = '';
+        if (data === null) {
+          createAndAppend('div', parent, { text: 'No Contribution Data.', class: 'alert-error' });
+        } else {
+          createAndAppend('p', parent, { text: 'Contributions', class: 'contributors-header' });
+          const ul = createAndAppend('ul', parent, { class: 'contributors-list' });
+
+          data.forEach(contributor => {
```
remarcmij: If you renamed `data` to `contributors` this line could read:

```js
```

suh3yb:
Hey @remarcmij,
Thank you very much for your time and very detailed feedback. I changed the parts of my code in accordance with your feedback.
Best regards

PR #247, opened 2019-06-13
7alip | Week1

7alip:


homework/index.js
```diff
@@ -16,32 +20,157 @@
...
+          a.name.toLowerCase().localeCompare(b.name.toLowerCase()),
+        );
+        const repositories = sortedData.map(repository => ({
+          name: repository.name,
+          description: repository.description,
+          forks: repository.forks,
+          updated: repository.updated_at,
+          repositoryUrl: repository.html_url,
+          contributorsUrl: repository.contributors_url,
+        }));
```
remarcmij: I saw this similar approach in the homework of @suh3yb. Did you guys work together?

homework/index.js
```diff
@@ -16,32 +20,157 @@
...
-  function createAndAppend(name, parent, options = {}) {
-    const elem = document.createElement(name);
-    parent.appendChild(elem);
-    Object.keys(options).forEach(key => {
-      const value = options[key];
+  function createAndAppend(tag, parent, attributes = {}) {
+    const child = document.createElement(tag);
+    parent.appendChild(child);
+    Object.keys(attributes).forEach(key => {
+      const value = attributes[key];
```
remarcmij: Are the names changes really necessary? If you left the code as is, these changes would have appeared as changes in the pull request, and the reviewer would not have to review these trivial changes.

homework/index.js
7alip: We didn't wotk together.

I tried many different approaches in order to get the expected data. My first approach was create components from an object model. I could create static components but I couldn't pass data to dinamic parts. It was totally different so I gave up after three days and turned back the normal way.

I've viewed some of the other works, also from the previous classes, in order to get inspired. That's why it's similar to one.

homework/index.js
7alip: That's right. It was not necessary to change the ready functions which was already written. I didn't think that point for the reviewer, sorry. It's because maybe I am obssesive about namings 😐

PR #249, opened 2019-06-13
abdullah JS3 week1

heartwhite:
single app made.! by using some extra functions
nom test shows a console .log . but it is in app.js which I did not touch.

all CSS taken from original link " https://hyf-github.netlify.com "

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -29,44 +29,17 @@
     });
     return elem;
   }
-  function tableMaker(data, index) {
-    const leftDiv = document.querySelector('.leftDiv');
-    const table = createAndAppend('table', leftDiv, {});
-    const tBody = createAndAppend('tbody', table, {});
-    const trRepo = createAndAppend('tr', tBody, {});
-    createAndAppend('td', trRepo, { text: 'Repository:' });
-    const tdRepoLink = createAndAppend('td', trRepo, {});
-    createAndAppend('a', tdRepoLink, {
-      text: data[index].name,
-      href: data[index].html_url,
-      target: '_blank',
-    });
-    if (data[index].description !== null) {
-      const trDescription = createAndAppend('tr', tBody, {});
-      createAndAppend('td', trDescription, { text: 'Description:' });
-      createAndAppend('td', trDescription, { text: data[index].description });
-    }
-    const trForks = createAndAppend('tr', tBody, {});
-    createAndAppend('td', trForks, { text: 'Forks:' });
-    createAndAppend('td', trForks, { text: data[index].forks_count });
-    const trUpdate = createAndAppend('tr', tBody, {});
-    createAndAppend('td', trUpdate, { text: 'Updated:' });
-    createAndAppend('td', trUpdate, {
-      text: new Date(data[index].updated_at).toLocaleString(),
-    });
-  }
 
-  function headerMaker(data) {
+  function makeHeader(repos) {
     const root = document.getElementById('root');
     const header = createAndAppend('header', root, { class: 'header' });
     header.style = 'display: flex;';
     createAndAppend('h3', header, { text: 'HYF Repositories' });
     const selectMenu = createAndAppend('select', header, { class: 'repo-selector' });
-    const optionArr = data.map(repository => repository.name);
     const optionsArr = [];
-    for (let i = 0; i < optionArr.length; i++) {
+    for (let i = 0; i < repos.length; i++) {
       const option = {
-        name: optionArr[i],
+        name: repos[i].name,
         index: i,
       };
       optionsArr.push(option);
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -29,14 +29,130 @@
...
+      document.querySelector('.rightDiv').innerHTML = '';
+      const index = selectMenu.options[selectMenu.selectedIndex].value;
+      tableMaker(data, index);
+      fetchJSON(data[index].contributors_url, (err, fetchedData) => {
+        if (err) {
+          createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        } else {
+          ulMaker(fetchedData);
+        }
+      });
```
remarcmij: This code here seems a repetition of what your function `constructForFirst` already does. I would think it should be possible to create a function that you can reuse for the first repo (e.g. `repositories[0]` and the selected repository (`repositories[index]`).

homework/style.css
```diff
@@ -1,3 +1,108 @@
 .alert-error {
```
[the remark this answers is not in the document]
heartwhite: You are right.
I took all css by clicking all html elements and copy from inspect window.
actually I didn't care much about styling (as a homework). but of course we re learning not just technical code writing. I need to learn and apply 'common and right' way.  I updated my comment of PR.

homework/index.js
```diff
@@ -29,14 +29,130 @@
...
+      const li = createAndAppend('li', ul, {
+        class: 'contributor-item',
+        tabindex: '0',
+        'aria-label': contributor.login,
+      });
+      createAndAppend('img', li, {
+        src: contributor.avatar_url,
+        class: 'contributor-avatar',
+        height: '48px',
+      });
```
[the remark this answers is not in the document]
heartwhite: I added 'px' to inline CSS by mistake, now corrected.
and I added alt attribute to images .

homework/index.js
```diff
@@ -29,14 +29,130 @@
     });
     return elem;
   }
+  function tableMaker(data, index) {
```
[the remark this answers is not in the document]
heartwhite: each article of this feedback applied to all functions. and i sincerely say that applying this things is made great affect. once again I read my code and I understood things those I did not understand while I was writing.
now my code is more mine.

homework/index.js
```diff
@@ -29,14 +29,130 @@
...
+    const leftDiv = document.querySelector('.leftDiv');
+    const table = createAndAppend('table', leftDiv, {});
```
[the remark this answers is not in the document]
heartwhite: all corrected

homework/index.js
heartwhite: yes. I was confused and I was stuck. than I solved the 'start' problem with a completely separate function. 
now I corrected that. and again I say, 'I understood things those I did not understand while I was writing'.
thanks for all Jim.

homework/index.js
```diff
@@ -30,13 +30,122 @@
...
+  }
+
+  function mainHtmlConstructor(repos) {
+    makeHeader(repos);
+    const root = document.getElementById('root');
+    const container = createAndAppend('div', root, { id: 'container' });
+    createAndAppend('div', container, { class: 'leftDiv whiteframe' });
+    createAndAppend('div', container, { class: 'rightDiv whiteframe' });
+    const selectMenu = document.querySelector('select');
+    makeHtmlOfContents(repos, selectMenu.options[selectMenu.selectedIndex].value);
```
remarcmij: I overlooked this in my first review: the index you need can be obtained either by `selectMenu.selectedIndex` or `selectMenu.value`. In this case both methods give the same result. It is not recommended (and unneeded) to obtain the index value by accessing the `options` elements. Please review the documentation for these links:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/selectedIndex

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event#%3Cselect%3E_element

homework/index.js
```diff
@@ -30,13 +30,122 @@
     return elem;
   }
 
+  function makeHeader(repos) {
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root, { class: 'header' });
+    header.style = 'display: flex;';
```
remarcmij: Use your CSS file for styling.

homework/index.js
```diff
@@ -30,13 +30,122 @@
...
+    createAndAppend('td', trForks, { text: 'Forks:' });
+    createAndAppend('td', trForks, { text: repo.forks_count });
+    const trUpdate = createAndAppend('tr', tBody);
+    createAndAppend('td', trUpdate, { text: 'Updated:' });
+    createAndAppend('td', trUpdate, {
+      text: new Date(repo.updated_at).toLocaleString(),
+    });
+  }
+
+  function makeUl(repos) {
```
remarcmij: Try to think of more descriptive names. What will this `ul` contain? Apples? Or perhaps Contributors? In the latter case `makeContributorList` would more clearly describe what this function does.

homework/index.js
heartwhite: ok. actually I don't know how it works. I just looked for some samples from google, those were almost same. I used that and it worked. I hope I will read documentation and understand <select>
and I can make changes I think, if you don't advise against

heartwhite: I now understood that my problem was sorting. 
when I used "selectMenu.selectedIndex" directly it works but order is different :) 
eg if I select one repository from Dropdown , table and list filled by another repo :)
and I solved that by sorting the data at the beginning . I sow this in Murat's PR. 
I applied same way. actually now it is easier to read and understand and even write :)
I will push the latest code

homework/index.js
heartwhite: ok.

homework/index.js
heartwhite: it was a code that I wrote for see a little better before I write CSS file. I deleted it.

homework/index.js
remarcmij: Yes, that is exactly right. First sort the repositories array that you get back from GitHub and then create the `<option>` element from the sorted array. In this way you maintain a one-to-one correspondence between the index of the selected option and the index of the corresponding repository object. There is no need to create a separate `optionsArr` array. Just use the original sorted repositories array throughout your code

PR #251, opened 2019-06-14
Single Page Application (SPA) that uses the GitHub API to display inf…

ghifo:
…ormation about the available HYF GitHub repositories.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -29,8 +29,20 @@
     });
     return elem;
   }
+
+  // a function to create td without link
+  function createTd(textKey, textContent, tbody) {
+    const parent = createAndAppend('tr', tbody);
+    createAndAppend('td', parent, {
+      text: `${textKey}: `,
+      class: 'label',
+    });
+    createAndAppend('td', parent, {
+      text: textContent,
+    });
+  }
   // table of repository's basic information
-  function createDetailsTable(index, data, container) {
+  function createDetailsTable(repository, container) {
     const repositoryData = createAndAppend('div', container, { class: 'left-div white-frame' });
     const table = createAndAppend('table', repositoryData);
     const tbody = createAndAppend('tbody', table);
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -29,19 +29,124 @@
...
+    });
+    // manually create td with a link
+    const td = createAndAppend('td', tr);
+    createAndAppend('a', td, {
+      href: data[index].html_url,
+      text: data[index].name,
+      target: '_blank',
+    });
+    // a function to create td without link
+    function createTd(textKey, textContent) {
```
remarcmij: You should avoid nested functions like this. The reason is that nested function obscure where the data comes from that they need. A nested function has access to the data of its outer function. In this case, the `createTd` function accesses the `tbody` variable from the scope of its parent function, but that is not immediately obvious. You can easily move this function out of the `createDetailsTable` function by passing the `tbody` variable as a third parameter, e.g.:

```js
function createTd(textKey, textContent, tbody) {
  ...
}

function createDetailsTable(index, data, container) {
  ...
  createTd('Description', data[index].description, tbody);
  ...
}
```
 Now it is absolutely clear what the `createTd` function needs from the outside world to do its work.

homework/index.js
```diff
@@ -29,19 +29,124 @@
     });
     return elem;
   }
+  // table of repository's basic information
+  function createDetailsTable(index, data, container) {
```
remarcmij: This function only works on a single repository. So rather than passing it through the argument list the complete array of repositories (`data`) and the `index` value which indicates which repository object to use it would be better to just call this function with the actual repository object to use, i.e.:

```js
function createDetailsTable(repo,  container) {
  ...
  createAndAppend('a', td, {
    href: repo.html_url,
    text: repo.name,
    target: '_blank',
  });
  ...
}

...

createDetailsTable(data[0], container);
...
createDetailsTable(data[index], container);
```

Also, use the name `data` only if the variable can represent any data (like internally in the `fetchJSON` function). If you have detailed knowledge about what the variable contains, use that knowledge to pick a more descriptive name. In this case, `data` contains an array of repository objects. So renaming `data` to `repositories` would take some of guessing away of what `data` represents here. (Notwithstanding as I said earlier that it is better to pass a single repository object `repo` here).

homework/index.js
```diff
@@ -29,19 +29,124 @@
...
+  }
+
+  // contributors list
+  function listContributors(index, data, container) {
+    const contributorUrl = data[index].contributors_url;
+    fetchJSON(contributorUrl, (err, contributors) => {
+      const root = document.getElementById('root');
+      if (err) {
+        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+      } else {
```
remarcmij: When dealing with errors like this at the start of the function body, it is best to 'exit early': there is no more useful work to do in the function:

```js
if (err) {
  return;
}

const contributorsData = createAndAppend('div', container, {
...
```

You can now also get rid of the `else` and the corresponding indentation for the rest of the function body.

You can do the same in lines 136-137.

homework/index.js
```diff
@@ -29,19 +29,124 @@
...
+    createDetailsTable(0, data, container);
+    listContributors(0, data, container);
+
+    select.addEventListener('change', event => {
+      while (container.firstChild) {
+        container.removeChild(container.firstChild);
+      }
+      // event.target refers to the clicked select repository
+      // It will fire when each select repository is clicked
+      const index = event.target.value;
```
remarcmij: Alternatively, you can also get this value with `select.value`. But your method is fine too.

homework/index.js
ghifo: OK, It works so

homework/index.js
ghifo: But I still need the `index` and `data` to call the `createTd` function

homework/index.js
ghifo: It works the same way, no need for the `else`.
Thanks for the tip.

homework/index.js
ghifo: Much more easier, thanks

homework/index.js
ghifo: It was containing all the repos, but after putting the `createTd` function apart, it deals with only one repo.
But still need it

remarcmij: I don't think so. You can call your `createTd` function like this, can't you?

```js
createTd('Description', repo.description, tbody);
```

etc.

ghifo: I did it, Thanks

ghifo: I have changed the `listContributors` also ):

homework/index.js
```diff
@@ -42,67 +54,55 @@
...
   // contributors list
-  function listContributors(index, data, container) {
-    const contributorUrl = data[index].contributors_url;
+  function listContributors(repository, container) {
+    const contributorUrl = repository.contributors_url;
     fetchJSON(contributorUrl, (err, contributors) => {
       const root = document.getElementById('root');
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
```
remarcmij: I seem to miss the `return` statement. Without it, the code just continues with the next statement, which is wrong. Instead, your function needs to 'return' (that is what I mean with 'exit early'). Same thing in the other place in your code where you want to 'exit early'.

ghifo: I have solved this too.
Thanks a lot

PR #252, opened 2019-06-14
week1-irfan

irfankaradeniz:


homework/index.js
```diff
@@ -30,13 +30,103 @@
     return elem;
   }
 
+  function removeChildren(element) {
+    while (element.firstChild) {
+      element.removeChild(element.firstChild);
+    }
+  }
+
   function main(url) {
```
remarcmij: 1. The assignment asked for the first repository to be displayed when your application starts up. This is not happening in your case. 
2. Your `main` function is fairly large, which makes it difficult (for others and yourself) to maintain. Your code could benefit from a more granular approach, with a couple of functions for dedicated tasks.

homework/index.js
```diff
@@ -30,13 +30,103 @@
...
+
+        const renewalSpace = createAndAppend('div', root, {
+          id: 'contributors',
+        });
+
+        let valueNumber = 0;
+        const repositories = JSON.parse(jsonString, null);
+        let contributorsURL = '';
+
+        repositories.forEach(repo => {
```
remarcmij: 1. The assignment asked for the repositories to be sorted by name. 
2. Instead of using a separate `valueNumber` variable you can use the optional `index` variable from `forEach`:

   ```js
   repositories.forEach((repo, index) => {
     ...
     value: index,
     ...
   });
   ```

homework/index.js
```diff
@@ -30,13 +30,103 @@
...
+          const options = createAndAppend('option', select, {
+            value: valueNumber,
+          });
+          options.textContent = repo.name;
+          valueNumber += 1;
+        });
+        const selectElement = document.getElementById('selection');
```
remarcmij: There is no need to get a reference to the `select` element like this as you already have it from line 50.

homework/index.js
```diff
@@ -30,13 +30,103 @@
...
+  }
+
   function main(url) {
     fetchJSON(url, (err, data) => {
       const root = document.getElementById('root');
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        const jsonString = JSON.stringify(data, null, 2);
```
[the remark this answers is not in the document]
irfankaradeniz: I didn't get that, Jim. When I tries to get rid one of them, data isnt shown in my web page, select element doesnt work I mean.

homework/index.js
irfankaradeniz: Done!

homework/index.js
irfankaradeniz: Done!

homework/index.js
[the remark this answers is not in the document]
remarcmij: The `data` parameter in the line `fetchJSON(url, (err, data) => {` is already the array of repository objects that you need (console.log it if necessary). The extra step of `JSON.stringify` and `JSON.parse` is unneeded (and therefore should be removed). Since you used the name `repositories` to store the result of the `JSON.parse`, you could just rename the `data` parameter from `fetchJSON` call to `repositories` and remove the lines containing `JSON.stringify` and `JSON.parse`.

homework/index.js
irfankaradeniz: I've shorten my main function at week2's homework. I understood, thanks Jim.

homework/index.js
irfankaradeniz: Now, its sorting repositories by the name. Also figured it out the problem of showing first thing when the page loaded in week2's homework.

PR #254, opened 2019-06-15
wk1 js3 class21

Nadoshka:


homework/index.js
```diff
@@ -30,18 +30,119 @@
...
+        const table = createAndAppend('table', repoContainer, { class: 'table' });
+
+        const tBody = createAndAppend('tbody', table);
+
+        // create row for each contributor
+        contributors.forEach(contributor => {
+          const row = createAndAppend('tr', tBody);
+
+          const imageCell = createAndAppend('td', row);
+          createAndAppend('img', imageCell, { src: contributor.avatar_url, class: 'img' });
```
remarcmij: The HTML validator complains about missing `alt` attributes (the line number below refer to the generated HTML, not to this JavaScript file):

```
Error: An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.
    From line 16, column 2358; to line 16, column 2434
    Error: An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.
    From line 16, column 2537; to line 16, column 2613
    There were errors.
```

homework/style.css
```diff
@@ -1,3 +1,71 @@
 .alert-error {
```
remarcmij: I guess you still have some styling work to do:

image

image

homework/index.js
```diff
@@ -30,18 +30,119 @@
...
+  // create reusable function (removeChildren),
+  // to remove children of an html element which is (the parent).
+  function removeChildren(element) {
+    while (element.firstChild) {
+      element.removeChild(element.firstChild);
+    }
+  }
+
+  // defined root outside the local scope of the function to be available to all functions.
+  const root = document.getElementById('root');
```
remarcmij: It is customary to place global variables at the top of the file (in this case right after the opening curly brace, just before the `fetchJSON` function), so that they are the first thing you see when you open up the file. I'm not fond of global variables, as it is not immediately clear where they are used. Your comment _"defined root outside the local scope of the function to be available to all functions"_ describes exactly the problem: it can be used all over the place.

In this particular case, you also run the risk of the variable being undefined if someone decides to move your `app.js` file from end of the `body` tag to the `head` tag. In that case, when this statement runs your root `div` will not exist yet (that's why we use `window.onload` here).

If you place this line inside your `main` function and pass it as an argument to any function that needs it, you don't run this risk. As far as I can tell you only reference this `root` variable in your function `main` and in your function `createContributor`. The latter is only called by your function `main` so you could easily pass it as a parameter and get rid of the global variable:

```js
unction main(url) {

   ...

   createContributor(selectedRepo, repoContainer, root);
   ...
}
```

homework/index.js
```diff
@@ -30,18 +30,119 @@
...
+  function createContributor(selectedRepo, repoContainer) {
+    // initialize the url of contributor.
+    const contributorUrl = selectedRepo.contributors_url;
+
+    // fetch the url of the contributor and handle error and data of the contributor
+    fetchJSON(contributorUrl, (err, contributors) => {
+      if (err) {
+        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+      } else {
+        // create table
```
remarcmij: Avoid 'noise' comments that explain what is already obvious from reading the code.

homework/index.js
```diff
@@ -30,18 +30,119 @@
...
+  // make a function which creates contributor info as a table.
+  function createContributor(selectedRepo, repoContainer) {
+    // initialize the url of contributor.
+    const contributorUrl = selectedRepo.contributors_url;
+
+    // fetch the url of the contributor and handle error and data of the contributor
+    fetchJSON(contributorUrl, (err, contributors) => {
+      if (err) {
+        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+      } else {
```
remarcmij: It is best to exit early from the function when you have detected an error like this. You can remove the `else` and the extra indentation in the rest of your code:

```js
if (err) {
  return;
}

const table = createAndAppend('table', repoContainer, { class: 'table' });
```

homework/index.js
Nadoshka: Thanks! I will add it soon.

homework/style.css
Nadoshka: Thanks! You are right! not totally complete, I'm working on it.

homework/index.js
Nadoshka: U are right! Thanks a lot for your explanation. It's done!

homework/index.js
Nadoshka: Thanks! it's good to know that! It's done!

PR #257, opened 2019-06-16
Class21/week2

ghifo:


homework/index.js
```diff
@@ -30,17 +32,122 @@
...
+            alt: "contributor's profile",
+            class: 'contributor-avatar',
+          });
+          const liDiv = createAndAppend('div', li, { class: 'contributor-data' });
+          createAndAppend('div', liDiv, { text: contributor.login });
+          createAndAppend('div', liDiv, {
+            text: contributor.contributions,
+            class: 'contributor-badge',
+          });
+        });
```
remarcmij: One of the requirements for ARIA compliance is that you must be able to navigate the UI with the keyboard. At present you cannot go to a contributor list item with the keyboard and activate it to go to the contributor's GitHub page by pressing Enter. The simplest way to achieve that is to wrap the `<li>` content into an `<a>` tag like this:

```js
  const li = createAndAppend('li', ul);
  const anchor = createAndAppend('a', li, {
    href: contributor.html_url,
    target: '_blank',
    class: 'contributor-item',
  });
  createAndAppend('img', anchor, {
    src: contributor.avatar_url,
    alt: "contributor's profile",
    class: 'contributor-avatar',
  });
  const liDiv = createAndAppend('div', anchor, { class: 'contributor-data' });
  createAndAppend('div', liDiv, { text: contributor.login });
  createAndAppend('div', liDiv, {
    text: contributor.contributions,
    class: 'contributor-badge',
  });
});
```

You will need to tweak the `contributor-item` CSS class a bit to get rid of the underline decoration that you normally get for an `<a>` tag.

ghifo: Sure, I’ll do that.
Thanks for your helpful feedback!

ghifo: May I ask you a Q?
Shall I use this way way in place of the `eventListner` ?
```li.addEventListener('click', () => {
            window.open(contributor.html_url);
          });```

remarcmij: You don't need / shouldn't use an event listener at all if you use an `<a>` tag. If your click and `<a>` tag it will automatically go to the `url` specified by its `href` attribute. That is the whole purpose of an `<a>` tag. It is also by default included in the tab order (you can uset the tab key to navigate to an `<a>` tag) and also activates the link if you press Enter while the `<a>` tag has the focus.

ghifo: Good to know!
Thanks for your response!

PR #258, opened 2019-06-17
Abdullah Al Hommada JS3-week2

aalhommada:


aalhommada:
Thanks Jam for your notification,  i is a mistake , i was curious to know how you can review and put notes and refuse and ....in github , but it seems that i made a mistake .

PR #263, opened 2019-06-19
Suheyb - Week2

suh3yb:


homework/index.js
```diff
@@ -30,15 +34,147 @@
...
+      throw new Error(`Error: No parent is specified.`);
+    }
+
+    parents.forRepo.innerHTML = '';
+    parents.forContributors.innerHTML = '';
+    const table = createAndAppend('table', parents.forRepo);
+
+    Object.keys(obj).forEach(key => {
+      const value = obj[key];
+      if (value !== null) {
```
remarcmij: You can return early here in case `value === null` (saves you the indentation and keeping track of where it should end) and may want to consider replacing the multiple `if` statements that test the same variable against a number of constants with a `switch` statement:

```js
Object.keys(obj).forEach(key => {
  const value = obj[key];
  if (value === null) {
    return;
  }
  let tableRow;
  switch (key) {
    case 'name':
      {
        tableRow = createAndAppend('tr', table, { class: 'tr' });
        createAndAppend('td', tableRow, {
          text: 'repository :',
          class: 'title',
        });
        const tdName = createAndAppend('td', tableRow);
        createAndAppend('a', tdName, {
          text: value,
          href: obj.repoUrl,
          target: '_blank',
        });
      }
      break;
    case 'description':
    case 'forks':
      tableRow = createAndAppend('tr', table, { class: 'tr' });
      createAndAppend('td', tableRow, {
        text: `${key} :`,
        class: 'title',
      });
      createAndAppend('td', tableRow, { text: value });
      break;
    case 'updated':
      tableRow = createAndAppend('tr', table, { class: 'tr' });
      createAndAppend('td', tableRow, {
        text: 'updated :',
        class: 'title',
      });
      createAndAppend('td', tableRow, {
        text: `${value.slice(0, 10)}, ${value.slice(11, 19)}`,
      });
      break;
    default:
  }
});
```

homework/index.js
```diff
@@ -30,15 +34,147 @@
...
+          showRepoAndContributorsInDetail(repos[selectMenu.value], {
+            forRepo: leftColumn,
+            forContributors: rightColumn,
+          });
+        });
+      })
+      .catch(error => {
+        createAndAppend('div', root, { text: error.message, class: 'alert-error' });
+      })
+      .then(() => {
```
remarcmij: Nice touch to add a spinner. This is exactly the right place to add the required `then` as it will be executed in the normal case as well as the error case. You also may want to take note of the `.finally()` method that can be used at the end of a promise chain, although at this time it is not yet universally supported by the various browsers.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/finally

suh3yb:
Hey @remarcmij, 
Thank you very much for your time and feedback. I am sorry I reply this late. I replaced the multiple `if` statements with a `switch` statement and although it is not yet universally supported I used `finally()` instead of last `then`s. If I should not use `finally()` for this time because of the lack of support, I can replace them again in the next week's homework.
By the way, I really appreciate the way you give feedback. Thanks again for spending so much time and even rewriting the code if necessary.
Best regards.

PR #264, opened 2019-06-20
7alip | Week2

7alip:


homework/index.js
```diff
@@ -123,79 +125,75 @@
   // Creates html elements from an object
   function createStaticComponentsFromModel(model, root) {
```
7alip: https://github.com/HackYourFuture/JavaScript3/pull/253#discussion_r294050179

I tried to make challenge in line with your recommendations in the link but I noticed that my approach didn't make sense. It should act as virtual dom. Otherwise, whenever I change the select element, it would re-create all the parts. 

I should focus the goal of the homeworks. I accept that my approach was more than enogh. Therefore, I left the model part and changed the creating elements part into traditional approach.

homework/index.js
```diff
@@ -30,17 +34,212 @@
...
+  function init(data, root) {
+    createStaticComponentsFromModel(componentModel, root);
+
+    const repositories = data.sort((a, b) =>
+      a.name.toLowerCase().localeCompare(b.name.toLowerCase()),
+    );
+
+    const select = document.querySelector('.header__select');
+
+    (function listAllRepositoriesIntoSelect() {
```
remarcmij: As I commented in week 1, these immediately invoked functions serve no purpose. I can comment out the function heading and closing parts without affecting the result. It is just confusing to other people trying to understand your code and your reasoning.

```js
// (function listAllRepositoriesIntoSelect() {
repositories.forEach((repository, index) => {
  createAndAppend('option', select, { text: repository.name, value: index });
});

const selectedRepository = repositories[select.selectedIndex];

  document.querySelector('tbody').innerHTML = '';
  document.querySelector('.contributors__list').innerHTML = '';
  fillRepositoryDetailsTable(selectedRepository);
  fillContributorsList(selectedRepository);
});
// })();

// Shows content randomly when the page is initialized
// (function initializePage() {
const initialRepositoryIndex = Math.floor(Math.random() * repositories.length);

select.selectedIndex = initialRepositoryIndex;
fillRepositoryDetailsTable(repositories[initialRepositoryIndex]);
fillContributorsList(repositories[initialRepositoryIndex]);
// })();
```

homework/index.js
```diff
@@ -30,17 +34,212 @@
     return elem;
   }
 
-  function main(url) {
-    fetchJSON(url, (err, data) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+  // You can easily modify static html elements with this model
+  const componentModel = [
```
remarcmij: While this is a creative alternative and original piece of work for defining the DOM structure, it is not necessarily better than the other approach, i.e. using direct calls to the DOM. To function as a virtual DOM you also need a 'diffing' algorithm which compares the virtual DOM to the real DOM and selectively updates the latter. If you want to learn more about how React does this, check this article:

https://medium.com/@gethylgeorge/how-virtual-dom-and-diffing-works-in-react-6fc805f9f84e

homework/index.js
```diff
@@ -30,17 +34,212 @@
...
+        document.querySelector('tbody').innerHTML = '';
+        document.querySelector('.contributors__list').innerHTML = '';
+        fillRepositoryDetailsTable(selectedRepository);
+        fillContributorsList(selectedRepository);
+      });
+    })();
+
+    // Shows content randomly when the page is initialized
+    (function initializePage() {
+      const initialRepositoryIndex = Math.floor(Math.random() * repositories.length);
```
remarcmij: Randomizing the initially displayed repo might seem a nice gimmick but it goes against expectation and is therefore confusing. It also does not meet the requirement from the assignment.

homework/index.js
remarcmij: I'm happy to accept your solution as-is.

homework/index.js
```diff
@@ -30,17 +34,212 @@
...
+            ? selectedRepository[item[1]]
+            : `No ${item[0].toLowerCase()}`,
+        });
+      }
+    });
+  }
+
+  function fillContributorsList(selectedRepository) {
+    const list = document.querySelector('.contributors__list');
+    fetchJSON(selectedRepository.contributors_url).then(contributors => {
```
remarcmij: You should also catch errors here.

7alip:
Thank you for reviewing my code and helping me to see my shortcomings.

I made some changes on my code after your feedback.

homework/index.js
7alip: I have reviewed the post about virtual dom. I understood the reason of virtual dom and diffing algorithm. I have also reviewed some posts about 'how to type your own virtual dom?'. But it's not requierement of the assignment and hard to make it for now.

Thanks for detailed informations.

PR #265, opened 2019-06-20
Hamza week2

lizozo:
Hi Jim!! Changes left to do from last week:
1. redundant 'id's are removed
2.some of the variable names are altered in order to avoid any confusion.(repoNames => repositories, selectedOption => selectedRepo)
3. Contributors division is added (I got the layout and styling from HYF page for this div). Unlike the repositories div from week1, I created contributors div in a function and called it in my main function.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -41,6 +41,7 @@
     fetchJSONPromise(url)
       .then(data => {
         const rightDiv = createAndAppend('div', container, { class: 'rightDiv', id: 'rightDiv' });
+
         createAndAppend('p', rightDiv, {
           text: 'Contributors',
           class: 'rightDivHeader',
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -29,16 +33,126 @@
...
+          class: 'rightDivHeader',
+        });
+        const ul = createAndAppend('ul', rightDiv, { class: 'contUl' });
+        data.forEach(cont => {
+          const li = createAndAppend('li', ul, { class: 'contLi' });
+          const contDiv = createAndAppend('div', li, { class: 'contDiv' });
+          createAndAppend('img', contDiv, {
+            src: cont.avatar_url,
+            class: 'contImg',
+          });
```
remarcmij: The HTML validator complains about a missing `alt` attribute on the `img` tag:

```
Error: An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.
```

homework/index.js
```diff
@@ -29,16 +33,126 @@
...
+          createAndAppend('div', contDiv, {
+            src: cont.html_url,
```
remarcmij: A `div` does not support an `src` attribute. This is what the HTML validator says:

```
Error: Attribute “src” not allowed on element “div” at this point.
```

homework/style.css
```diff
@@ -1,3 +1,123 @@
 .alert-error {
```
remarcmij: Your app is responsive, but the select element seems to run off the screen:

image

homework/index.js
```diff
@@ -29,16 +33,126 @@
...
+          if (!descP.innerText) {
+            document.getElementById('descDiv').style.display = 'none';
+          } else document.getElementById('descDiv').style.display = 'flex';
+
+          addContributorsDiv(selectedRepo.contributors_url, container);
+        };
+      })
+      .catch(err => {
+        const root = document.getElementById('root');
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
```
remarcmij: This is the absolute minimal implementation for rendering the error. It would be nice to show the top toolbar as shown in the assignment example.

image

PR #266, opened 2019-06-20
abdullah JS3 week2

heartwhite:
week1's index.js converted to promise version.
and made some changes based on Jim's last comments on week1

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -37,19 +37,11 @@
     const header = createAndAppend('header', root, { class: 'header' });
     createAndAppend('h3', header, { text: 'HYF Repositories' });
     const selectMenu = createAndAppend('select', header, { class: 'repo-selector' });
-    const optionsArr = [];
-    for (let i = 0; i < repos.length; i++) {
-      const option = {
-        name: repos[i].name,
-        index: i,
-      };
-      optionsArr.push(option);
-    }
-    optionsArr.map(element =>
+
+    repos.map((repo, index) =>
       createAndAppend('option', selectMenu, {
-        value: element.index,
-        text: element.name,
-        id: element.name,
+        value: index,
+        text: repo.name,
       }),
     );
     return header;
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -30,15 +32,116 @@
...
+        window.open(contributor.html_url, '_blank');
+      };
+    });
+  }
+
+  function makeHtmlOfContents(repos, index) {
+    document.querySelector('.leftDiv').innerHTML = '';
+    document.querySelector('.rightDiv').innerHTML = '';
+    makeTable(repos[index]);
+    fetchJSON(repos[index].contributors_url).then(data => makeContributorsList(data));
```
remarcmij: You also need to catch and render errors here.

homework/index.js
```diff
@@ -30,15 +32,116 @@
...
+        class: 'contributor-badge',
+        text: contributor.contributions,
+      });
+      li.onclick = () => {
+        window.open(contributor.html_url, '_blank');
+      };
+    });
+  }
+
+  function makeHtmlOfContents(repos, index) {
```
remarcmij: This function has no use for the complete array of repositories. It deals with just one repository (the selected one). In general, it is best to just give functions the exact data they need and nothing more (on a 'need to know' basis). This helps to make your code more robust as there is less chance of unintended data corruption. You could define this function like this:

```js
function  makeHtmlOfContents(selectedRepo) {
  ...
}
```

and call it like this:

```js
makeHtmlOfContents(repos[selectMenu.selectedIndex]);
```

homework/index.js
```diff
@@ -30,15 +32,116 @@
     return elem;
   }
 
+  function makeHeader(repos) {
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root, { class: 'header' });
+    createAndAppend('h3', header, { text: 'HYF Repositories' });
+    const selectMenu = createAndAppend('select', header, { class: 'repo-selector' });
+    const optionsArr = [];
```
remarcmij: You can get rid of the complete `optionArr` argument: use the `repos` array directly:

```js
function makeHeader(repos) {
  createAndAppend('h3', header, { text: 'HYF Repositories' });
  const selectMenu = createAndAppend('select', header, { class: 'repo-selector' });
  repos.map((repo, index) =>
    createAndAppend('option', selectMenu, {
      value: index,
      text: repo.name,
    }),
  );
  return header;
}
```

homework/style.css
```diff
@@ -1,3 +1,108 @@
 .alert-error {
```
remarcmij: While your app is responsive, the select element needs some margin on the right:

image

homework/index.js
heartwhite: I added .catch

homework/index.js
heartwhite: yes I applied and .. I don't know why I wrote that much code before..

homework/style.css
heartwhite: I could figure out with 'min-width'

homework/index.js
heartwhite: I solved it but I forgot comment here

PR #267, opened 2019-06-20
Veli JS3-wk2

veliataseven:


homework/style.css
```diff
@@ -1,3 +1,163 @@
...
+
+.contributor-badge {
+  font-size: 12px;
+  padding: 2px 8px;
+  line-height: 1rem;
+  background-color: gray;
+  color: white;
+  border-radius: 4px;
+}
+
```
rohni: One thing I have noticed is that if you select `AngularJS` option, and then set the mobile/tablet emulator on, select `iphone 5/SE`, the Description field does not `wrap` but `crops` the value, so you can't see the full description.

This should be fixed.

homework/index.js
```diff
@@ -30,18 +32,129 @@
...
+      fork.textContent = selectedRepository.forks;
+    }
+    if (selectedRepository.updated_at !== null) {
+      const tr4 = createAndAppend('tr', tbody);
+      createAndAppend('td', tr4, { class: 'label', text: 'Updated :' });
+      const updated = createAndAppend('td', tr4, { text: '' });
+      updated.textContent = new Date(selectedRepository.updated_at).toLocaleString();
+    }
+  }
+
```
rohni: Nicely done. Clean way to clean up the children. :-)

homework/index.js
```diff
@@ -30,18 +32,129 @@
...
+      const tr3 = createAndAppend('tr', tbody);
+      createAndAppend('td', tr3, { class: 'label', text: 'Forks :' });
+      const fork = createAndAppend('td', tr3, { text: '' });
+      fork.textContent = selectedRepository.forks;
+    }
+    if (selectedRepository.updated_at !== null) {
+      const tr4 = createAndAppend('tr', tbody);
+      createAndAppend('td', tr4, { class: 'label', text: 'Updated :' });
+      const updated = createAndAppend('td', tr4, { text: '' });
+      updated.textContent = new Date(selectedRepository.updated_at).toLocaleString();
```
rohni: Why not just put the `new Date(...)` statement directly in the `options` object you pass to `createAndAppend`?

homework/index.js
```diff
@@ -30,18 +32,129 @@
...
+    if (selectedRepository.name !== null) {
+      const tr1 = createAndAppend('tr', tbody);
+      createAndAppend('td', tr1, { class: 'label', text: 'Repository :' });
+      const td2 = createAndAppend('td', tr1);
+      createAndAppend('a', td2, {
+        href: selectedRepository.html_url,
+        target: '_blank',
+        text: selectedRepository.name,
+      });
+    }
```
rohni: All the code from here on down essentially, could be refactored into a function that generates a role.  Each of the label and value are just strings.
If you want to add a detector that the `value` parameter is an object and not `null` you can use the function to generate the first row as well. 😄

homework/index.js
```diff
@@ -30,18 +32,129 @@
...
+      createAndAppend('div', contInfoDiv, {
+        class: 'contributor-badge',
+        text: element.contributions,
+      });
+      li.addEventListener('click', () => {
+        window.open(element.html_url, '_blank');
+      });
+    });
+  }
+
```
rohni: Nice job refactoring the event handler into a separate function.

homework/index.js
```diff
@@ -30,18 +32,129 @@
...
+  function listenSelectElement(repositories, root) {
+    const container = document.getElementById('container');
+    const select = document.getElementById('select');
+    select.addEventListener('change', () => {
+      clearContainer(container);
+      const newSelectedRepository = repositories[select.value];
+      createContainer(newSelectedRepository);
```
rohni: This code to the end of the function is *exactly* the same as that used in create Header.
I would have refactored this code into it's own function e.g. `setContent` or something.  And just called that from those two places.
It is not essential, but will make maintaining the code easier.

veliataseven:
> This is nicely put together.
> Only the responsive problem below needs fixing, otherwise my comments are more about stylistic choices in the code.
> Everything works, and it looks good.
> Well Done! 🥇

Hello @rohni Thanks a lot for feedback. I have done your suggestions.. new changes => https://github.com/HackYourFuture/JavaScript3/pull/267/commits/691b956295a6daa7330bf18753efeb7514642d6d

homework/index.js
rohni: I should have given an example here, sorry:
```javascript
 const newRow = (parentTable, label, content) => {
    const row = createAndAppend('tr', parentTable);
    createAndAppend('td', row, { text: label, class: 'label' });
    // insert content
    if (typeof content !== 'object' || content === null) {
      createAndAppend('td', row, { text: `${!content ? '' : content}` });
    } else {
      const cell = createAndAppend('td', row);
      cell.appendChild(content);
    }
    return row;
  };
```
You can see here that we check if the content is not an `object` or is `null` (because `null` is considered an object in javascript) and then we know it is something that can be put in a string template, so we shove it in. 😄   Otherwise we are assuming it is a collection of HTML objects that just need to be appended to the parent.

rohni:
Thanks for implementing the changes. 👍

homework/index.js
veliataseven: Sorry @rohni  I had got wrong you. Should I delete function that `detectRepositoryAndValue()` that checking if  the repository and repository's property and value are exist?

PR #268, opened 2019-06-20
zehra JS3 Week2

zehrakopar:


homework/index.js
```diff
@@ -30,15 +31,111 @@
...
+    });
+    const descriptionRow = createAndAppend('tr', tBody, { class: 'row' });
+    if (repo.description !== null) {
+      createAndAppend('td', descriptionRow, {
+        text: 'Description :',
+        class: 'label',
+      });
+      createAndAppend('td', descriptionRow, { text: repo.description });
+    }
+
```
rohni: As a style choice you could have put created a function to create a row.  This would help with not having to repeat the same code for each row.  So the code would look something like this
e.g. 
```
const descriptionRow = appendRow(tBody, 'Description', repo.description);
const forksRow = appendRow(tBody, 'Forks', repo.forks);
const dateRow = appendRow(tBody, 'Updated', new Date(repo.updated_at).toLocaleString())
```
You might have to do some magic in `appendRow` to allow the appending of HTML elements, but the first element which is an `<a/>` tag could be done manually as you do in your code above.
This would still make things a lot shorter and more readable, and if something needed adjusting it could be done in one place.

homework/style.css
```diff
@@ -1,3 +1,128 @@
...
+}
+
+.contributor-badge {
+  background-color: #808080;
+  color: white;
+  font-size: 12px;
+  padding: 3px 5px;
+  border-radius: 4px;
+}
+
```
rohni: Nice work.  I have tested the media settings and they work perfectly.

homework/index.js
zehrakopar: you are right it is shorter and readable. I will change it.

homework/style.css
zehrakopar: thank you..

PR #271, opened 2019-06-24
Ramazan Ince JS3 HW2

RamazanInce:


Week2/HYF-JS3-HW2/index.js
```diff
@@ -0,0 +1,124 @@
...
+        createAndAppend('a', cellContent, {
+          href: repo.html_url,
+          text: repo.name,
+          target: '_blank',
+        });
+      } else {
+        cellContent.textContent = repo[content[i].attribute];
+      }
+    }
+    fetchJSON(repo.contributors_url).then(contributors => renderContributors(contributors));
```
remarcmij: You must also catch the errors here.

Week2/HYF-JS3-HW2/index.js
```diff
@@ -0,0 +1,124 @@
...
+  const table = createAndAppend('table', leftContainer);
+  const listContributor = createAndAppend('tbody', table);
+
+  const rightContainer = createAndAppend('div', mainContainer, { class: 'right-block frame' });
+  createAndAppend('p', rightContainer, {
+    text: 'Contributions',
+    class: 'contributor',
+  });
+
+  const list = createAndAppend('ul', rightContainer, { class: 'list1' });
```
remarcmij: The code from lines 35-39 and 46-57 should be inside a function, not directly at the global level. If other functions need to access variables that are global in the current code, you should pass them as arguments to those functions.

Week2/HYF-JS3-HW2/index.js
```diff
@@ -0,0 +1,124 @@
...
+    });
+    return elem;
+  }
+
+  const root = document.getElementById('root');
+  const header = createAndAppend('header', root, { class: 'header' });
+  createAndAppend('p', header, { text: 'HYF Repositories' });
+
+  const select = createAndAppend('select', header, { id: 'selector' });
+  const option = x => {
```
remarcmij: This is a function but I had to look hard to find it:

1. A function should always have a verb as its main word, to indicate the action that is performed by the function. In this case you are creating `<option>` elements, so this function might be named `createOptions`.
2. The parameter name `x` is meaningless. I had to look elsewhere in your code to find out that you are passing an array of repository objects to this function, therefore the parameter name should be `repositories`. (Reserve the name `x` for the indepedent in math.)

Week2/HYF-JS3-HW2/index.js
```diff
@@ -0,0 +1,124 @@
...
+  function errorHandler(err) {
+    createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+  }
+
+  const HYF_REPS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+
+  fetchJSON(HYF_REPS_URL)
+    .then(data => test(data))
+    .catch(error => errorHandler(error));
+}
```
remarcmij: What happened to `function main()` and `window.onload` which was in the original code?

Week2/HYF-JS3-HW2/index.js
```diff
@@ -0,0 +1,124 @@
...
+          target: '_blank',
+        });
+      } else {
+        cellContent.textContent = repo[content[i].attribute];
+      }
+    }
+    fetchJSON(repo.contributors_url).then(contributors => renderContributors(contributors));
+  }
+
+  function test(data) {
```
remarcmij: What do you mean with `test`?

Week2/HYF-JS3-HW2/index.js
```diff
@@ -0,0 +1,124 @@
...
+    let optionArr = data;
```
remarcmij: Isn't the data an array of `repositories` rather than `optionArr`?

Week2/HYF-JS3-HW2/index.js
```diff
@@ -0,0 +1,124 @@
...
+
+  const rightContainer = createAndAppend('div', mainContainer, { class: 'right-block frame' });
+  createAndAppend('p', rightContainer, {
+    text: 'Contributions',
+    class: 'contributor',
+  });
+
+  const list = createAndAppend('ul', rightContainer, { class: 'list1' });
+
+  function renderContributors(selRep) {
```
remarcmij: You call this function with an array of contributors. The parameter `selRep` wrongly suggests that you are passing the selected repository.

Week2/HYF-JS3-HW2/index.js
RamazanInce: Actually they were inside the function but I remove outside to able to use .then for promises. :)

Week2/HYF-JS3-HW2/index.js
RamazanInce: Main problem is I did not understand the JS3 Week 2 and 3. And more than two months pass on it. First I should remember what I did in the JS3 HW1 and it was not easy. As you mention my coding is not easy to understand what is going on. I study promises today by myself after that try to do homework. :) . I am planing to find a way to study promises deeply.

Week2/HYF-JS3-HW2/index.js
RamazanInce: I will correct it.

Week2/HYF-JS3-HW2/index.js
RamazanInce: I try to used the .then and .catch in the function main() but it did not work then I find it in that way.
Let me try it again...

Week2/HYF-JS3-HW2/index.js
RamazanInce: just name it any how to write .then....

RamazanInce:
I will do the necessary updates and PR it again. 
Thanks for your efforts.

Week2/HYF-JS3-HW2/index.js
RamazanInce: Hi @remarcmij I did not understand this part  

"You also removed some code from the starter file for reasons unclear to me, e.g. function main(), window.onload."

Please can you put more light on it.
Thanks

Week2/HYF-JS3-HW2/index.js
remarcmij: >... I did not understand this part
>"You also removed some code from the starter file for reasons unclear to me, e.g. function main(), window.onload."
> Please can you put more light on it.

Look at the original code:

https://github.com/HackYourFuture/JavaScript3/blob/9bd7d0d563cb847ee7be566d66d05f7f7b42ee3d/homework/index.js#L33

homework/index.js
```diff
@@ -93,18 +116,21 @@
         cellContent.textContent = repo[content[i].attribute];
       }
     }
-    fetchJSON(repo.contributors_url, (error, contributors) => renderContributors(contributors));
+
+    fetchJSON(repo.contributors_url)
+      .then(contributors => renderContributors(contributors))
+      .catch(error => {
+        console.error(error);
```
remarcmij: You should render errors to the page, not to the console. Regular users do not run your app with the Chrome console open, so they will not see this error.

homework/index.js
```diff
@@ -113,9 +139,12 @@
         });
 
         renderRep(listContributor, optionArr[0]);
-      }
-    });
+      })
+      .catch(error => {
+        createAndAppend({ div: root }, { text: err.message, class: 'alert-error' });
```
remarcmij: `err` is not defined.

Did you test error handling?

homework/index.js
```diff
@@ -93,18 +116,21 @@
...
   }
 
   function main(url) {
-    fetchJSON(url, (err, data) => {
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
+    fetchJSON(url)
+      .then(data => {
         let optionArr = data;
```
remarcmij: `optionsArr`? I think you get an array of repository objects, so a name such as `repositories` would make more sense to me. And I would use that name direct;y as the parameter name of the `.then()` method, so you do not have to use line 130:

```js
.then( repositories => {
```

homework/index.js
```diff
@@ -32,39 +35,59 @@
...
+  }
 
   const mainContainer = createAndAppend('div', root, { id: 'container' });
-  const leftContainer = createAndAppend('div', mainContainer, { class: 'left-block frame' });
-  const table = createAndAppend('table', leftContainer);
-  const listContributor = createAndAppend('tbody', table);
+  function createListContributor() {
+    const leftContainer = createAndAppend('div', mainContainer, { class: 'left-block frame' });
+    const table = createAndAppend('table', leftContainer);
+    const listContributor = createAndAppend('tbody', table);
```
remarcmij: It looks like you are returning a reference to a `tbody` element; the name `listContributor` suggest that.

homework/index.js
```diff
@@ -32,39 +35,59 @@
...
 
-  const rightContainer = createAndAppend('div', mainContainer, { class: 'right-block frame' });
-  createAndAppend('p', rightContainer, {
-    text: 'Contributions',
-    class: 'contributor',
-  });
+    return listContributor;
+  }
+  const listContributor = createListContributor();
```
remarcmij: This statement should not be at the file level but inside a function body, perhaps inside `main()`.

homework/index.js
```diff
@@ -32,39 +35,59 @@
...
 
-  const list = createAndAppend('ul', rightContainer, { class: 'list1' });
+  function createMainContainer() {
```
remarcmij: This function returns a reference to a `ul` that is to contain a list of contributors. So a name such as `createContributorContainer` would be a more informative name.

homework/index.js
```diff
@@ -32,39 +35,59 @@
...
+    const rightContainer = createAndAppend('div', mainContainer, { class: 'right-block frame' });
+    createAndAppend('p', rightContainer, {
+      text: 'Contributions',
+      class: 'contributor',
+    });
+
+    return createAndAppend('ul', rightContainer, { class: 'list1' });
+  }
+  const list = createMainContainer();
```
remarcmij: 1. This statement should not be at the file level but inside a function body, e.g. `main()`.
2. The name `list` is too generic: list of what? Perhaps `contributorList`?

homework/index.js
```diff
@@ -32,39 +35,59 @@
 
   const root = document.getElementById('root');
   const header = createAndAppend('header', root, { class: 'header' });
```
remarcmij: Lines 36-37 should not be at the file level, but inside function `main()` so that they get executed along with `window.onload`.

homework/index.js
RamazanInce: i IMPLEMENT IT

homework/index.js
RamazanInce: It is done

homework/index.js
RamazanInce: the table will contain the list of contributors.

homework/index.js
RamazanInce: It did not work I tried.

homework/index.js
RamazanInce: It sounds good, I did it

homework/index.js
RamazanInce: 1 is not ok.
2. I did it.

homework/index.js
RamazanInce: I try but the code gave an error.

RamazanInce:
Hi @remarcmij, I am happy to get an assistance from a wonderful man like you. I try the implement the comments and I am going to PR it again.

Thanks

homework/index.js
```diff
@@ -30,18 +33,169 @@
...
+      .then(contributors => renderContributors(contributors))
+      .catch(err => {
+        createAndAppend('div', root, {
+          text: err.message,
+          class: 'alert-error',
+        });
+      });
+  }
+
+  function main(url) {
```
remarcmij: Here is how we can place the statements from the file level into the `main()` function body:

image

Or course, we some red underlines about unused variables. We also have some problems to fix as shown in the **PROBLEMS** tab:

image

The first one is easy: `root1 is not defined in line 37. To fix that, we can `root` through the argument list:

image

And in `main()`:

```js
const header = createHeader(root);
```

Following the same process for the other variables, passing them through the parameter list of the respective functions that need those values, your `main()` function becomes:

```js
function main(url) {
  const header = createHeader(root);
  const mainContainer = createMainContainer(root);
  const listContributor = createListContributor(mainContainer);
  const contributorList = createContributorContainer(mainContainer);

  fetchJSON(url)
    .then(data => {
      let repositories = data;
      repositories = repositories.sort((a, b) => a.name.localeCompare(b.name));

      createOptions(repositories, header);

      const selectedValue = document.getElementById('selector');
      selectedValue.addEventListener('change', event => {
        const selectedRepo = event.target.value;
        renderRep(listContributor, repositories[selectedRepo], root, contributorList);
      });

      renderRep(listContributor, repositories[0], root, contributorList);
    })
    .catch(err => {
        text: err.message,
        class: 'alert-error',
      });
    });
}
```

And the corresponding functions parameter lists become:

```js
function createHeader(root) {
  ...
}

function createOptions(repositoryList, header) {
  ...
}

function createMainContainer(root) {
  ...
}

function createListContributor(mainContainer) {
  ...
}

function createListContributor(mainContainer) {
  ...
}

function createContributorContainer(mainContainer) {
   ...
}

function renderRep(listContributors, repo, root, contributorList) {
   ...
}
```

It may seem that we have done more work as compared to your version with global variables, but now we can see exactly what each function needs without having to examine their function body and check for access to global variables. This makes the code easier to understand.

homework/index.js
```diff
@@ -30,18 +33,169 @@
...
+
+  function createMainContainer() {
+    const mainContainer = createAndAppend('div', root, {
+      id: 'container',
+    });
+    return mainContainer;
+  }
+  const mainContainer = createMainContainer();
+
+  function createListContributor() {
```
remarcmij: This function does not deal with the contributor list but with the table containing repository details. The name `createListContributor` does not match what this function is doing.

RamazanInce: I chance it as containerListContributor.

remarcmij: No, I'm afraid you are not getting the point. Contributors play no role at all in this function. Therefore the 'contributors' should not be part of the function name.

RamazanInce: containerListFotContributor may be ok but it will be to long
just tbodyCondainer can be also?
About naming the things I am not good enough. I need to work more...

homework/index.js
RamazanInce: Hi @remarcmij I pushed it again I hope this time around it will be ok.

homework/index.js
remarcmij: You are still not getting my point. Again, this function has nothing to do with contributors. So the word `contributors` should not be part of the name or even be considered to be part of the name. A good name could, for instance, could be `createRepoTableContainer` or something like that. The table is used to display details about the repository and not about contributors.

RamazanInce:
Hi @remarcmij thanks for your efforts.
And also great thanks to the @wouterkleijn for his follow ups.

PR #272, opened 2019-06-24
7alip | Week3

7alip:


homework/index.js
```diff
@@ -1,47 +1,238 @@
...
+        const repositories = await Helper.fetchJSON(url);
+        this.init(repositories, root);
+      } catch (err) {
+        Helper.createAndAppend('div', root, { class: 'alert-error', text: err.message });
       }
-    });
+    }
   }
 
   const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
```
remarcmij: Your app doesn't handle HTTP errors correctly. For instance, try and change `orgs` in `orgsx` in this url to force an HTTP 404 error and reload the browser.

homework/index.js
```diff
@@ -1,47 +1,238 @@
...
+          text: contributor.contributions,
+        });
+      });
+    } catch (error) {
+      Helper.createAndAppend('li', list, { text: error.message, class: 'alert-error' });
+    }
+  }
+
+  class App {
+    constructor(select) {
```
remarcmij: This constructor is never executed because you never create an instance of App (with `new App(select)`). Instead, you use the  `App` class as a `static` class.

homework/index.js
```diff
@@ -1,47 +1,238 @@
...
+      this.select = select;
```
remarcmij: If you use this in a static method, the this value refers to the class itself rather than an instance of the class. Therefore, `this` and `App` are equivalent. You could change all references to `this` to `App` and the code would still work:

```js
class App {
  static init(data, root) {
    createStaticComponentsFromModel(componentModel, root);

    const repositories = data.sort((a, b) =>
      a.name.toLowerCase().localeCompare(b.name.toLowerCase()),
    );

    App.select = document.querySelector('.header__select');

      Helper.createAndAppend('option', App.select, { text: repository.name });
    });

    App.select.addEventListener('change', () => {
      document.querySelector('tbody').innerHTML = '';
      document.querySelector('.contributors__list').innerHTML = '';
      fillRepositoryDetailsTable(repositories[App.select.selectedIndex]);
      fillContributorsList(repositories[App.select.selectedIndex]);
    });

    fillRepositoryDetailsTable(repositories[App.select.selectedIndex]);
    fillContributorsList(repositories[App.select.selectedIndex]);
  }

  static async main(url) {

    try {
      const repositories = await Helper.fetchJSON(url);
      App.init(repositories, root);
    } catch (err) {
      Helper.createAndAppend('div', root, { class: 'alert-error', text: err.message });
    }
  }
}
```

Of course, using a static class like this, while your other code still uses traditional functions seems inconsistent. I know that your second PR takes a different approach, so no need to fix this here.

PR #273, opened 2019-06-25
Veli JavaScript3-Week3

veliataseven:


homework/Repository.js
```diff
@@ -8,13 +8,54 @@ class Repository {
...
-    // TODO: replace the next line with your code.
-    Util.createAndAppend('pre', container, JSON.stringify(this.repository, null, 2));
+    const leftDiv = Util.createAndAppend('div', container, {
+      id: 'leftDiv',
+      class: 'left-div white_Frame',
+    });
+    const table = Util.createAndAppend('table', leftDiv);
+    const tbody = Util.createAndAppend('tbody', table, { id: 'tbody' });
+
+    if (this.detectRepositoryAndValue(this.repository, this.repository.name)) {
```
remarcmij: See my comment for `index.js`.

veliataseven:
Thank you for feedback Jim, I have done your comments.. Here it is https://github.com/HackYourFuture/JavaScript3/commit/873ed84aaaabdf27db2b31c2e3e63080cb17e07f

homework/Repository.js
```diff
@@ -8,8 +8,8 @@ class Repository {
     this.repository = repository;
   }
 
-  detectRepositoryAndValue(repository, value) {
-    if (typeof repository !== 'undefined' && value !== null) {
+  detectValueOfRepository(value) {
```
remarcmij: If you don't mind my saying so, this is a completely useless method. If the only thing you want to do is check whether a value is null or not, surely you don't need a function for that.

veliataseven:
> Thank you for feedback Jim, I have done your comments.. Here it is [873ed84](https://github.com/HackYourFuture/JavaScript3/commit/873ed84aaaabdf27db2b31c2e3e63080cb17e07f)

I have changed some codes on class version of homework..please notice that.

homework/Repository.js
```diff
@@ -27,7 +27,7 @@ class Repository {
     const table = Util.createAndAppend('table', leftDiv);
     const tbody = Util.createAndAppend('tbody', table, { id: 'tbody' });
 
-    if (this.detectRepositoryAndValue(this.repository, this.repository.name)) {
+    if (this.detectValueOfRepository(this.repository.name)) {
```
remarcmij:
```js
if (this.repository.name !== null) {
```

See: no method needed.

homework/Repository.js
veliataseven: I have changed already that method in last changes , After last changes.. that method is checking that İf value of repository is equal to null or not

remarcmij: I still do not think this function adds value. Firstly, its name pretends to do more than it actually does. If you look at the function body, you can find that it only tests whether the value passed to it as a parameter is `null` or not and returns a boolean value to indicate the result. Based on what is actually does this function is better called `isNotNull(value)`. It is not at all exclusive to working with repositories as its name suggests.

Secondly, this function does trivial work. You could rewrite as:

```js
function isNotNull(value) {
  return value != null;
}
```

This is not worth to write a function for. Just use the test for `null` directly, e.g.:

```js
if (selectedRepository.name != null) {
```

veliataseven:
Okay @remarcmij , Got it. Thanks a lot

veliataseven:
Dear @remarcmij , I've applied your comments. new improvements of JS3-week3 => https://github.com/HackYourFuture/JavaScript3/commit/aed4322623bd2bd968a70da764bf1abb2ff0f101
Thank you

PR #274, opened 2019-06-26
Suheyb - Week3

suh3yb:


homework/App.js
```diff
@@ -12,18 +14,41 @@ class App {
...
+      if (reposResponse.status !== 200) {
+        throw new Error(`Oops, error: ${reposResponse.status}, ${reposResponse.statusText}`);
+      }
       this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      this.repos
+        .sort((a, b) =>
+          a
+            .name()
+            .toLowerCase()
```
remarcmij: I can't remember if I commented on this in earlier homework, but when using `localeCompare` you should not convert to lower case first. Some languages do not even have different cases (e.g. Arabic). All language-specific aspects for comparing two strings are completely handled by `localeCompare`, which uses the language set up by the user in the computer's operating system for its rules.

suh3yb:
Hi @remarcmij, thank you very much for your feedback. I deleted `toLowerCase()` and noted your point about `localeCompare`.
Best regards.

PR #275, opened 2019-06-26
Abdullah JS3 week3

heartwhite:
JS3 week3 index.js the fetct function and async-await added

### DIFF
--- a/homework/App.js
+++ b/homework/App.js
@@ -12,18 +12,35 @@ class App {
    * @param {string} url The GitHub URL for obtaining the organization's repositories.
    */
   async initialize(url) {
-    // Add code here to initialize your app
-    // 1. Create the fixed HTML elements of your page
-    // 2. Make an initial XMLHttpRequest using Util.fetchJSON() to populate your <select> element
-
     const root = document.getElementById('root');
-
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
+    const header = Util.createAndAppend('header', root, { class: 'header' });
+    Util.createAndAppend('h3', header, { text: 'HYF Repositories' });
+    const selectMenu = Util.createAndAppend('select', header, {
+      class: 'repo-selector',
+      tabindex: 1,
+    });
+    Util.createAndAppend('div', root, { id: 'container' });
 
     try {
       const repos = await Util.fetchJSON(url);
       this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      this.repos
+        .sort((one, two) =>
+          one
+            .name()
+            .toLowerCase()
+            .localeCompare(two.name().toLowerCase()),
+        )
+        .map((repo, index) =>
+          Util.createAndAppend('option', selectMenu, {
+            value: index,
+            text: repo.name(),
+          }),
+        );
+      this.fetchContributorsAndRender(selectMenu.selectedIndex);
+      selectMenu.onchange = () => {
+        this.fetchContributorsAndRender(selectMenu.selectedIndex);
+      };
     } catch (error) {
       this.renderError(error);
     }
@@ -52,10 +69,10 @@ class App {
       const container = document.getElementById('container');
       App.clearContainer(container);
 
-      const leftDiv = Util.createAndAppend('div', container);
-      const rightDiv = Util.createAndAppend('div', container);
+      const leftDiv = Util.createAndAppend('div', container, { class: 'leftDiv whiteframe' });
+      const rightDiv = Util.createAndAppend('div', container, { class: 'rightDiv whiteframe' });
 
-      const contributorList = Util.createAndAppend('ul', rightDiv);
+      const contributorList = Util.createAndAppend('ul', rightDiv, { class: 'rightDiv' });
 
       repo.render(leftDiv);
 
@@ -72,7 +89,8 @@ class App {
    * @param {Error} error An Error object describing the error.
    */
   renderError(error) {
-    console.log(error); // TODO: replace with your own code
+    const root = document.getElementById('root');
+    Util.createAndAppend('div', root, { text: error.message, class: 'alert-error' });
   }
 }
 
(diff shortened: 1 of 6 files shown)


homework/App.js
```diff
@@ -12,18 +12,35 @@ class App {
...
-    // 2. Make an initial XMLHttpRequest using Util.fetchJSON() to populate your <select> element
-
     const root = document.getElementById('root');
-
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
+    const header = Util.createAndAppend('header', root, { class: 'header' });
+    Util.createAndAppend('h3', header, { text: 'HYF Repositories' });
+    const selectMenu = Util.createAndAppend('select', header, {
+      class: 'repo-selector',
+      tabindex: 1,
```
remarcmij: If you want to set the focus on the `<select>` element when the page loads,  setting the `tabindex` doesn't work. Use `autofocus` instead:

```js
 class: 'repo-selector',
 autofocus: true,
```

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/select#Attributes

homework/style.css
```diff
@@ -1,3 +1,112 @@
...
+  padding: 0;
+  margin: 0;
+}
+.contributor-item {
+  border-bottom: solid 1px rgb(0, 0, 0, 12%);
+  padding: 16px;
+  display: flex;
+  flex-direction: row;
+  align-items: center;
+  cursor: pointer;
```
remarcmij: To get rid of the underlines in the list items you can add:

```css
text-decoration: none;
```

And you can remove `cursor: pointer` as that is the default for an anchor tag.

homework/index.js
```diff
@@ -30,17 +15,118 @@
...
+      const root = document.getElementById('root');
+      createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+    }
+  }
+
+  function mainHtmlConstructor(repos) {
+    makeHeader(repos);
+    const root = document.getElementById('root');
+    const container = createAndAppend('div', root, { id: 'container' });
+    createAndAppend('div', container, { class: 'leftDiv whiteframe' });
```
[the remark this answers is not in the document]
heartwhite: my class names are converted to kebab case.

homework/index.js
```diff
@@ -30,17 +15,118 @@
...
+    const selectMenu = document.querySelector('select');
+    makeHtmlOfContents(repos[selectMenu.selectedIndex]);
+    selectMenu.onchange = () => {
+      makeHtmlOfContents(repos[selectMenu.selectedIndex]);
+    };
+  }
+
+  async function main(url) {
+    try {
+      const responseOfHyfRepos = await fetch(url);
```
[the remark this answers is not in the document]
heartwhite: yes. I handled error by using response.ok inside try area.
I read the documentation you sent. I hope I understood that

homework/index.js
```diff
@@ -30,17 +15,118 @@
...
+      const repositories = await responseOfHyfRepos.json();
+      mainHtmlConstructor(
+        repositories.sort((one, two) =>
+          one.name.toLowerCase().localeCompare(two.name.toLowerCase()),
```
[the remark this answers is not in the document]
heartwhite: I removed lowercase thing. it works

homework/App.js
heartwhite: I removed lowercase thing. it works

homework/App.js
```diff
@@ -12,18 +12,35 @@ class App {
...
 
     try {
       const repos = await Util.fetchJSON(url);
       this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      this.repos
+        .sort((one, two) =>
+          one
+            .name()
+            .toLowerCase()
```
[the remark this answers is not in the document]
heartwhite: I removed lowercase thing. it works

homework/style.css
heartwhite: yes. I applied text-decoration, removed th cursor thing.

homework/index.js
```diff
@@ -30,17 +15,118 @@
...
+    createAndAppend('td', trForks, { text: 'Forks:' });
+    createAndAppend('td', trForks, { text: repo.forks_count });
+    const trUpdate = createAndAppend('tr', tBody);
+    createAndAppend('td', trUpdate, { text: 'Updated:' });
+    createAndAppend('td', trUpdate, {
+      text: new Date(repo.updated_at).toLocaleString(),
     });
   }
 
+  function makeContributorsList(repos) {
```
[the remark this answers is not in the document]
heartwhite: oh! what I were thinking. ...
I changed to contributors

homework/index.js
```diff
@@ -30,15 +15,115 @@
...
+    makeHtmlOfContents(repos[selectMenu.selectedIndex]);
+    selectMenu.onchange = () => {
+      makeHtmlOfContents(repos[selectMenu.selectedIndex]);
+    };
+  }
+
+  async function main(url) {
+    try {
+      const responseOfHyfRepos = await fetch(url);
+      const repositories = await responseOfHyfRepos.json();
```
remarcmij: This must be done **_after_** you have checked for errors: if there is an error there is no JSON to convert.

homework/index.js
```diff
@@ -30,15 +15,115 @@
...
+      if (!responseOfHyfRepos.ok) {
+        throw Error(responseOfHyfRepos.statusText);
```
remarcmij: You must always prefix `Error` with `new` (although JavaScript may be forgiving if you forget it):

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

homework/App.js
```diff
@@ -12,18 +12,30 @@ class App {
...
+      autofocus: true,
+    });
+    Util.createAndAppend('div', root, { id: 'container' });
 
     try {
       const repos = await Util.fetchJSON(url);
       this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      this.repos
+        .sort((one, two) => one.name().localeCompare(two.name().toLowerCase()))
```
remarcmij: There is still a `.toLowerCase()` in here, which shouldn't be there.

PR #276, opened 2019-06-26
Improvements using classes on HW/W2

ghifo:


homework/index.js
```diff
@@ -30,17 +15,123 @@
...
+    });
+    createTd('Description', repository.description, tbody);
+    createTd('Forks', repository.forks, tbody);
+    createTd('Updated', repository.updated_at, tbody);
+  }
+
+  // contributors list
+  async function listContributors(repository, container) {
+    const contributorUrl = repository.contributors_url;
+    const response = await fetch(contributorUrl);
```
remarcmij: 1. All `await` expressions should be inside the `try` block.
2. You need to check for HTTP errors. See my message in the #class21 slack channel.

homework/index.js
```diff
@@ -30,17 +15,123 @@
...
+      });
+    }
+  }
+
+  // create the select options and the event listener for changing repositories
+  function selectAndChange(repositories, container, header) {
+    // localeCompare enables a case-insensitive sort of an array.
+    repositories.sort((a, b) => a.name.localeCompare(b.name));
+
+    const select = createAndAppend('select', header, { class: 'repository-selector' });
```
remarcmij: You can add the `autofocus` attribute so that the `<select>` element gets the keyboard focus when the page is loaded:

```js
const select = createAndAppend('select', header, {
  class: 'repository-selector',
  autofocus: true,
});
```

homework/index.js
```diff
@@ -30,17 +15,123 @@
...
+      const index = select.value;
+      createDetailsTable(repositories[index], container);
+      listContributors(repositories[index], container);
     });
   }
 
+  async function main(url) {
+    const root = document.getElementById('root');
+    // const response = await fetch(contributorUrl);
+    // const contributors = await response.json();
```
remarcmij: Do not leave commented-out code in pull requests.

homework/index.js
```diff
@@ -30,17 +15,123 @@
...
+    const response = await fetch(url);
+    const repositories = await response.json();
```
remarcmij: All `await` expressions should be inside the `try` block.

homework/index.js
ghifo: Sorry, I totally forgot them.
I will solve them right now.

homework/index.js
ghifo: Oh yes, I see that.
Thanks!

homework/index.js
ghifo: To move between the select choices you mean!
When I press `tab` it starts from the `select` already.

remarcmij: With `autofocus` you don't need to press `tab` at all. You can straight away use the up and down keys to select a repo.

ghifo: Good to know, 
Thanks!

PR #277, opened 2019-06-26
Week 3

lizozo:
1.Changes made in "hamza homework week 2 update" commit:
1.1 Added 'alt' properties to the img elements.
1.2 Navigation by tab key.
1.3 Select element no more overflows
2. Commit "hamza week3 changes made to align App.js and index.js styling"
I had to change some element id's in my index.js because index2.html uses same style.css with index.html.
3. hamza week3 Classes homework

### DIFF
--- a/homework/App.js
+++ b/homework/App.js
@@ -12,18 +12,31 @@ class App {
    * @param {string} url The GitHub URL for obtaining the organization's repositories.
    */
   async initialize(url) {
-    // Add code here to initialize your app
-    // 1. Create the fixed HTML elements of your page
-    // 2. Make an initial XMLHttpRequest using Util.fetchJSON() to populate your <select> element
-
-    const root = document.getElementById('root');
-
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
-
     try {
+      const root = document.getElementById('root');
+      const header = Util.createAndAppend('header', root, { class: 'header' });
+      Util.createAndAppend('div', root, {
+        class: 'container',
+        id: 'container',
+      });
+      const headerDiv = Util.createAndAppend('div', header, { class: 'headerDiv' });
+      Util.createAndAppend('span', headerDiv, { text: 'HYF Repositories', id: 'headerSpan' });
+      const select = Util.createAndAppend('select', headerDiv, { class: 'select' });
+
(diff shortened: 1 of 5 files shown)


homework/index.js
```diff
@@ -29,16 +26,126 @@
...
+        class: 'pClass',
+      });
+      const repoA = createAndAppend('a', repoDiv, {
+        text: selectedRepo.name,
+        href: selectedRepo.html_url,
+        target: '_blank',
+      });
+
+      addContributorsDiv(selectedRepo.contributors_url, container);
+      select.onchange = function func() {
```
remarcmij: Why not use an arrow function here?

homework/App.js
```diff
@@ -12,18 +12,31 @@ class App {
...
+      Util.createAndAppend('span', headerDiv, { text: 'HYF Repositories', id: 'headerSpan' });
+      const select = Util.createAndAppend('select', headerDiv, { class: 'select' });
+
       const repos = await Util.fetchJSON(url);
+      repos.sort((a, b) => a.name.localeCompare(b.name));
       this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      this.repos.forEach(repo => Util.createAndAppend('option', select, { text: repo.name() }));
+      let index = select.selectedIndex;
+      this.fetchContributorsAndRender(index);
```
remarcmij: I would just forget about creating an extra variable `index` here, as you have no other use for this variable other than to pass it as an argument to `fetchContributorsAndRender()`:

```js
this.fetchContributorsAndRender( select.selectedIndex);
```

homework/App.js
```diff
@@ -12,18 +12,31 @@ class App {
...
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
-
     try {
+      const root = document.getElementById('root');
+      const header = Util.createAndAppend('header', root, { class: 'header' });
+      Util.createAndAppend('div', root, {
+        class: 'container',
+        id: 'container',
+      });
+      const headerDiv = Util.createAndAppend('div', header, { class: 'headerDiv' });
```
remarcmij: You could add an `autofocus` attribute here so that the select element get the keyboard focus when the page is first loaded.  A visually impaired user can then immediately change the selected repo by using up and down keys.

PR #280, opened 2019-06-27
Alusine - JavaScript3 - Week3

alquinex:


homework/index.js
```diff
@@ -30,15 +32,126 @@
...
+        class: 'contributor-badge',
+      });
+    });
+  }
+
+  function fetchAndRenderData(selectedRepo, repoContainer, contributorsContainer, root) {
+    repoContainer.innerHTML = '';
+    contributorsContainer.innerHTML = '';
+
+    const updatedAt = new Date(selectedRepo.updated_at);
```
remarcmij: This variable is unused and produces an ESLint error (red underlined in VSCode). You should fix ESLint errors before sending in a PR.

Note that the repository date as displayed on the page is not a user-friendly format. This variable was probably intended to fix that but is left unused. Please pay attention to detail.

homework/index.js
```diff
@@ -30,15 +32,126 @@
...
+      text: `${selectedRepo.name}`,
+      href: selectedRepo.html_url,
+      target: '_blank',
+      class: 'repo-child right-cell',
+    });
+
+    Object.keys(repoTitles).forEach(key => {
+      const tr = createAndAppend('tr', tbody);
+      const leftCell = createAndAppend('td', tr);
+      const rightCell = createAndAppend('td', tr);
```
remarcmij: Line 95 and 96 produce ESLint errors:

```
95:13  error  'leftCell' is already declared in the upper scope   no-shadow
96:13  error  'rightCell' is already declared in the upper scope  no-shadow
```

This is because you have already used these names in an outer scope (lines 82-83) and using these same names makes the variable from the outer scope unreachable in the inner scope.  At this particular point, you have no use for the variables in the outer scope, but it is considered a bad practice anyway. 

A simple way to solve this is to declare the variables in lines 82 with `let` instead of `const` and to reuse those same variables here without `const` or `let`.

```js
let leftCell = createAndAppend('td', firstRow);  // line 82
... 
leftCell = createAndAppend('td', tr); // line 95
```

homework/index.js
```diff
@@ -30,15 +32,126 @@
...
+      });
+
+      createAndAppend('span', rightCell, {
+        text: `${selectedRepo[key]}`,
+        class: 'repo-child right-cell',
+      });
+    });
+
+    fetchJSON(selectedRepo.contributors_url)
+      .then(contributors => {
```
remarcmij: For this homework, this should use `async/await` too.

homework/index.js
```diff
@@ -30,15 +32,126 @@
...
+    const table = createAndAppend('table', repoContainer, { class: 'table' });
+    const tbody = createAndAppend('tBody', table);
+
+    const firstRow = createAndAppend('tr', tbody);
+    const leftCell = createAndAppend('td', firstRow);
+    const rightCell = createAndAppend('td', firstRow);
+
+    createAndAppend('span', leftCell, { text: 'Repository: ', class: 'repo-child left-cell' });
+    createAndAppend('a', rightCell, {
+      text: `${selectedRepo.name}`,
```
remarcmij: There is not much point to use a template literal here. That only makes sense if you want to combine fix pieces of text with variable pieces, which is not the case here. You can just use the string directly:

```js
text: selectedRepo.name,
```

The same holds true for a couple of other places in your code.

homework/style.css
```diff
@@ -1,3 +1,165 @@
...
+  font-weight: bold;
+}
+
+.right-cell {
+  margin-left: 8px;
+}
+
+.repo-child {
+  font-size: 16px;
+}
```
remarcmij: The table formatting doesn't seem completely right. Specifically, a repo description that spans multiple lines should be left-justified, including the first line:

image

homework/App.js
```diff
@@ -18,17 +18,44 @@ class App {
 
     const root = document.getElementById('root');
 
-    Util.createAndAppend('h1', root, { text: 'It works!' }); // TODO: replace with your own code
+    Util.createAndAppend('h1', root, { text: 'It works!' });
```
remarcmij: Why is this line still here?

homework/index.js
alquinex: Hi @remarcmij, 
Thanks for the feedback and for your time. Regarding all the issues  I am still working on them. I should have stated that in the pull request. 
I could have probably received this feedback on the week1 or week2 homework if I had used this format earlier. The week3 is like an overhaul of week1 & 2 because I want to make it look better. But I will fix the issues and let you know once it's all done.

alquinex:
Hi @remarcmij, the changes have been made.

homework/Repository.js
```diff
@@ -13,8 +13,42 @@ class Repository {
...
+    const table = Util.createAndAppend('table', container, { class: 'table' });
+    const tbody = Util.createAndAppend('tBody', table);
+
+    const firstRow = Util.createAndAppend('tr', tbody);
+    let leftCell = Util.createAndAppend('td', firstRow);
+    let rightCell = Util.createAndAppend('td', firstRow);
+
+    Util.createAndAppend('span', leftCell, { text: 'Repository: ', class: 'repo-child left-cell' });
+    Util.createAndAppend('a', rightCell, {
+      text: this.repository.name,
```
remarcmij: Instead of `this.repository` you could (for consistency) use `repo` here, which is an exact clone of `this.repository`, except for the formatted date.

alquinex:
Hi @remarcmij, thanks for the feedback and your precious time.

PR #281, opened 2019-06-28
homework3

Nadoshka:


homework/App.js
```diff
@@ -12,18 +12,27 @@ class App {
...
-      // TODO: add your own code here
+      this.repos = repos
+        .sort((a, b) => a.name.localeCompare(b.name))
+        .map(repo => new Repository(repo));
+
+      repos.forEach((repository, index) => {
+        Util.createAndAppend('option', select, { text: repository.name, value: index });
+      });
+      const index = select.value;
+      this.fetchContributorsAndRender(index);
```
remarcmij: Why not call it like this:

```js
this.fetchContributorsAndRender(select.value);
```

and get rid of the `index` variable that you do not use elsewhere.

homework/Repository.js
```diff
@@ -12,9 +12,29 @@ class Repository {
    * Render the repository info to the DOM.
    * @param {HTMLElement} container The container element in which to render the repository.
    */
+  // container = ul
   render(container) {
```
remarcmij: A two-column HTML table instead of an unordered list would have produced a bit sleeker UI, but okay, this works too.

Nadoshka: Thanks, U are right! I already used the table when I was in class 20, so I thought it could be better if I try it in another way which is an unordered list :)

homework/App.js
Nadoshka: Thanks for your feedback! I just want to make it clear for me "index = select.value"

Nadoshka:
> Hi @Nadoshka , overall pretty good work, no real issues. The only thing I would like you to try is to replace `XMLHttpRequest` with `fetch`. This you can do by modifying `fetchJSON` in the `Util` class.

Thanks @remarcmij it is done. It was interesting to figure it out 👍

homework/Util.js
```diff
@@ -18,18 +18,17 @@ class Util {
 
   static fetchJSON(url) {
     return new Promise((resolve, reject) => {
```
remarcmij: You no longer need to create a promise yourself when you use `fetch()` because it already returns a promise itself. You only need to use `new Promise(...)` if you need to convert a callback-based into a promised based one. Can you try it without `new Promise(..)`?

remarcmij: Hi @Nadoshka, I would still like to ask you to try and make this change. See it as another opportunity to practice your skills and to get feedback.

Nadoshka: Thanks for the explanation! I appreciate it! 
Change is done!  Thank u in advance for your feedback!

homework/Util.js
```diff
@@ -17,18 +17,11 @@ class Util {
...
-          }
-          return response.json();
-        })
-        .then(response => resolve(response))
-        .catch(error => {
-          reject(new Error(`Network error: ${error.message}`));
-        });
+    return fetch(url).then(response => {
+      if (!response.ok) {
+        throw new Error(`Network request failed`);
```
remarcmij: You can use `response.status` and `response.statusText` to give a more informative error message.

PR #282, opened 2019-06-28
JavaScript3 Week3 Zehra

zehrakopar:


homework/index.js
```diff
@@ -30,15 +15,105 @@
...
+      }
+
+      const container = createAndAppend('div', root, { id: 'container' });
+      createAndAppend('div', container, { class: 'left-div' });
+      createAndAppend('div', container, { class: 'right-div' });
+      reposAbstract(selector.value, repositories);
+      reposContributor(repositories[selector.value].contributors_url);
+      selector.addEventListener('change', () => {
+        document.querySelector('.left-div').innerHTML = '';
+        document.querySelector('.right-div').innerHTML = '';
```
remarcmij: You can use JavaScript variables rather than looking up the elements in the DOM here:

```js
const container = createAndAppend('div', root, { id: 'container' });
const leftDiv = createAndAppend('div', container, { class: 'left-div' });
const rightDiv = createAndAppend('div', container, { class: 'right-div' });
reposAbstract(selector.value, repositories);
reposContributor(repositories[selector.value].contributors_url);
selector.addEventListener('change', () => {
  leftDiv.innerHTML = '';
  rightDiv.innerHTML = '';
```

homework/App.js
```diff
@@ -17,13 +17,27 @@ class App {
...
       const repos = await Util.fetchJSON(url);
-      this.repos = repos.map(repo => new Repository(repo));
-      // TODO: add your own code here
+      this.repos = repos
+        .sort((a, b) => a.name.localeCompare(b.name))
+        .map(repo => new Repository(repo));
+      for (let i = 0; i < repos.length; i++) {
+        Util.createAndAppend('option', selector, { text: repos[i].name, value: i });
+      }
+      this.repos[0].render(leftDiv);
```
remarcmij: The `.render()` method of the `Repository` object is also called in `fetchContributorsAndRender()`, line 73. This redundant call at initialization time should not be there.

homework/Repository.js
```diff
@@ -12,9 +12,29 @@ class Repository {
...
+    const repositoryRow = Util.createAndAppend('tr', tBody, { class: 'row' });
+    Util.createAndAppend('td', repositoryRow, { text: 'Repository :', class: 'label' });
+    const repositoryRowSecondTd = Util.createAndAppend('td', repositoryRow);
+    Util.createAndAppend('a', repositoryRowSecondTd, {
+      href: this.repository.html_url,
+      target: '_blank',
+      text: this.repository.name,
+    });
+
+    function appendRow(parent, text1, text2) {
```
remarcmij: Avoid embedded names functions (other than unnamed callbacks and event handlers). They cause the function body of the enclosing to become larger (and therefore more cumbersome to oversee). In this case, since this function does not need anything from the object instance, you could make it a static method, like this:

```js
render(leftDiv) {
  ...
   if (this.repository.description !== null) {
    Repository.appendRow(tBody, 'Description :', this.repository.description);
  }
  Repository.appendRow(tBody, 'Forks :', this.repository.forks);
  Repository.appendRow(tBody, 'Updated :', new Date(this.repository.updated_at).toLocaleString());
}

static appendRow(parent, text1, text2) {
  const tr = Util.createAndAppend('tr', parent, { class: 'row' });
  Util.createAndAppend('td', tr, { text: text1, class: 'label' });
  Util.createAndAppend('td', tr, { text: text2 });
}
```

homework/index.js
zehrakopar: ok i got it.

homework/Repository.js
zehrakopar: yes, after learning class static function, using it will be better.

zehrakopar: Thanks for your good wishes. And Thank you for your all efforts.

PR #284, opened 2019-07-08
aalhommada-js3-week1

aalhommada:


aalhommada:
Thanks Jim for your nice comment , indeed i am convienent with every step i do it in this homework,  and actually i finish it after looking for more than 10 others homework and the main one is HYF example on netlify , and help from someone, And write every step by my self even if some parts look like the others code , and it it not rehashed . But if passing this test depend on this changing i will do it .

remarcmij:
Hi @aalhommada , I'll give you a response as a DM in slack.

homework/index.js
```diff
@@ -30,18 +31,102 @@
...
+  function createContributors(selectedRepository, root) {
+    const contributorsContainer = createAndAppend('div', mainContainerForRepo, {
+      class: 'right-div',
+    });
+    createAndAppend('h3', contributorsContainer, {
+      text: 'Contributors',
+      class: 'contributors-header',
+    });
+    const ul = createAndAppend('ul', contributorsContainer, { class: 'contributor-list' });
+    fetchJSON(selectedRepository.contributors_url, (err, ShowContributors) => {
```
remarcmij: Parameter names should be camelCase: `showContributors`. Moreover, you should not use verbs in variable or parameters names, only for function names (functions perform an action). So the name `contributors` would be more fitting here.

homework/index.js
```diff
@@ -30,18 +31,102 @@
...
+      if (err) {
+        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+      } else {
+        ShowContributors.forEach(contributor => {
+          const li = createAndAppend('li', ul, {
+            class: 'contributor-item',
+          });
+          const alink = createAndAppend('a', li);
```
remarcmij: You are creating an `<a>` tag here, but it doesn't allow you to go to the contributor page, because you are not setting its `href` attribute.

homework/index.js
```diff
@@ -30,18 +31,102 @@
...
+    repositories.forEach((repo, index) => {
+      createAndAppend('option', select, { value: index, text: repo.name });
+    });
+    select.addEventListener('change', event => {
+      const selectedIndex = event.target.value;
+      const selectedRepository = repositories[selectedIndex];
+      while (mainContainerForRepo.firstChild) {
+        mainContainerForRepo.removeChild(mainContainerForRepo.firstChild);
+      }
+      createDescription(selectedRepository, root);
```
remarcmij: Your `createDescription()` function takes one parameter only.

homework/index.js
```diff
@@ -30,18 +31,102 @@
...
+  }
   function main(url) {
     fetchJSON(url, (err, data) => {
       const root = document.getElementById('root');
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        creatHeader(data, root);
+        createDescription(data[0], mainContainerForRepo);
```
remarcmij: Your `createDescription()` function takes one parameter only.

homework/index.js
```diff
@@ -30,18 +31,102 @@
...
+        createContributors(data[0], mainContainerForRepo);
```
remarcmij: Your `createContributors()`  function expects the root element as its second parameter.

homework/index.js
```diff
@@ -30,18 +31,102 @@
...
+        });
+      }
+    });
+  }
   function main(url) {
     fetchJSON(url, (err, data) => {
       const root = document.getElementById('root');
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
       } else {
```
remarcmij: It is best to exit early from the callback function if you detect an error here. This saves you from having to track down where the `else` clause should end and also saves you some indentation:

```js
if (err) {
  return;
}
creatHeader(data, root);
...
```

You can do the same thing in your `createContributors()` function.

homework/.vscode/settings.json
```diff
@@ -0,0 +1,7 @@
+{
+  "cSpell.words": [
+    "REPOS",
+    "cotributor",
```
remarcmij: You accepted this as a spelling error to be ignored. The better option was to correct the spelling error.

homework/index.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-use-before-define */
```
remarcmij: Do not disable this rule but abide by it.

homework/index.js
```diff
@@ -15,7 +17,6 @@
     xhr.onerror = () => cb(new Error('Network request failed'));
     xhr.send();
   }
-
```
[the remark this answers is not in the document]
aalhommada: I check and modify it in the whole file .

homework/index.js
```diff
@@ -30,18 +31,102 @@
     return elem;
   }
 
+  function creatHeader(repositories, root) {
```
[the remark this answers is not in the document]
aalhommada: thanks for mention this , it is done

homework/index.html
```diff
@@ -1,23 +1,25 @@
...
-<body>
-  <div id="root"></div>
-  <script src="./index.js"></script>
-</body>
-
-</html>
\ No newline at end of file
+  <body>
+    <div id="root"></div>
+    <div class="container"></div>
```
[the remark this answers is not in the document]
aalhommada: i will replace it with these two lines code in index.js

const mainBody = document.body;
  const mainContainerForRepo = createAndAppend('div', mainBody, { class: 'container' });

homework/index.js
```diff
@@ -30,18 +31,102 @@
...
+      const selectedIndex = event.target.value;
+      const selectedRepository = repositories[selectedIndex];
+      while (mainContainerForRepo.firstChild) {
+        mainContainerForRepo.removeChild(mainContainerForRepo.firstChild);
+      }
+      createDescription(selectedRepository, root);
+      createContributors(selectedRepository, root);
+    });
+  }
+  const mainContainerForRepo = document.querySelector('.container');
```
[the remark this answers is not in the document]
aalhommada: thanks for the explanation .
indeed it is outside the function because i build the document on two d
DIVs , one for the header ( which is root ) , and the other for Description , and Contributors ( which is  mainContainerForRepo ) , if this is not acceptable i will try to rebuild the document to make in one div (root) .?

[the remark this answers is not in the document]
remarcmij: Currently, you have this structure:

```html
<body>
  <div id="root">
    <header class="header">
      <h2 class="nav-title">HYF repositories</h2>
      <select id="repo-selector">
        <option value="0">AngularJS</option>
        <option value="1">tdd-game</option>
        <!-- ... -->
        </select>
    </header>
  </div>
  <div class="container">
    <div class="left-div">
      <!-- ... -->
    </div>
    <div class="right-div">
      <!-- ... -->
    </div>
  </div>
  <script src="./index.js"></script>
</body>
```

Since you have a separate `<header>` element that contains your complete header definition, you can add your container div to `root`, instead of `<body>`. So `root` becomes the true 'root', that contains your header as well as your main container, like this:

```html
<body>
  <div id="root">
    <header class="header">
      <h2 class="nav-title">HYF repositories</h2>
      <select id="repo-selector">
        <option value="0">AngularJS</option>
        <option value="1">tdd-game</option>
        <!-- ... -->
      </select>
    </header>
    <div class="container">
      <div class="left-div"><!-- ... --></div>
      <div class="right-div"><!-- ... --></div>
    </div>
  </div>
  <script src="./index.js"></script>
</body>
```

homework/index.js
```diff
@@ -30,18 +31,102 @@
...
+    const descriptionContainer = createAndAppend('div', mainContainerForRepo, {
+      class: 'left-div',
+    });
+    const table = createAndAppend('table', descriptionContainer, { class: 'table' });
+    const tbody = createAndAppend('tbody', table);
+    const details = [
+      { title: 'Repository', value: selectedRepository.name },
+      { title: 'Description', value: selectedRepository.description },
+      { title: 'Forks', value: selectedRepository.forks },
+      { title: 'Updated', value: selectedRepository.updated_at },
```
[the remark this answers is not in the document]
aalhommada: thanks for the nice note , i will do that

aalhommada:
I follow all the comments , not with the perfect styling , but i will work on it . 
hope now the homework will be accepted.

homework/index.js
```diff
@@ -30,18 +30,104 @@
...
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-      } else {
-        createAndAppend('pre', root, { text: JSON.stringify(data, null, 2) });
+        return;
       }
+      const header = createAndAppend('header', root, { class: 'header' });
+      const mainContainerForRepository = createAndAppend('div', root, { class: 'container' });
+
+      function createHeader(repositories) {
```
remarcmij: You have nested this function inside the `main` function. You have done this because you are accessing variables inside the scope of the containing `main` function. The problem with that is that it makes it not immediately clear when reading the code what these dependencies on the outer scope variables are. I would recommend that you move this function outside of the `main()` function to make these dependencies explicit. If you that in VSCode you will red underlines of the variables that are now undefined:

image


The solution is easy: just pass the missing variables as parameters to the function:

image

Now, in your `main()` function call `createHeader()` with the extra parameters needed:

```js
function main(url) {
    if (err) {
      return;
    }
    const container = createAndAppend('div', root, { class: 'container' });

    createHeader(data, header, container);
    createDescription(data[0], container);
    createContributors(data[0], container);
  });
}
```

Your `main()` function is now shorter and uncluttered by the nested function. And your `createHeader()` function gets everything it needs explicitly through its parameters list, taking all the guessing away.

homework/index.js
```diff
@@ -30,18 +30,104 @@
...
+        href: selectedRepository.html_url,
+        target: '_blank',
+      });
+
+      createAndAppend('td', linkToRepository, {
+        text: detail.valueOfName,
+        class: 'repository-data',
+      });
+      createAndAppend('td', tr, { text: detail.value, class: 'repository-data' });
+    });
```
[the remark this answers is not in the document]
aalhommada: thanks for mention that , i modify it .

homework/index.js
aalhommada: thanks , very clear explanation .

homework/index.js
```diff
@@ -30,18 +30,104 @@
...
+    const detailsHeader = [{ title: 'Repository', value: selectedRepository.name }];
+    const detailsBody = [
+      { title: 'Description', value: selectedRepository.description },
+      { title: 'Forks', value: selectedRepository.forks },
+      { title: 'Updated', value: new Date(selectedRepository.updated_at).toLocaleString() },
+    ];
+    detailsHeader.forEach(detail => {
+      const tr = createAndAppend('tr', tbody);
+      createAndAppend('td', tr, { text: detail.title, class: 'label' });
+      createAndAppend('a', tr, {
```
remarcmij: An `a` tag cannot be used as a child of a `tr` tag. A `tr` tag can only have `td` or `th` tags as its children.

aalhommada:
Thanks for the reviewing of the homework ,  i will fix the issue. 
 and improve the style in the following 2 homeworks ( week2 , week3,  
And about promises ,yes i should add it ASAP .

PR #286, opened 2019-07-17
Ramazan_Ince_JS3_HW3_Version1

RamazanInce:


homework/index.js
```diff
@@ -30,18 +15,178 @@
...
+  }
+
+  function createMainContainer(root) {
+    const mainContainer = createAndAppend('div', root, {
+      id: 'container',
+    });
+    return mainContainer;
+  }
+
+  function containerListContributor(mainContainer) {
```
remarcmij: Here is the conversation of the week 2 review. This function still has an incorrect, misleading name.

@remarcmij 7 days ago Member
This function does not deal with the contributor list but with the table containing repository details. The name createListContributor does not match what this function is doing.

 @RamazanInce
RamazanInce 6 days ago Author Member
I chance it as containerListContributor.

 @remarcmij
remarcmij 6 days ago • 
 Member
No, I'm afraid you are not getting the point. Contributors play no role at all in this function. Therefore the word 'contributors' should not be part of the function name.

 @RamazanInce
RamazanInce 6 days ago Author Member
containerListFotContributor may be ok but it will be to long
just tbodyCondainer can be also?
About naming the things I am not good enough. I need to work more...

 @remarcmij
remarcmij 6 days ago Member
You are still not getting my point. Again, this function has nothing to do with contributors. So the word contributors should not be part of the name or even be considered to be part of the name. A good name could, for instance, could be createRepoTableContainer or something like that. The table is used to display details about the repository and not about contributors.

homework/index.js
```diff
@@ -30,18 +15,178 @@
...
+    const rightContainer = createAndAppend('div', mainContainer, {
+      class: 'right-block frame',
+    });
+    createAndAppend('p', rightContainer, {
+      text: 'Contributions',
+      class: 'contributor',
+    });
+
+    return createAndAppend('ul', rightContainer, {
+      class: 'list1',
```
remarcmij: Avoid names with a number suffix, unless there is a clear reason to do so. In this case it looks like you didn't bother to find a good name.

homework/index.js
```diff
@@ -30,18 +15,178 @@
...
+        text: contributors[i].login,
+      });
+      createAndAppend('div', divCont, {
+        text: contributors[i].contributions,
+        class: 'badge',
+      });
+    }
+  }
+
+  async function renderRep(listContributors, repo, root, contributorList) {
```
remarcmij: Avoid unnecessary abbreviations. What do you mean with `Rep`? If you can type the full word `Contributors` in line 71, why then shouldn't you type the full word `Repository` here?

homework/index.js
```diff
@@ -30,18 +15,178 @@
...
+    } catch (error) {
+      createAndAppend('div', root, {
+        text: error.message,
+        class: 'alert-error',
+      });
+    }
   }
 
-  const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+  const HYF_REPS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
```
remarcmij: Why is your change an improvement over the existing name?

homework/App.js
```diff
@@ -12,20 +12,41 @@ class App {
...
+        throw new Error(`HTTP ${response.status} - ${response.statusText}`);
+      }
+      const repositories = await response.json();
+      this.repos = repositories
+        .sort((a, b) => a.name.localeCompare(b.name))
+        .map(repo => new Repository(repo));
+      this.repos.forEach((repo, index) => {
+        Util.createAndAppend('option', select, {
+          value: index,
+          id: repo.name,
```
remarcmij: What is the purpose of this `id`?

homework/style.css
```diff
@@ -1,3 +1,154 @@
+body {
```
remarcmij: The classes version of your app is not responsive:

image

homework/index.js
RamazanInce: The main problem I am not good in naming the things and secondly when I am trying something I used something sort to do quick sometimes i am using that one in many places and not to change for it to cause new problems, sometimes I am not taking it enough important. But none of those can not be excuse the excuse for that. I will try to improve my self more. I learn a lot from your reviews thanks..

homework/index.js
RamazanInce: Definitely. .

homework/index.js
RamazanInce: I will improve my self.

homework/style.css
RamazanInce: I change some of the id name for it to be better and forget to implement in css.
I will look in to it.

RamazanInce:
HI @remarcmij later on when I have time, I will implement all your advice then ask for your review again. :)

PR #288, opened 2019-07-19
Week3 homework

hanialkudsi:


homework/index.js
```diff
@@ -1,19 +1,21 @@
...
-    xhr.onload = () => {
-      if (xhr.status < 400) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
```
remarcmij: While this works fine, can you try and use the built-in `fetch` function instead of `fetchJSON`? When you do, make sure that you also handle HTTP errors. See the assignment description how you can force an HTTP error.

homework/style.css
```diff
@@ -1,3 +1,94 @@
...
+  box-shadow: 4px 3px 19px -5px rgba(0, 0, 0, 0.75);
+  padding: 2%;
+  flex: 1;
+}
+.right-div {
+  margin-left: 5px;
+}
+.left-div {
+  margin-right: 5px;
+}
```
remarcmij: In lines 52-57 you are making the assumption that the divs will always be displayed side by side. This does not hold true on a small screen width.

homework/style.css
```diff
@@ -1,3 +1,94 @@
...
+#root {
+  max-width: 650px;
+  margin: 0 auto;
+}
+/* td:first-of-type {
+  font-weight: bold;
+}
+p:first-of-type {
+  background-color: dodgerblue;
+} */
```
remarcmij: Remove dead code/CSS from a PR (i.e. commented out code)

fedefu:
@hanialkudsi What happened here? Did you made the necessary adjustments as agreed?

remarcmij:
I closed this PR due to inactivity with regards to follow-up and because a new class was starting JS3. I'll reopen it now.

homework/App.js
```diff
@@ -14,16 +14,37 @@ class App {
...
+      const data = await fetch(url);
+      const fetchedData = await (function error() {
+        if (data.ok) {
+          return data.json();
+        }
+        throw new Error('There is fetching error');
+      })();
+      this.repos = fetchedData
+        .sort((a, b) => a.name.localeCompare(b.name))
+        .map(repo => new Repository(repo));
```
remarcmij: You don't need this immediately invoked function (line 32) here. Instead, use something like this:

```js
const response = await fetch(url);
if (!response.ok) {
  throw new Error('There is fetching error');
}
const data = response.json();
this.repos = data
  .map(repo => new Repository(repo));
```

homework/App.js
```diff
@@ -14,16 +14,37 @@ class App {
...
+      fetchedData.forEach((repo, index) => {
+        Util.createAndAppend('option', select, {
+          text: repo.name,
+          value: index,
+        });
+      });
+      this.fetchContributorsAndRender([0]);
```
remarcmij: This method expects a number as an argument. You are passing it an array with a single numeric element. That the code works all the same is pure coincidence.

homework/Repository.js
```diff
@@ -12,9 +12,38 @@ class Repository {
...
+    Util.createAndAppend('td', rowUpDate, {
+      text: 'Updated:',
+      class: 'repo-update',
+    });
+    Util.createAndAppend('td', rowUpDate, {
+      text: `${new Date(this.repository.updated_at).toDateString()}`,
+    });
   }
 
   /**
```
remarcmij: In line 52 (not shown here), you are still using the `fetchJSON()` function. This is inconsistent with what you did in the `App` class. Can you modify this to also use `fetch`?

fedefu:
Thanks for the feedback @remarcmij! Can you please apply the feedback so we can close this pr @hanialkudsi? Thanks

homework/App.js
hanialkudsi: yes of course this way is much easier

homework/App.js
hanialkudsi: i  just fixed it 
thanks a lot!

homework/Repository.js
hanialkudsi: oh yesss, i found it and replace it with fetch

PR #291, opened 2019-07-24
Salih S Class 22 JS3 Week1 Homework

salih18:


homework/index.js
```diff
@@ -29,18 +33,218 @@
...
+  };
+  // create repository layout which is under the header
+  const createRepositoryLayout = repositoriesData => {
+    const mainParent = document.getElementById('main');
+    const leftColumn = createAndAppend('div', mainParent, {
+      class: 'd-flex flex-column repo-detail',
+      id: 'repo-detail',
+    });
+    //
+    const divWidgetRepoName = createAndAppend('div', leftColumn, {
```
remarcmij: Your four 'widget' rows are almost identical. In the spirit of _Don't Repeat Yourself_ (**DRY**) you could extract the common code into a function and call it four times, supplying the relevant data as parameters. For instance, you could create a function named `createRepositoryWidget` with the following declaration:

```js
function createRepositoryWidget(container, options = {}) {
  ...
}
```

You could then call that function like this, taking into consideration that the first widget need some extra work:

```js
const createRepositoryLayout = repositoriesData => {
  ...
  const firstWidget = createRepositoryWidget(leftColumn, {
    faClasses: 'fab fa-github',
    title: 'Repository Name:',
    valueTag: 'a',
    value: repositoriesData.name,
  });
  const anchorTag = firstWidget.lastChild;
  anchorTag.setAttribute('href', repositoriesData.html_url);
  anchorTag.setAttribute('target', '_blank');

  createRepositoryWidget(leftColumn, {
    faClasses: 'fas fa-pen-alt',
    title: 'Description',
    value: repositoriesData.description,
  });

  ...
}
```

I'm sure you will have no difficulty devising what the implementation should be of the `createRepositoryWidget` function should be.

homework/index.js
```diff
@@ -29,18 +33,218 @@
...
+            });
+            return;
+          }
+          createContributionsLayout(contributionsObj);
+        });
+      });
+    });
+  };
+  // create page layout under the root div
+  const buildWebPage = url => {
```
remarcmij: If you take out the function call at the end (see my comment further below) this function just deals with creating the header. So `createHeader` or something similar would seem a more accurate name.

homework/index.js
```diff
@@ -29,18 +33,218 @@
...
+      selected: '',
+      text: 'Select a repository',
+      class: 'text-muted options',
+    });
+    fetchRepositoriesAndContributions(url);
+  };
+  // main() ==> buildWebPage() ==> fetchRepositoriesAndContributions() ==> appendRepositoriesToSelect() ==>  createRepositoryLayout() && createContributionsLayout()
+  function main(url) {
+    buildWebPage(url);
   }
```
remarcmij: I would move your call to `fetchRepositoriesAndContributions` to `main()`. In that case `buildWebPage()` no longer needs a `url` parameter:

```js
function main(url) {
  buildWebPage();
  fetchRepositoriesAndContributions(url);
}
```

homework/index.js
```diff
@@ -29,18 +33,218 @@
...
+      src: './hyf.png',
+      alt: 'hack your future thumbnail',
+    });
+    const selectList = createAndAppend('select', header, {
+      id: 'repo-select',
+      class: 'form-control',
+      'aria-label': 'Hack Your Future Repositories Selection',
+    });
+    // Create Select Box and its options
+    createAndAppend('option', selectList, {
```
remarcmij: We used this in class (a dummy first option), but the homework assignment asks for you to select the first repo from the sorted list by default and display its details.

homework/index.js
salih18: `const anchorTag = firstWidget.lastChild;` 
my solution is a little bit different because the funct i wrote does not return value thats why i use `const anchorTag = leftColumn.firstChild.lastChild;`

Also for the anchor tag creation i have used
 `  createAndAppend(options.valueTag || 'h4', widgetContainer, {
      text: options.value,
    ...
    });`

homework/index.js
```diff
@@ -15,7 +19,7 @@
     xhr.onerror = () => cb(new Error('Network request failed'));
     xhr.send();
   }
-
+  // create dom elements and append them to parent
```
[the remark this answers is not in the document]
salih18: i have done it.

homework/index.js
```diff
@@ -29,18 +33,218 @@
     });
     return elem;
   }
+  // create contributions layout as cards
+  const createContributionsLayout = contributionsData => {
```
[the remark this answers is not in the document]
salih18: i have corrected it.

homework/index.js
```diff
@@ -29,18 +33,218 @@
...
+    const mainParent = document.getElementById('main');
+    const contributionCardsContainer = createAndAppend('div', mainParent, {
+      id: 'repo-contributor',
+    });
+    const cardsGroup = createAndAppend('div', contributionCardsContainer, {
+      class: 'card-group d-flex container-fluid',
+    });
+    contributionsData.forEach(contribution => {
```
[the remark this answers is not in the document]
salih18: i have done it.

homework/index.js
salih18: I corrected it.

homework/index.js
salih18: I have done it.

homework/index.js
remarcmij: Your `createRepositoryWidget` function could simply return the top-level container element of the widget as its return value:

```js
function createRepositoryWidget(container, options = {}) {
  ...
  return widgetContainer;
}


function createRepositoryLayout(repositoriesData) {
  ...
  const widgetContainer = createRepositoryWidget(leftColumn, {
    faClasses: 'fab fa-github',
    title: 'Repository Name:',
    value: repositoriesData.name,
    valueTag: 'a',
  });

  const anchorTag = widgetContainer.lastChild;
  anchorTag.setAttribute('href', repositoriesData.html_url);
  anchorTag.setAttribute('target', '_blank');
  ...
}
```

In other cases where you call `createRepositoryWidget()` and have no use for its return value, just ignore it (the same as is done with `createAndAppend()`).

homework/index.js
salih18: I have tried to get the dummy values for the first option, It is easy to get the repository value, but in my structure i haven't found a minor solution for the contributors dummy value because i wrote in the beginning only if get data when there is any change in the select-list.  Can you help me to find a solution?

remarcmij: The `change` event for the `<select>` element is only fired when you actually change the selection. This event is not fired when your application loads. So you need to call the same functions that your event handler calls manually, using the data for the first repo in the list.

salih18: Ok. I found now. Thank you.

remarcmij: Logically speaking, there should be a one-to-one correspondence between the current option displayed in the `<select>` box and the currently displayed repo details. So either you have a placeholder option which says **Select a Repository** without displaying any repo details, or you display the first repo name in the `<select>` box along with the corresponding repo details in the page.

In other words, you should now remove this placeholder option so that the `<select>` box displays the name of the first repo in the list (**alumni**).

homework/index.js
```diff
@@ -30,17 +33,197 @@
...
+        });
+      });
+
+    // dummy values
+    cb(repositoriesData[0], repositoriesData[0].contributors_url);
+
+    // create an eventlistener for select list options
+    selectList.addEventListener('change', () => {
+      selectList.parentElement.nextElementSibling.innerHTML = '';
+      cb(repositoriesData[selectList.value], repositoriesData[selectList.value].contributors_url);
```
remarcmij: As per my comment in slack, you don't need this second parameter for your callback as the `contributors_url` is a property of the selected repository object that you pass as the first parameter. So this code could be simplified as:

```js
function appendRepositoriesToSelect(repositoriesData, cb) {
  ...
  cb(repositoriesData[0]);

  // create an eventlistener for select list options
  selectList.addEventListener('change', () => {
    selectList.parentElement.nextElementSibling.innerHTML = '';
    cb(repositoriesData[selectList.value]);
  });
}
```

See also line 174 below.

homework/index.js
```diff
@@ -30,17 +33,197 @@
...
+          text: errRepository.message,
+          class: 'alert alert-danger display-3 text-center',
+        });
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+
+      appendRepositoriesToSelect(repositoriesData, (repositoryObj, contributorURL) => {
+        createRepositoryLayout(repositoryObj);
+        fetchJSON(contributorURL, (errContributors, contributorData) => {
```
remarcmij: With reference to my previous comment above, this line can now become:

```js
fetchJSON(repositoryObj.contributors_url, (errContributors, contributorData) => {
```

homework/index.js
salih18: I have removed the placeholder and pushed again . Thank you for your time, J

homework/index.js
```diff
@@ -30,17 +33,189 @@
...
+    const hyfLink = createAndAppend('a', headerDiv, {
+      href: 'https://www.hackyourfuture.net/',
+      target: '_blank',
+    });
+    createAndAppend('img', hyfLink, {
+      class: 'img-fluid rounded img-thumbnail mx-auto',
+      src: './hyf.png',
+      alt: 'hack your future thumbnail',
+    });
+    const selectList = createAndAppend('select', header, {
```
remarcmij: Now that you have removed the default option, the `selectList` variable is no longer needed. ESLint flags it as unused, causing the check by Travis for your PR to fail. Can you fix this final thing?

salih18: Yes, i didnt see them on the logs. I noticed now.

homework/index.js
salih18: I corrected es lint errors. and pushed again. Thank you for all.

PR #293, opened 2019-07-24
Submit Week1 homework - Oguzhan

oguzhanturgut:


homework/index.js
```diff
@@ -23,24 +23,147 @@
       const value = options[key];
       if (key === 'text') {
         elem.textContent = value;
+      } else if (key === 'html') {
+        elem.innerHTML = value;
```
remarcmij: By adding an option to set the `innerHTML` this function has now become ambiguous in the sense that it is no longer clear which of the two element-creation methods is to be used: `document.createElement()` or `innerHTML`.

homework/index.js
```diff
@@ -23,24 +23,147 @@
...
     });
     return elem;
   }
 
-  function main(url) {
-    fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+  const renderError = error => {
```
remarcmij: Here you have chosen to use an (arrow) function expression assigned to a `const`. In other places there are regular function declarations, with the `function` keyword. It is best to be consistent and choose one style or the other.

My personal preference is to use regular function declarations for (named) functions at the file/global level and use (anonymous) arrow functions for callbacks and event handlers. This makes it easy to discriminate between data constants (defined with `const`) and functions (declared with `function`) at the file/global level.

homework/index.js
```diff
@@ -23,24 +23,147 @@
...
+    const wrapper = document.getElementById('wrapper');
+    wrapper.innerHTML = '';
+    createAndAppend('div', wrapper, {
+      text: error.message,
+      class: 'alert',
+    });
+  };
+
+  const addRepoInfoRow = (parent, label, content) => {
```
remarcmij: 👍 for doing this in a separate function.

homework/index.js
```diff
@@ -23,24 +23,147 @@
...
+    const row = createAndAppend('tr', parent);
+    createAndAppend('th', row, {
+      text: label,
+      class: 'table-header',
+    });
+    createAndAppend('td', row, {
+      html: content,
```
remarcmij: Can you make this work without reverting to using `innerHTML` in `createAndAppend`? You only need it for the `<a>` tag.

homework/index.js
```diff
@@ -23,24 +23,147 @@
...
+
+  function renderRepoInfo(repository) {
+    const leftSide = document.getElementById('left-side');
+    leftSide.innerHTML = '';
+    const repoTable = createAndAppend('table', leftSide);
+    const repoTableBody = createAndAppend('tbody', repoTable);
+    addRepoInfoRow(
+      repoTableBody,
+      'Repository',
+      `<a href="${repository.html_url}" target="_blank">${repository.name}</a>`,
```
remarcmij: You could set the `text` attribute to the empty string `''` and 'fix up' the last child if the table row after your call to `addRepoInfoRow`:

```js
const row = addRepoInfoRow(repoTableBody, 'Repository', '');
createAndAppend('a', row.lastChild, {
  href: repository.html_url,
  target: '_blank',
  text: repository.name,
});
```

Of course, your `addRepoInfoRow` function should then return the row it created as its return value.

homework/index.js
```diff
@@ -23,24 +23,147 @@
...
+        return;
+      }
+      repositories.sort((a, b) => a.name.localeCompare(b.name));
+      repositories.forEach((repository, index) => {
+        createAndAppend('option', select, {
+          value: index,
+          text: repository.name,
+        });
+      });
+      createAndAppend('main', wrapper, {
```
remarcmij: Semantically speaking, you should only have one `main` element in your HTML page. I think `section` would be better here.

homework/index.js
```diff
@@ -23,24 +23,147 @@
...
+        class: 'left-side',
+        id: 'left-side',
+      });
+      renderRepoInfo(repositories[select.value]);
+      createAndAppend('main', wrapper, {
```
remarcmij: Same here.

homework/style.css
```diff
@@ -1,3 +1,127 @@
-.alert-error {
```
remarcmij: On a mobile phone the boxes appear to be clipped on the right-hand side.

image

homework/index.js
oguzhanturgut: I reverted the createAndAppend function to the unmodified version. You are right that my version was going outside of its concept somehow.

homework/index.js
oguzhanturgut: I changed all to regular function declaration.

homework/index.js
oguzhanturgut: Thank you 🙏

homework/index.js
oguzhanturgut: I reverted `createAndAppend` function to its original state.

homework/index.js
oguzhanturgut: That is very useful way for escaping from using `html` in `createAndAppend` function. Many thanks for the tip.

homework/index.js
oguzhanturgut: Changed to section tag. Thanks for your advice

homework/style.css
oguzhanturgut: Fixed the styling problem on mobile screens. It now shows as expected

oguzhanturgut:
Hello @remarcmij thanks for making time to review the PR. See you on Sunday 👋
Best Regards

PR #295, opened 2019-07-25
JS3 Week1 homework files

gulyener:
I could not make use of ESlint again (just like in JS2 week1 homework) so I couldn't test my homework, and there will be many errors.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -44,6 +44,7 @@
     const tr = createAndAppend('tr', tbody);
     createAndAppend('td', tr, { text: `${label}:`, class: 'label' });
     createAndAppend('td', tr, { text: value });
+    return tr;
   }
 
   function renderRepos(repoName, ul) {
@@ -58,10 +59,12 @@
           const li = createAndAppend('li', ul);
           const table = createAndAppend('table', li);
           const tbody = createAndAppend('tbody', table);
-          const tr1 = createAndAppend('tr', tbody);
-          createAndAppend('td', tr1, { text: `Name:`, class: 'label' });
-          const a = createAndAppend('a', tr1, { href: `${repo.html_url}` });
-          createAndAppend('td', a, { text: `${repo.name}` });
+          const firstRow = addRow(tbody, 'Name', '');
+          createAndAppend('a', firstRow.lastChild, {
+            text: `${repo.name}`,
+            href: repo.html_url,
+            target: '_blank',
+          });
           addRow(tbody, 'Description', `${repo.description}`);
           addRow(tbody, 'Forks', `${repo.forks}`);
           addRow(tbody, 'Updated', `${repo.updated_at.slice(0, 10)}`);
--- a/homework/index.js
+++ b/homework/index.js
@@ -35,9 +35,9 @@
   }
 
   function renderError(error) {
-    const body = document.body;
-    body.innerHTML = '';
-    createAndAppend('h1', body, { text: error.message, class: `alert-error` });
+    const root = document.getElementById('root');
+    root.innerHTML = '';
+    createAndAppend('h1', root, { text: error.message, class: `alert-error` });
   }
 
   function addRow(tbody, label, value) {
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -30,17 +34,110 @@
...
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+
+      function setDefault() {
+        const repoName = 'alumni';
+        renderRepos(repoName, ul);
+        renderContributions(repoName, ul1);
+      }
+      setDefault();
```
remarcmij: The benefit of this short directly invoked function is questionable. You could simply reduce it to using lines 125-126 directly. Taking note of my other comments, to display the first repo you could simply do:

```js
renderRepos(repositories[0], ul);
renderContributions(repositories[0], ul1);
```

homework/style.css
```diff
@@ -1,3 +1,107 @@
+* {
```
remarcmij: The responsiveness of your app can be improved:

image

homework/index.js
```diff
@@ -30,17 +34,110 @@
...
+      `https://api.github.com/repos/HackYourFuture/${repoName}/contributors`,
+      (err, repositories) => {
+        if (err) {
+          renderError(err);
+          return;
+        }
+        ul.innerHTML = '';
+        createAndAppend('h3', ul, { text: `Contributors:` });
+        repositories.forEach(repo => {
+          const a = createAndAppend('a', ul, { href: `${repo.html_url}` });
```
remarcmij: The `a` tag link should open in a new browser tab, as per the assignment,

homework/index.js
```diff
@@ -30,17 +34,110 @@
     return elem;
   }
 
+  function renderError(error) {
+    const body = document.body;
+    body.innerHTML = '';
```
remarcmij: Clearing out the whole body is a bit drastic. Perhaps just clear our the root div? But see my other comments below what this root div should contain.

homework/index.js
```diff
@@ -30,17 +34,110 @@
...
+          createAndAppend('img', tr1, { src: `${repo.avatar_url}` });
+          createAndAppend('td', tr1, { text: `${repo.login}` });
+          createAndAppend('td', tr1, { text: `${repo.contributions}` });
+        });
+      },
+    );
+  }
+
+  function createOptionElements(repositories, select) {
+    createAndAppend('option', select, { text: 'Select a repository', disabled: `disabled` });
```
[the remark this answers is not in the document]
gulyener: Okay thanks!

homework/index.js
```diff
@@ -30,17 +34,110 @@
...
+      }
+      ul.innerHTML = '';
+      repositories.forEach(repo => {
+        if (repo.name === repoName) {
+          const li = createAndAppend('li', ul);
+          const table = createAndAppend('table', li);
+          const tbody = createAndAppend('tbody', table);
+          const tr1 = createAndAppend('tr', tbody);
+          createAndAppend('td', tr1, { text: `Name:`, class: 'label' });
+          const a = createAndAppend('a', tr1, { href: `${repo.html_url}` });
```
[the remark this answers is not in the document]
gulyener: I made the changes you suggested thank you

homework/index.js
```diff
@@ -30,17 +34,110 @@
...
+      },
+    );
+  }
+
+  function createOptionElements(repositories, select) {
+    createAndAppend('option', select, { text: 'Select a repository', disabled: `disabled` });
+    repositories
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .forEach(repo => {
+        createAndAppend('option', select, { text: repo.name, value: repo.name });
```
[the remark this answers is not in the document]
gulyener: I spent a long time trying to make this work. I finally figured it out for renderRepos function but I still could not remove the forEach loop in renderContributions function. Was that necessary? Is there a way of doing this without looping through contributors?

homework/index.js
```diff
@@ -30,17 +34,110 @@
...
+          addRow(tbody, 'Forks', `${repo.forks}`);
+          addRow(tbody, 'Updated', `${repo.updated_at.slice(0, 10)}`);
+        }
+      });
+    });
+  }
+
+  function renderContributions(repoName, ul) {
+    fetchJSON(
+      `https://api.github.com/repos/HackYourFuture/${repoName}/contributors`,
```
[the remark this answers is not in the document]
gulyener: You are certainly right I removed it thank you

homework/index.js
```diff
@@ -30,17 +34,110 @@
...
+        createAndAppend('option', select, { text: repo.name, value: repo.name });
+      });
+  }
+
   function main(url) {
+    const root = document.getElementById('root');
+    const body = document.body;
+    const header = document.createElement('header');
+    body.prepend(header);
+    createAndAppend('h1', header, { text: 'HYF Repositories' });
```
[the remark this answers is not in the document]
gulyener: Thanks now I  added a main container div and two sub divs. I have to fix my css file but I haven't had the chance yet.

homework/index.js
```diff
@@ -30,17 +34,110 @@
...
+          const li = createAndAppend('li', ul);
+          const table = createAndAppend('table', li);
+          const tbody = createAndAppend('tbody', table);
+          const tr1 = createAndAppend('tr', tbody);
+          createAndAppend('td', tr1, { text: `Name:`, class: 'label' });
+          const a = createAndAppend('a', tr1, { href: `${repo.html_url}` });
+          createAndAppend('td', a, { text: `${repo.name}` });
+          addRow(tbody, 'Description', `${repo.description}`);
+          addRow(tbody, 'Forks', `${repo.forks}`);
+          addRow(tbody, 'Updated', `${repo.updated_at.slice(0, 10)}`);
```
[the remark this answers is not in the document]
gulyener: It looks much better now thanks

homework/index.js
gulyener: You are right thanks i added it

gulyener:
I think my page loads much faster after the changes I made according to your review, thank you so much for your time.

homework/index.js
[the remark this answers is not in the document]
remarcmij: You need the `.forEach()` for in `renderContributors` because you need to create a list item for each of contributor in the `contributors` array.

But perhaps you confused yourself by using the parameter name `repositories` for the `fetchJSON()` function inside `renderContributors()`. Your are not fetching `repositories`, you are fetching `contributors`.

```js
function renderContributions(repo, ul) {
  fetchJSON(repo.contributors_url, (err, contributors) => {
    if (err) {
      renderError(err);
      return;
    }
    ul.innerHTML = '';
    createAndAppend('h3', ul, { text: `Contributors:` });
      const a = createAndAppend('a', ul, { href: contributor.html_url });
      ...
    });
  });
}
```

homework/index.js
```diff
@@ -30,17 +34,102 @@
...
+    createAndAppend('td', tr, { text: value });
+    return tr;
+  }
+
+  function renderRepos(repo, div) {
+    fetchJSON(HYF_REPOS_URL, err => {
+      if (err) {
+        renderError(err);
+        return;
+      }
```
remarcmij: What is the purpose of this `fetchJSON()` call? You already have the data you need in the `repo` argument. You can remove lines 51-55 and the corresponding closing brace and everything will still work.

homework/index.js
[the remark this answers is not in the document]
gulyener: Yes I understand now thanks

homework/index.js
remarcmij: You gave me a thumbs up but didn't change your code 😞.

homework/index.js
```diff
@@ -30,17 +34,96 @@
...
+      });
+  }
+
   function main(url) {
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root);
+    createAndAppend('h1', header, { text: 'HYF Repositories' });
+    const select = createAndAppend('select', header);
+    const mainContainer = createAndAppend('div', root, { id: `main-container` });
+    const div1 = createAndAppend('div', mainContainer, { id: 'table-container' });
```
remarcmij: Try and avoid variable names with (meaningless) number suffixes. You already came up with good, intention-revealing names for the `id` attributes, so why not also use those names in camelCase form as the variable name?

```js
const tableContainer = createAndAppend('div', mainContainer, { id: 'table-container' });
```

homework/index.js
```diff
@@ -30,17 +34,96 @@
...
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+
+      function setDefault() {
+        renderRepos(repositories[0], div1);
+        renderContributions(repositories[0], ul);
+      }
+      setDefault();
+
+      createOptionElements(repositories, select);
```
remarcmij: You have a bug in your code. As you can see in the picture below, in the select box the `alumni` repo name is shown, whereas the details are shown for the `AngularJS` repo:

image

Can you solve this bug?

homework/index.js
```diff
@@ -30,17 +34,96 @@
...
+  function renderContributions(repo, ul) {
+    fetchJSON(repo.contributors_url, (err, contributors) => {
+      if (err) {
+        renderError(err);
+        return;
+      }
+      ul.innerHTML = '';
+      createAndAppend('h3', ul, { text: `Contributors:` });
+      contributors.forEach(contributor => {
+        const a = createAndAppend('a', ul, { href: contributor.html_url, target: '_blank' });
```
remarcmij: I forgot this one: a `ul` can only have `li` children. The `a` tag needs to be child of the `li`, not the other way around.

homework/index.js
gulyener: I think I fixed it now

homework/index.js
gulyener: You are right. I changed the variable names I will be more careful about this in the future

homework/index.js
gulyener: I put the a tag inside li tag

homework/index.js
gulyener: `function setDefault() {
        renderRepos(repositories[0], tableContainer);
        renderContributions(repositories[0], ul);
      }
      setDefault();`
Isn't this the right way?

remarcmij: I mean you do not need this immediately invoked function at all. It doesn't add value. You can just use the code directly in-line:

```js
...
createOptionElements(repositories, select);

renderRepos(repositories[0], tableContainer);
renderContributions(repositories[0], ul);

  renderRepos(repo, tableContainer);
  renderContributions(repo, ul);
});
...
```

homework/style.css
```diff
@@ -1,3 +1,121 @@
...
+h1 {
+  font-size: 3em;
+  letter-spacing: 2px;
+  color: rgb(210, 249, 222);
+}
+
+select {
+  color: rgb(36, 63, 111);
+  margin-top: 20px;
+  padding: 10px;
```
remarcmij: This padding causes the text inside the select box to become unreadable. Best to remove it altogether.

gulyener: Okay thank you so much for all the feedback!

PR #296, opened 2019-07-25
JavaScript3 week1/Shyar

ShyarK:


homework/index.js
```diff
@@ -16,31 +30,149 @@
...
+      createAndAppend('td', trForUpdated, {
+        text: 'Updated:',
+        class: 'td-header',
+      });
+
+      createAndAppend('td', trForUpdated, {
+        text: `${data.updated_at}`,
+      });
+    });
+  }
```
remarcmij: You are repeating yourself here four times: create a `tr` with two `td` elements. Can you make this code DRYer by creating a function called, say, `addRow` and call it four times, passing the data it needs for each row?

For the first row you need to do something special as it needs to contains the `a` tag. You could pass an empty string `''` as the value and bolt in the `a` tag after calling `addRow()` for the first row. Make sure that `addRow` returns the `tr` it created so that you can do something like:

```js
createAndAppend('a', firstRow.lastChild, {href: ...});
```

homework/index.js
```diff
@@ -16,31 +30,149 @@
...
+        text: `${data.forks_count}`,
+      });
+
+      createAndAppend('td', trForUpdated, {
+        text: 'Updated:',
+        class: 'td-header',
+      });
+
+      createAndAppend('td', trForUpdated, {
+        text: `${data.updated_at}`,
```
remarcmij: Can you display the date in a more user-friendly format? Hint: create a  JavaScript `Date` object and use one of its string methods to format the date string.

homework/index.js
```diff
@@ -16,31 +30,149 @@
...
+        target: '_blank',
+      });
+
+      createAndAppend('td', trForDescription, {
+        text: 'Description:',
+        class: 'description-td td-header',
+      });
+
+      createAndAppend('td', trForDescription, {
+        text: `${data.description}`,
```
remarcmij: There is no need to wrap the repository properties in a template literal. They are already strings or are automatically converted to strings:

```js
text: data.description,
```

homework/index.js
```diff
@@ -16,31 +30,149 @@
...
+      });
+
+      createAndAppend('td', trForUpdated, {
+        text: `${data.updated_at}`,
+      });
+    });
+  }
+
+  function renderContributions(repositoryName, contributesDiv) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repositoryName}/contributors`;
```
remarcmij: Do not construct this url yourself. Use the `contributors_url` property from the repository object. GitHub may change the internal format of the url, but will not change the property name as that is part of the published API.

homework/index.js
```diff
@@ -16,31 +30,149 @@
...
+
+      const ul = createAndAppend('ul', contributesDiv);
+
+      ul.addEventListener('click', e => {
+        window.open(`https://github.com/${e.target.textContent}`);
+      });
+
+      Object.keys(data).forEach(contributor => {
+        const listItem = createAndAppend('li', ul, { class: 'list flex-div' });
+        createAndAppend('img', listItem, { src: `${data[contributor].avatar_url}` });
```
remarcmij: The HTML validator complains about this `img` tag:

> An “img” element must have an “alt” attribute, except under certain conditions. For details, consult guidance on providing text alternatives for images.

homework/index.js
```diff
@@ -16,31 +30,149 @@
...
+      createAndAppend('td', trForUpdated, {
+        text: `${data.updated_at}`,
+      });
+    });
+  }
+
+  function renderContributions(repositoryName, contributesDiv) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repositoryName}/contributors`;
+
+    fetchJSON(url, (err, data) => {
```
remarcmij: The `data` we are fetching here are `contributors`. Why not use that name?

homework/index.js
```diff
@@ -16,31 +30,149 @@
...
+
+      createAndAppend('h1', contributesDiv, { class: 'contributes-header', text: 'Contributions' });
+
+      const ul = createAndAppend('ul', contributesDiv);
+
+      ul.addEventListener('click', e => {
+        window.open(`https://github.com/${e.target.textContent}`);
+      });
+
+      Object.keys(data).forEach(contributor => {
```
remarcmij: This line would be more readable like this:

```js
Object.keys(contributors).forEach(key => {
  const contributor = contributors[key];
   ...
  createAndAppend('img', listItem, { src:  contributor.avatar_url});
  ...
}
```

homework/index.html
```diff
@@ -1,23 +1,24 @@
...
-
-<body>
-  <div id="root"></div>
-  <script src="./index.js"></script>
-</body>
-
-</html>
\ No newline at end of file
+  <body>
+    <div class="root-div" id="root"></div>
```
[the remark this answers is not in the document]
ShyarK: Thanks a lot for having time to review my assignment! In the previous assignments with HTML and CSS, I used ID in CSS, then I got a comment from one of the teachers that, I shouldn't use the ID with CSS. Sorry about that but I am a little bit confused!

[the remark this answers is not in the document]
remarcmij: Generally speaking, for styling you should use classes. However, when using JavaScript we prefer an ID if we need to access a specific element in the DOM with `getElementById()`. If we already have an ID for that purpose then we might as well use it for CSS styling. Unless of course there are other elements that also need to same styling. In that case, an additional, shared class is required.

Note that it is not an error or bad practice to have this extra class. It is a matter of personal preference.

homework/index.js
```diff
@@ -16,31 +30,116 @@
...
 
   function main(url) {
-    fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
+    const root = document.getElementById('root');
+
+    fetchJSON(url, (err, data) => {
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-        return;
```
remarcmij: Why did you take out the return? By taking it out you now stumble into code that you should not be executing in case of errors. It will produce this error in the console:

```
index.js:123 Uncaught TypeError: Cannot read property 'sort' of undefined
    at index.js:123
    at XMLHttpRequest.xhr.onload (index.js:26)
```

homework/index.js
```diff
@@ -16,31 +30,116 @@
...
+      text: description,
+      class: className,
+    });
+    createAndAppend('td', tableRow, {
+      text: elementContent,
+    });
+    return tableRow;
+  }
+
+  function renderRepository(tableDiv, repository) {
```
remarcmij: This now looks nice and DRY. 👍

homework/index.js
```diff
@@ -16,31 +30,116 @@
...
+        return;
       }
+
+      contributesDiv.innerHTML = '';
+      createAndAppend('h1', contributesDiv, { class: 'contributes-header', text: 'Contributions' });
+
+      const ul = createAndAppend('ul', contributesDiv);
+
+      Object.keys(contributors).forEach(contributor => {
+        const hyperlink = createAndAppend('a', ul, {
```
remarcmij: An `ul` element can only have `li` elements as its children. The `a` tag should be a child of the `li`, not the other way around. And then, all elements that now have the `li` as their parent should have the `a` tag as their parent.

PR #297, opened 2019-07-25
Adham js3 homweork week1

Adham-Elias-Botours:
This is the first API homework.
***bad things in it: 
* the code is not tidy and it is difficult to be read. 
* thing I did not do: managing how the error message will be displayed.
*the css could also be much better, but it is now not so. There is a problem: the space on the left side is a bit  more than the space on the right side. 
***Good thing: it is working nice and responsive on the screen.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -41,7 +41,7 @@
     const url = intext.innerHTML;
     const infoContainer = document.getElementById('info_container');
     const infoRight = createAndAppend('div', infoContainer, { class: 'info_right' });
-    const title = createAndAppend('h4', infoRight, { text: 'info_right' });
+    createAndAppend('h4', infoRight, { text: 'info_right' });
     fetchJSON(url, (err, data) => {
       if (err) {
         renderError(err);
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+// modifying
 {
   const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
 
@@ -77,7 +78,7 @@
     });
   }
 
-  function viewRepInfo() {
+  function viewRepoInf() {
     fetchJSON(HYF_REPOS_URL, (err, data) => {
       if (err) {
         renderError(err);
@@ -135,7 +136,6 @@
       const infoContainer = createAndAppend('div', document.body, {
         id: 'info_container',
         class: 'info_container',
-        text: 'You can choose a repository name from the list',
       });
       if (err) {
         infoContainer.innerHTML = '';
@@ -154,15 +154,17 @@
         class: 'select_list',
         id: 'repositories_list',
       });
-      // createAndAppend('OPTION', selectEl, { text: 'Select one', disabled: 'disabled' });
       const repoNames = repositories
         .map(element => element.name)
         .sort((a, b) => a.localeCompare(b));
+      // (jijm)repoNames.forEach((repoName, index) => {
+      //   createAndAppend('option', selectEl, { text: repoName, value: index });
+      // });
       repoNames.forEach(element => {
         createAndAppend('OPTION', selectEl, { text: element });
       });
-      viewRepInfo();
-      selectEl.addEventListener('change', viewRepInfo);
+      viewRepoInf();
+      selectEl.addEventListener('change', viewRepoInf);
     });
   }
 
--- a/homework/index.js
+++ b/homework/index.js
@@ -161,7 +161,9 @@
         id: 'repositories_list',
       });
       createAndAppend('OPTION', selectEl, { text: 'Select one', disabled: 'disabled' });
-      const repoNames = repositories.map(element => element.name).sort();
+      const repoNames = repositories
+        .map(element => element.name)
+        .sort((a, b) => a.localeCompare(b));
       repoNames.forEach(element => {
         createAndAppend('OPTION', selectEl, { text: element });
       });
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+let data = '';
 {
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
@@ -7,6 +8,7 @@
     xhr.responseType = 'json';
     xhr.onload = () => {
       if (xhr.status >= 200 && xhr.status <= 299) {
+        data = xhr.response;
         cb(null, xhr.response);
       } else {
         cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
@@ -30,6 +32,28 @@
     return elem;
   }
 
+  function viewRepInfo() {
+    const repoName = document.getElementById('repositories_list').value;
+    const wantedRepo = data.filter(element => element.name === repoName)[0];
+    const ToDisplayLeft = [
+      ['Name', wantedRepo.name],
+      ['Description', wantedRepo.description],
+      ['Forks', wantedRepo.forks],
+      ['Last update', new Date(wantedRepo.updated_at)],
+    ];
+    const root = document.getElementById('root');
+
+    if (root.childNodes[1]) {
+      root.removeChild(root.childNodes[1]);
+    }
+    const inRootEl = createAndAppend('div', root, { class: 'in_root_left' });
+    ToDisplayLeft.forEach(array =>
+      array.forEach(element => createAndAppend('p', inRootEl, { text: element })),
+    );
+    console.log(ToDisplayLeft);
+    console.log(typeof root);
+  }
+
   function main(url) {
     fetchJSON(url, (err, repositories) => {
       const root = document.getElementById('root');
@@ -37,14 +61,16 @@
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
         return;
       }
-      createAndAppend('SELECT', root, { class: 'select_list', id: 'repositories_list' });
-      // createAndAppend('pre', root, { text: JSON.stringify(repositories[0].name, null, 2) });
-      const selectEl = document.getElementById('repositories_list');
+      const selectEl = createAndAppend('SELECT', root, {
+        class: 'select_list',
+        id: 'repositories_list',
+      });
       createAndAppend('OPTION', selectEl, { text: 'Select one', disabled: 'disabled' });
       const repoNames = repositories.map(element => element.name).sort();
       repoNames.forEach(element => {
         createAndAppend('OPTION', selectEl, { text: element });
       });
+      selectEl.addEventListener('change', viewRepInfo);
     });
   }
 
--- a/homework/index.js
+++ b/homework/index.js
@@ -79,11 +79,8 @@
   }
 
   function viewRepoInf(arrayOfObjects, indexNumber) {
-    const test = arrayOfObjects[indexNumber].name;
-    console.log(test);
-
     const wantedObject = arrayOfObjects[indexNumber];
-    console.log(wantedObject);
+
     const infoContainer = document.getElementById('info_container');
     infoContainer.innerHTML = '';
     const infoLeft = createAndAppend('div', infoContainer, {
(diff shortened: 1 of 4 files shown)


Adham-Elias-Botours:
Now I am wondering about using (data attributes) instead of using DOM element to save data "in my case web link"? is that a good technique?
https://developer.mozilla.org/en-US/docs/Learn/HTML/Howto/Use_data_attributes
Thanks in advance

remarcmij:
If you create all your HTML using JavaScript there is not much point in using `data-` attributes, as you already have direct access to all data used for creating the HTML elements. The `data-` attributes are useful if your HTML comes from another source, e.g. a server-generated template, etc.

homework/index.js
```diff
@@ -16,31 +37,139 @@
...
+      createAndAppend('span', LastUpdate, {
+        class: 'span_infoLeft',
+        text: new Date(wantedRepo.updated_at).toLocaleString(),
+      });
+      const contributorsUrl = createAndAppend('span', infoLeft, {
+        text: wantedRepo.contributors_url,
+        id: 'contributors_url',
+        class: 'hidden',
+      });
+      contributorsUrl.addEventListener('load', Contributions(contributorsUrl));
```
remarcmij: The purpose of this hidden element and the event listener is unclear to me. You can simply replace lines 129-134 with

```js
Contributions(wantedRepo.contributors_url);
```

and change your `Contributions` function as follows:

```js
function Contributions(url) {
  const infoContainer = document.getElementById('info_container');
  const infoRight = createAndAppend('div', infoContainer, {
  ...
}
```

Function names should be in camelCase (except for constructor function) and include a verb that indicates the action the function perform. Your `Contributions` function could for instance be named `fethcAndRenderContributions`.

homework/index.js
```diff
@@ -16,31 +37,139 @@
...
-      } else {
-        elem.setAttribute(key, value);
+  function Contributions(inText) {
+    const url = inText.innerHTML;
+    const infoContainer = document.getElementById('info_container');
+    const infoRight = createAndAppend('div', infoContainer, {
+      class: 'info_right',
+      text: 'Contributions:',
+    });
+    fetchJSON(url, (err, data) => {
```
remarcmij: Avoid generic names such as `data` if a better name can be found. In this case you are fetching `contributors`.

homework/index.js
```diff
@@ -16,31 +37,139 @@
...
+      if (err) {
+        renderError(err);
+        return;
       }
+      data.forEach(element => {
```
remarcmij: Say what you mean:

```js
```

homework/index.js
```diff
@@ -16,31 +37,139 @@
...
+        const ContributorEl = createAndAppend('a', infoRight, {
```
remarcmij: Use camelCase variable and parameter names: `contributorEl`

homework/index.js
```diff
@@ -16,31 +37,139 @@
...
+          text: element.contributions,
+        });
+        createAndAppend('hr', infoRight, {
+          class: 'hr_contributor',
+        });
+      });
+    });
+  }
+
+  function viewRepInfo() {
```
remarcmij: If you want to abbreviate the word repository I would make it `repo` instead of `rep` . The word `repo` is often use in speech to mean repository. But nobody would conclude that you mean repository when you say the word `rep`.

homework/normalize.css
```diff
@@ -0,0 +1,349 @@
+/*! normalize.css v8.0.1 | MIT License | github.com/necolas/normalize.css */
```
remarcmij: On a wide screen you UI could be improved by reducing the width and centering the content. Stretching it out across the screen is a bit over the top.

image

homework/index.js
Adham-Elias-Botours: for now , I replaced what I can. (the new work is on the way)

homework/index.js
Adham-Elias-Botours: By this "contributorEl" I mean : contributor element. So , it is really a camelCase . But yes, the name appears  confusing. 
for the next times I should be more careful with the names . I have to use one standard way in the whole file.

remarcmij: camelCase names should start with a lowercase letter. `ContributorEl` -> `contributorEl`. I don't object to the name itself, although I would probably use a few more letters myself, e.g. `contributorElem`.

homework/index.js
Adham-Elias-Botours: I changed the names is the whole function. 
I should be more careful about that in my whole work.

homework/index.js
```diff
@@ -16,31 +37,139 @@
...
+    fetchJSON(HYF_REPOS_URL, (err, data) => {
+      if (err) {
+        renderError(err);
+        return;
+      }
+      const repoName = document.getElementById('repositories_list').value;
+      const wantedRepo = data.filter(element => element.name === repoName)[0];
+      const infoContainer = document.getElementById('info_container');
+      infoContainer.innerHTML = '';
+      const infoLeft = createAndAppend('div', infoContainer, {
```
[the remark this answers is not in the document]
Adham-Elias-Botours: yes that is true. 
I have to practice more with tables and  lists (ul. ol).

homework/index.js
Adham-Elias-Botours: I deleted that hidden element . and the function is adjusted.

homework/index.js
```diff
@@ -16,31 +37,139 @@
...
+        class: 'info_left',
```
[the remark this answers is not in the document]
Adham-Elias-Botours: but if it is as id! than should we use _ underscore?
I think that the answer is right.
and then That what makes me confused.

homework/index.js
Adham-Elias-Botours: I should use verbs as names for function. while I was writing this homework ,I forgot all the standers of making names.
I have to be careful with that.

homework/index.js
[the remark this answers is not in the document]
remarcmij: kebab-case uses hyphens (`-`), not underscores.

Adham-Elias-Botours:
hi @remarcmij ,
the last important adjustments are:
-few names are changed (JS variables and functions..)
-the hidden HTML element is deleted and the wanted data passed as parameter to the next function.
- the repositories data is sorted, and the requested values added to the Select element as value and index.
- all the HTML element are moved to be inside the root element.
-one XMLHttpRequest is deleted, the wanted data passed in  as parameter to the function. 
Best regards

homework/index.js
```diff
@@ -1,162 +1,164 @@
...
-          href: contributor.html_url,
-          target: '_blank',
+    function createAndAppend(name, parent, options = {}) {
+        const elem = document.createElement(name);
+        parent.appendChild(elem);
+        Object.keys(options).forEach(key => {
+            const value = options[key];
+            if (key === 'text') {
+                elem.textContent = value;
+            } else if (key === 'html') {
```
remarcmij: Adding this option waters down the concept of this function. It adds ambiguity on how this function should be used: you can now create child elements through `innerHTML` in addition to creating elements one at the time and explicitly making one the parent of the other.

Adham-Elias-Botours:
Now I can say:  I finish working on this homework for this week. The result sounds very good to me (comparing with the first version).
In this last version:

- All the names all changed (variables, functions, CSS classes.). I used the _ underscore just in the names of the IDs.
- The function aadAndAppend is modified , so it can deal with innerHTML.
- The style for big screens is modified. 
Thanks in advanced.

Adham-Elias-Botours:
to view it online:
https://adham-elias-botours.github.io/my-js3/

homework/index.js
Adham-Elias-Botours: yes , I can now  create child elements through innerHTML.  I think that it gives me more options, and bigger concept . So I can decide what to use and when. 
Sure InnerHTML just when it is really necessary and makes the code shorter.

remarcmij: You are of course entitled to your own opinion. I just happen to disagree.

homework/index.js
```diff
@@ -1,6 +1,29 @@
...
+        elem.textContent = value;
+      } else if (key === 'html') {
+        elem.innerHTML = value;
+      } else {
+        elem.setAttribute(key, value);
+      }
+    });
+    return elem;
+  }
+  // ********************************************************
```
remarcmij: Most developers, including myself,  would prefer a simple blank line between function definitions. Just like you would use a blank line, rather than a drawn line to separate paragraphs in a piece of text.

homework/index.js
```diff
@@ -15,32 +38,127 @@
...
-    const elem = document.createElement(name);
-    parent.appendChild(elem);
-    Object.keys(options).forEach(key => {
-      const value = options[key];
-      if (key === 'text') {
-        elem.textContent = value;
-      } else {
-        elem.setAttribute(key, value);
+  // *********************************************************
+  function ShowContributions(urlText) {
```
remarcmij: Function names should be camelCase (start with a lowercase letter), unless it is a constructor function (which in this case, it is not).

homework/index.js
```diff
@@ -15,32 +38,127 @@
...
+          class: 'hr-contributor',
+        });
+      });
     });
-    return elem;
   }
-
-  function main(url) {
+  // *************************************************
+  function ShowRepoInfo(arrayOfObjects, indexNumber) {
```
remarcmij: 1. camelCase: `showRepoIInfo`.
2. You are passing too much information to this function. It only needs to know about the selected repository (you called it `wantedRepo`. In general, you should pass just enough info to functions to do their, not anything more than that (i.e. on a need-to-know basis). This helps to make your code more robust as the called function then cannot mess up the data is not supposed to have access to.

In this case, your function could be defined as:

```js
function showRepoInfo(wantedRepo) {
```

Then, when you call this function, for instance in line 158 you could call it like:

```js
showRepoInfo(repositories[selectEl.value]);
```

homework/index.js
```diff
@@ -15,32 +38,127 @@
...
+    const wantedRepo = arrayOfObjects[indexNumber];
+    const infoContainer = document.getElementById('info_container');
+    infoContainer.innerHTML = '';
+    const infoLeft = createAndAppend('div', infoContainer, {
```
remarcmij: An HTML table would provide a nicer layout here.

homework/index.js
```diff
@@ -15,32 +38,127 @@
...
+      class: 'p-info-left',
+      text: 'Name:',
+    });
+    createAndAppend('a', nameAndLink, {
+      class: 'repo-link',
+      text: wantedRepo.name,
+      href: wantedRepo.html_url,
+      target: '_blank',
+    });
+    const Description = createAndAppend('p', infoLeft, {
```
remarcmij: Variable names should be camelCase: `description`.

homework/index.js
```diff
@@ -15,32 +38,127 @@
...
+        text: 'HYF Repositories',
+      });
+      const selectEl = createAndAppend('select', header, {
+        class: 'select-list',
+        id: 'repositories_list',
+      });
+      const infoContainer = createAndAppend('div', root, {
+        id: 'info_container',
+        class: 'info-container',
+      });
```
remarcmij: If I'm not mistaken, you asked this question in class about where to put the code that is not dependent on the data you get from `fetchJSON()`. You could do it like this:

```js
function starter(url) {
  createAndAppend('span', header, {
    class: 'page-title',
    text: 'HYF Repositories',
  });
  const selectEl = createAndAppend('select', header, {
    class: 'select-list',
    id: 'repositories_list',
  });
  const infoContainer = createAndAppend('div', root, {
    id: 'info_container',
    class: 'info-container',
  });

    if (err) { 
   ...
```

homework/index.js
Adham-Elias-Botours: yes , that is my typing mistake while I was renaming almost everything.

homework/index.js
Adham-Elias-Botours: Yes I agree 100% and I have been thinking about that form yesterday .

homework/index.js
Adham-Elias-Botours: I am going to try that. I am reading the code in JavaScript 3 Examples. 
I am still having problem there with <a> tag. 
Actually just now (while I am writing here) I got an idee to solve it. 
I think it is going to work, otherwise I will ask again about it.

homework/index.js
Adham-Elias-Botours: type mistake. I will fix it in the next  version.

homework/index.js
Adham-Elias-Botours: yes, exactly that was my question.  Thanks for the explanation. 
Best regards.

PR #298, opened 2019-07-26
add week1 homework/Adil

musafir928:


homework/index.js
```diff
@@ -30,14 +30,93 @@
     return elem;
   }
 
+  function createOption(repositories, select) {
+    createAndAppend('option', select, { text: 'Select a repository', disabled: 'disabled' });
+    repositories
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .forEach(repository => {
+        createAndAppend('option', select, { text: repository.name, value: repository.name });
```
remarcmij: In the assigment, it was hinted to use the (sorted) array index for the `value` attribute of the `option` element, e.g.:

```js
.forEach((repository, index) => {
  createAndAppend('option', select, { text: repository.name, value: index });
});
```

By using this index it is now easy to get the repository object from the `repositories` array that corresponds to the selected repo in your event handler in `main()`:

```js
  const repository = repositories[select.value];
  renderRepoBasisInfo(repository, ul);
  renderContributors(repository, ulContributions);
});
```

The variable `repository` is now no longer just the name of the selected repository but the repository object itself, containing all the details you need to render the data.

And of course, the object for the first repository to display is `repositories[0]`.

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+      const trUpd = createAndAppend('tr', tbody);
+      createAndAppend('th', trUpd, { text: 'Updated at:' });
+      createAndAppend('td', trUpd, { text: repo.updated_at });
+    });
+  }
+
+  function renderContributors(repository, ul) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repository}/contributors`;
+    fetchJSON(url, (err, contributions) => {
+      if (err) return;
```
remarcmij: You should render the error to the page. Otherwise you leave the user in the dark why he/she is not seeing contributor details.

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+      createAndAppend('th', trForks, { text: 'Forks:' });
+      createAndAppend('td', trForks, { text: repo.forks_count });
+      const trUpd = createAndAppend('tr', tbody);
+      createAndAppend('th', trUpd, { text: 'Updated at:' });
+      createAndAppend('td', trUpd, { text: repo.updated_at });
+    });
+  }
+
+  function renderContributors(repository, ul) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repository}/contributors`;
```
remarcmij: You should not construct this url yourself. Instead, use the `contributors_url` property of the repository object. GitHub might change the internal format of the url, causing your code to break, but is bound to keep supporting the documented `contributors_url` property.

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+    fetchJSON(url, (err, contributions) => {
+      if (err) return;
+      ul.innerHTML = '<h3>Contributions</h3>';
```
remarcmij: Have you forgotten about `createAndAppen()`. This here is a style break from the rest of the code.

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+      });
+  }
+
+  function renderRepoBasisInfo(repository, ul) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repository}`;
+    const container = document.getElementById('root');
+    fetchJSON(url, (err, repo) => {
+      if (err) {
+        createAndAppend('div', container, { text: err.message, class: 'alert-error' });
+      }
```
remarcmij: In case of an error, you render the error to the page but then stumble into code you should be executing. What about adding a `return` statement?

But, as I mentioned, we do not need this fetch at all.

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+    createAndAppend('option', select, { text: 'Select a repository', disabled: 'disabled' });
+    repositories
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .forEach(repository => {
+        createAndAppend('option', select, { text: repository.name, value: repository.name });
+      });
+  }
+
+  function renderRepoBasisInfo(repository, ul) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repository}`;
```
remarcmij: There is no need to fetch the repo details again. We already have this data (for all repos), fetched in `main()`. You pass the selected repository object as a parameter to this function rather than fetching it again.

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+      const div = createAndAppend('div', root, { class: 'info-box' });
+      const ul = createAndAppend('ul', div, { class: 'info-ul' });
+      // right side
+      const right = createAndAppend('div', root, { class: ' contributions' });
+      const ulContributions = createAndAppend('ul', right, { class: 'right-ul' });
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
```
remarcmij: It is best to sort the repositories at this point, as soon as you have established that there was no error.


In your event handler, as I mentioned elsewhere, you can then use `repositories[select.value]` to get the select repository object.

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+  }
+
+  function renderRepoBasisInfo(repository, ul) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repository}`;
+    const container = document.getElementById('root');
+    fetchJSON(url, (err, repo) => {
+      if (err) {
+        createAndAppend('div', container, { text: err.message, class: 'alert-error' });
+      }
+      ul.innerHTML = '';
```
remarcmij: This `ul` with a single `li` that contains just `table` doesn't make much sense here. Why do we need a list? This the HTML you generate:

```html
<div class="info-box">
  <ul class="info-ul">
    <li class="info-list">
      <table>
        <tbody>
          <tr>
            <th>Repository:</th>
          </tr>
          <tr>
            <th>Description:</th>
          </tr>
          <tr>
            <th>Forks:</th>
            <td>9</td>
          </tr>
          <tr>
            <th>Updated at:</th>
            <td>2019-06-21T11:38:21Z</td>
          </tr>
        </tbody>
      </table>
    </li>
  </ul>
</div>
```

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+
+      const li = createAndAppend('li', ul, { class: 'info-list' });
+      const table = createAndAppend('table', li);
+      const tbody = createAndAppend('tbody', table);
+      const trName = createAndAppend('tr', tbody);
```
remarcmij: You are repeating yourself here four times: create a `tr` with two `td` children. Can you make your code DRYer by introducing a function called, say, `addRow` that you call four times, supplying the information it needs for each row?

For the first element, containing the `a` you need to do something extra. If you were to create an `addRow` function, you could do something like this:

```js
createAddAppend('a', firstRow.lastChild, { href: ..., text: .... });
```

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+      createAndAppend('a', tdName, { text: repo.name, href: repo.html_url, target: '_blank' });
+      const trDes = createAndAppend('tr', tbody);
+      createAndAppend('th', trDes, { text: 'Description:' });
+      createAndAppend('td', trDes, { text: repo.description });
+      const trForks = createAndAppend('tr', tbody);
+      createAndAppend('th', trForks, { text: 'Forks:' });
+      createAndAppend('td', trForks, { text: repo.forks_count });
+      const trUpd = createAndAppend('tr', tbody);
+      createAndAppend('th', trUpd, { text: 'Updated at:' });
+      createAndAppend('td', trUpd, { text: repo.updated_at });
```
remarcmij: Can you display this date in a more user-friendly format? Hint: create a JavaScript `Date` object from the date string and use one of its string methods to format the date.

homework/index.js
```diff
@@ -30,14 +30,93 @@
     return elem;
   }
 
+  function createOption(repositories, select) {
```
remarcmij: Since you are creating more than one option, use the plural form `createOptions`.

homework/index.js
```diff
@@ -30,14 +30,93 @@
...
+  }
+
+  function renderContributors(repository, ul) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repository}/contributors`;
+    fetchJSON(url, (err, contributions) => {
+      if (err) return;
+      ul.innerHTML = '<h3>Contributions</h3>';
+      contributions.forEach(contributor => {
+        const li = createAndAppend('li', ul, { class: 'contributor' });
+        const contributorTable = createAndAppend('table', li);
```
remarcmij: An HTML `table` is not the best choice here. It is better to use divs in combination with flex-box to position and line up the three components making up the list item.

It would also be nice if you could click anywhere on the list item to go to the contributors page. You can achieve that by wrapping the inner content of the list item in an `a` tag and set its `href` to the appropriate url.

homework/style.css
```diff
@@ -1,3 +1,116 @@
+* {
```
remarcmij: On a wide screen your UI is stretched out ovet the full width. It would be better to limit the width to, say, 960px max and center the content.

Note also that the table columns do not line up correctly vertically.

image

homework/index.js
```diff
@@ -30,14 +30,93 @@
     return elem;
   }
 
+  function createOption(repositories, select) {
+    createAndAppend('option', select, { text: 'Select a repository', disabled: 'disabled' });
```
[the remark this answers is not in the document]
musafir928: Thank you for your time and comment, I began to fix these issues.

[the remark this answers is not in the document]
musafir928: I have solved this problem. After I solved all the technical problems, I will improve my code's look.

homework/index.js
musafir928: solved;

homework/index.js
musafir928: solved;

remarcmij:
Superseded by #305

PR #299, opened 2019-07-26
submit Js3 Week1 Hw -- Mehmet Ergorgec

Mehmet-Ergorgec:
Hello @remarcmij ;
This week was very challenging for me. I have listened your videos 2 times:) In addition to hw, I got CORS errors and 403:forbidden errors. It took much time to solve.

I will try to dry my code.
Actually I havent done default settings for section element. I will work on it until Sunday..
And css file needs to be improved. I will work on it until Sunday.
I think after your review I am gonna learn lots of things.

What I have learned from this hw:
1- Divide problems into small pieces.
2- Console.log() is almost everything .
3- Something fires while in break.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -15,6 +15,20 @@
     return elem;
   }
 
+  function addRow(tbody, label, value) {
+    const tr = createAndAppend('tr', tbody);
+    createAndAppend('td', tr, { text: `${label}:`, class: 'label' });
+    createAndAppend('td', tr, { text: value });
+  }
+
+  function addHref(value, parent, text) {
+    createAndAppend('a', value, {
+      href: parent.html_url,
+      target: '_blank',
+      text: `${text}`,
+    });
+  }
+
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
     xhr.open('GET', url);
@@ -41,23 +55,11 @@
       const tr1 = createAndAppend('tr', tbody);
       createAndAppend('td', tr1, { text: 'Repository:', class: 'label' });
       createAndAppend('td', tr1);
-      createAndAppend('a', tr1.lastChild, {
-        href: repoInfo.html_url,
-        target: '_blank',
-        text: repoInfo.name,
-      });
+      addHref(tr1.lastChild, repoInfo, repoInfo.name);
 
-      const tr2 = createAndAppend('tr', tbody);
-      createAndAppend('td', tr2, { text: 'Description:', class: 'label' });
-      createAndAppend('td', tr2, { text: `${repoInfo.description}` });
-
-      const tr3 = createAndAppend('tr', tbody);
-      createAndAppend('td', tr3, { text: 'Forks:', class: 'label' });
-      createAndAppend('td', tr3, { text: `${repoInfo.forks}` });
-
-      const tr4 = createAndAppend('tr', tbody);
-      createAndAppend('td', tr4, { text: 'Updated:', class: 'label' });
-      createAndAppend('td', tr4, { text: new Date(repoInfo.updated_at).toLocaleString() });
+      addRow(tbody, 'Description', repoInfo.description);
+      addRow(tbody, 'Forks', repoInfo.forks);
+      addRow(tbody, 'Updated', new Date(repoInfo.updated_at).toLocaleString());
     });
   }
 
@@ -72,11 +74,7 @@
           alt: `${contributor.login} photo`,
         });
         createAndAppend('h1', listItems);
-        createAndAppend('a', listItems.lastChild, {
-          href: contributor.html_url,
-          target: '_blank',
-          text: `${contributor.login}`,
-        });
+        addHref(listItems.lastChild, contributor, contributor.login);
         createAndAppend('p', listItems, { text: `${contributor.contributions}` });
       });
     });
--- a/homework/index.js
+++ b/homework/index.js
@@ -69,19 +69,21 @@
       ulContributorInfo.innerHTML = '';
 
       contributors.forEach(contributor => {
-        const contributorLink = createAndAppend('a', ulContributorInfo, {
+        const listItem = createAndAppend('li', ulContributorInfo, {
+          class: 'contributor-list-item',
+        });
+        const contributorLink = createAndAppend('a', listItem, {
+          class: 'contributorLink-a',
           href: contributor.html_url,
           target: '_blank',
         });
-        const listItem = createAndAppend('li', contributorLink, {
-          class: 'contributor-list-item',
-        });
-        createAndAppend('img', listItem, {
+
+        createAndAppend('img', contributorLink, {
           src: contributor.avatar_url,
           alt: `${contributor.login} photo`,
         });
-        createAndAppend('h1', listItem, { text: contributor.login });
-        createAndAppend('p', listItem, { text: contributor.contributions });
+        createAndAppend('h1', contributorLink, { text: contributor.login });
+        createAndAppend('p', contributorLink, { text: contributor.contributions });
       });
     });
   }
@@ -111,17 +113,17 @@
       }
       repositories.sort((a, b) => a.name.localeCompare(b.name));
       createOptionElements(repositories, select);
-      const repo = repositories[select.value];
+      let repo = repositories[select.value];
       renderRepositories(repo, repoInfo);
       renderContributors(repo, contributorInfo);
 
       select.addEventListener('change', () => {
-        // const repo = repositories[select.value];
+        repo = repositories[select.value];
         renderRepositories(repo, repoInfo);
         renderContributors(repo, contributorInfo);
       });
     });
   }
   const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
-  window.onload = main(HYF_REPOS_URL);
+  window.onload = () => main(HYF_REPOS_URL);
 }
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,6 +1,20 @@
 'use strict';
 
 {
+  function createAndAppend(name, parent, options = {}) {
+    const elem = document.createElement(name);
+    parent.appendChild(elem);
+    Object.keys(options).forEach(key => {
+      const value = options[key];
+      if (key === 'text') {
+        elem.textContent = value;
+      } else {
+        elem.setAttribute(key, value);
+      }
+    });
+    return elem;
+  }
+
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
     xhr.open('GET', url);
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -16,31 +30,97 @@
...
+      class: 'li-container',
+    });
+
+    fetchJSON(url, (error, data) => {
+      if (error) {
+        createAndAppend('div', root, { text: error.message, class: 'alert-error' });
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+      createOptionElements(data, select);
```
remarcmij: Create `option` elements from the sorted array. Take the hint from the assignment to set the value attribute from each option element to the array index of the corresponding repository object, e.g. in your function `createOptionElements`:

```js
repositories.forEach((repo, index) => {
  createAndAppend('option', select, { text: repo.name, value: index });
});
```

Now, in your event hander you can use this index to directly get the repository object that corresponds to the selected repo from the select box like this:

```js
  renderRepositories(repo, ulRepoInfo);
  renderContributors(repo, ulContributorInfo);
});
```

No need to fetch details for the selected repo again. Just pass the information already in your possession as an argument to the functions that require the information.

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
+        createAndAppend('p', listItems, { text: `${contributor.contributions}` });
+      });
+    });
+  }
+
+  function createOptionElements(repositories, select) {
+    createAndAppend('option', select, {
+      text: 'Select a repository',
+      disabled: 'disabled',
+    });
```
remarcmij: The assignment specifies that you should display the first repository from the (sorted) list at application startup. Therefore you should not add this placeholder option to the list. At startup the select box should show the name of the first repo (at present 'alumni') and the page should render the details of the repo. 

Note that the repository object for the first repo is `repositories[0]`.

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
+      const tr4 = createAndAppend('tr', tbody);
+      createAndAppend('td', tr4, { text: 'Updated:', class: 'label' });
+      createAndAppend('td', tr4, { text: new Date(repoInfo.updated_at).toLocaleString() });
     });
-    return elem;
+  }
+
+  function renderContributors(repoName, ulContributorInfo) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repoName}/contributors`;
+    fetchJSON(url, (err, contributors) => {
```
remarcmij: You are not handling the error.

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
-      const value = options[key];
-      if (key === 'text') {
-        elem.textContent = value;
-      } else {
-        elem.setAttribute(key, value);
-      }
+  function renderRepositories(repo, ulRepoInfo) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repo}`;
+    fetchJSON(url, (err, repoInfo) => {
+      ulRepoInfo.innerHTML = '';
```
remarcmij: You are not handling the error (leaving aside that you do not need this fetch at all).

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
+
+      const tr4 = createAndAppend('tr', tbody);
+      createAndAppend('td', tr4, { text: 'Updated:', class: 'label' });
+      createAndAppend('td', tr4, { text: new Date(repoInfo.updated_at).toLocaleString() });
     });
-    return elem;
+  }
+
+  function renderContributors(repoName, ulContributorInfo) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repoName}/contributors`;
```
remarcmij: Do not contruct this url yourself. Use the `contributors_url` as mentioned in the assignment. GitHub may choose to alter the internal format of this url, but will not likely change the name of the property as this is the documented way to obtain the contributors information.

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
+      createAndAppend('td', tr1);
+      createAndAppend('a', tr1.lastChild, {
+        href: repoInfo.html_url,
+        target: '_blank',
+        text: repoInfo.name,
+      });
+
+      const tr2 = createAndAppend('tr', tbody);
+      createAndAppend('td', tr2, { text: 'Description:', class: 'label' });
+      createAndAppend('td', tr2, { text: `${repoInfo.description}` });
```
remarcmij: There is not need to use a template literal for something that is already a string (or can be automatically converted to string):

```js
createAndAppend('td', tr2, { text: repoInfo.description });
```

See line 47 where you already used did this.

Check other places in your code where you have unneeded template literals.

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
-      if (key === 'text') {
-        elem.textContent = value;
-      } else {
-        elem.setAttribute(key, value);
-      }
+  function renderRepositories(repo, ulRepoInfo) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repo}`;
+    fetchJSON(url, (err, repoInfo) => {
+      ulRepoInfo.innerHTML = '';
+      const li = createAndAppend('li', ulRepoInfo, { id: 'li1' });
```
remarcmij: An `ul` with a single `li` child containing a single `table` does not make sense. You do not need a list (of one element) here. This is the HTML you generate:

```html
<div class="info-section-div">
  <div>
    <ul class="li-container">
      <li id="li1">
        <table>
          <tbody>
            <tr>
              <td class="label">Repository:</td>
            </tr>
            <tr>
              <td class="label">Description:</td>
            </tr>
            <tr>
              <td class="label">Forks:</td>
              <td>9</td>
            </tr>
            <tr>
              <td class="label">Updated:</td>
            </tr>
          </tbody>
        </table>
      </li>
    </ul>
  </div>
  <div>
    <ul class="li-container">
      <li class="li2">
        <img src="https://avatars0.githubusercontent.com/u/15912395?v=4" alt="gijscor photo" />
        <h1><a href="https://github.com/gijscor" target="_blank">gijscor</a></h1>
        <p>47</p>
      </li>
      <li class="li2">
        <img src="https://avatars1.githubusercontent.com/u/37488847?v=4" alt="wouterkleijn photo" />
        <h1><a href="https://github.com/wouterkleijn" target="_blank">wouterkleijn</a></h1>
        <p>11</p>
      </li>
    </ul>
  </div>
</div>
```

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
     });
-    return elem;
+  }
+
+  function renderContributors(repoName, ulContributorInfo) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repoName}/contributors`;
+    fetchJSON(url, (err, contributors) => {
+      ulContributorInfo.innerHTML = '';
+      contributors.forEach(contributor => {
+        const listItems = createAndAppend('li', ulContributorInfo, { class: 'li2' });
```
remarcmij: 1. You are creating a single list item at a time. So the singular form `listItem` should be used for this variable.
2. Avoid (CSS or otherwise) names with numeric suffixes. Take a little more time to come up with a good name.
3. It would be nice if you could click anywhere on the list item to go to the contributors page. You can achieve that by wrapping all inner content of each list item into an `a` tag with the relevant url for the `href`. This, in place of the `a` you have used for the contributor login name.

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
-      }
+  function renderRepositories(repo, ulRepoInfo) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repo}`;
+    fetchJSON(url, (err, repoInfo) => {
+      ulRepoInfo.innerHTML = '';
+      const li = createAndAppend('li', ulRepoInfo, { id: 'li1' });
+      const table = createAndAppend('table', li);
+      const tbody = createAndAppend('tbody', table);
+
+      const tr1 = createAndAppend('tr', tbody);
```
remarcmij: You are repeating yourself four times here: create a `tr` with two `td1 children. Can you make your code DRYer by creating a function, say `addRow`, that you could call like this:

```js
```

For the first row you would need to do something special as that one is different from the rest. For instance:

```js
createAndAppend('a', firstRow.lastChild, { href: ..., text: ... });
```

Update: while I was typing this you pushed some code that addresses this. Still, you may want to consider the solution for the first row.

homework/style.css
```diff
@@ -1,3 +1,137 @@
+html {
```
remarcmij: You still need to work on responsiveness.

image

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
+    const contributorInfoSection = createAndAppend('div', infoSection);
+
+    createAndAppend('h1', header, { text: 'HYF Repositories' });
+    const select = createAndAppend('select', header);
+    const ulRepoInfo = createAndAppend('ul', repoInfoSection, { class: 'li-container' });
+    const ulContributorInfo = createAndAppend('ul', contributorInfoSection, {
+      class: 'li-container',
+    });
+
+    fetchJSON(url, (error, data) => {
```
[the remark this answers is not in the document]
Mehmet-Ergorgec: I thought that later I possibly can get information from another API to the website. So I used **data** for a general purpose. I thought I could use this fetchJSON. **But I understand what you mean**. When I want to add another fetchJSON , first I should create another fetchJSON so I can easily name that the  data *what is inside of it*. So the codes can get bigger and bigger after a while I can try to to remember what the data is. Thank you

homework/index.js
```diff
@@ -16,31 +30,97 @@
...
+      if (error) {
+        createAndAppend('div', root, { text: error.message, class: 'alert-error' });
         return;
       }
```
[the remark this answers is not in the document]
Mehmet-Ergorgec: Yes I guess your videos effected me. Now It is clear. Thank you

homework/index.js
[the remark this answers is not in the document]
remarcmij: You must make a distinction between the `fetchJSON()` function definition and the places where you **_call_** it. The function definition can deal with any type of JSON data. Therefore the parameter name `data` in the function definition is fine: the `fetchJSON()` function does not know or care about the nature of the data, other than that is in the JSON format.

In the places where you **_call_** `fetchJSON()` you _do_ know and care about the nature of the data. So there a descriptive name for the data is appropriate.

homework/index.js
Mehmet-Ergorgec: 1. Yes single item at a time. But some repositories have more than one contributor. So should it be plural?
3. I guess I misunderstood hw. Yes I am gonna correct it.

homework/index.js
Mehmet-Ergorgec: Thanks for the tip. I have tried to do it so I had created addHref func .Now no need that function.

homework/index.js
```diff
@@ -16,31 +42,86 @@
...
+        // const repo = repositories[select.value];
+        renderRepositories(repo, repoInfo);
+        renderContributors(repo, contributorInfo);
+      });
     });
   }
-
   const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
-  window.onload = () => main(HYF_REPOS_URL);
+  window.onload = main(HYF_REPOS_URL);
```
remarcmij: I overlooked this change that you made which actually introduces a subtle bug. Now, the main function is directly called in 126 and its return value (`undefined`) is assigned to the `window.onload` property. You will not have noticed any issues, but if somebody decides to move script tag that loads this JavaScript file from the `body` to the `head` tag of the `index.htm` file your application will no longer run.

Try it, and see if you can explain what is happening.

homework/index.js
```diff
@@ -16,31 +42,86 @@
...
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+      repositories.sort((a, b) => a.name.localeCompare(b.name));
+      createOptionElements(repositories, select);
+      const repo = repositories[select.value];
+      renderRepositories(repo, repoInfo);
+      renderContributors(repo, contributorInfo);
+
+      select.addEventListener('change', () => {
+        // const repo = repositories[select.value];
```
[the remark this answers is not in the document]
Mehmet-Ergorgec: Thank you again your kind review. i worked different files. I think I tested another file. Because of that reason I skipped original file. I will be more careful than that. Sorry for that.

homework/index.js
Mehmet-Ergorgec: Luckily or unluckily I have made this mistake. So I watched and read about window.onload event after your review.  You are right. You made me thinking about that problem.  While I was writing window.onload it was automatic for me .But now I hope I understood what it means. So:
**Scenario 1: js inside the head tag. in Js file  ` window.onload = main(HYF_REPOS_URL)`;**
I got an error when I try to use js in head tag. Because Js reads files from top to bottom. So when it come  to script tag Js will not understand the code. Because `window.onload` property is actually null.
**Scenario 2: js in different file. but declared in body tag. `window.onload = main(HYF_REPOS_URL)`;**
This was my code. I did not get an error. Because  when js was reading body tag since my script tags were at the end of the body tag ,  Js understood what the main function means.
 Am I right?

remarcmij: No, it is not quite like that. As soon as the browser encounters your script tag it will load the JavaScript file and will immediately start executing all executable code. Function _definitions_ are by themselves not directly executable, but function _calls_ are.

This is the erroneous code:

```js
window.onload = main(HYF_REPOS_URL);
```

You immediately call function `main` and assign its return value, `undefined`, to the `window.onload` property. Your `main` function starts executing immediately and tries to get the `root` div using `getElementById`. If your script tag is in the `head` element, the body of the document will not have been processed yet, therefore there is no such div to be found and `getElementById` will return `null`. Hence your code starts to fail. When the window 'load' event fires, nothing happens because the event handler is `undefined`.

If you place the script tag at the end of the body, then the root div will already have been created and everything seems to work fine, despite the `window.onload` event handler being `undefined`.

This is the correct code:

```js
 window.onload = () => main(HYF_REPOS_URL);
```

You have now _not_ called `main()` directly, but instead assigned an arrow function to the `window.onload` property that is to be called when the document is fully loaded (including the `body` with your root div). Now, it no longer matters where you place the script tag.

Mehmet-Ergorgec:
Thank you for all. now It is clear.I learned a lot from this homework

PR #300, opened 2019-07-26
Js3 week1

elicarcar:


homework/index.js
```diff
@@ -30,14 +36,102 @@
...
+
     fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        renderError(err);
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+
```
remarcmij: It is best to sort the whole array of repositories here:

```js
```

homework/index.js
```diff
@@ -30,14 +36,102 @@
...
 
+  function createOptions(repoNames, select) {
+    createAndAppend('option', select, { text: 'Select module', disabled: 'disabled' });
+    repoNames.sort((a, b) => a.name.localeCompare(b.name));
+    repoNames.forEach((repository, i) => {
+      createAndAppend('option', select, {
+        text: repository.name,
+        value: repository.name,
+      });
+    });
```
remarcmij: As hinted in the assignment, use the array index as the value attribute of the `option` element (it will also get rid of an ESLint error about an unused `i` variable: it is now used):

```js
  text: repository.name,
  value: i,
});
```

This allows you to directly access the relevant repository information in your event listener for the `<select>` element in function `main()` with:

```js
  const repository = repositories[select.value];
  renderRepoInformation(repository, ul);
  renderContributorsInformation(repository, contrList);
});
```

No need for an extra fetch!

homework/index.js
```diff
@@ -30,14 +36,102 @@
     return elem;
   }
 
+  function createOptions(repoNames, select) {
+    createAndAppend('option', select, { text: 'Select module', disabled: 'disabled' });
+    repoNames.sort((a, b) => a.name.localeCompare(b.name));
+    repoNames.forEach((repository, i) => {
```
remarcmij: These are not just names (`repoNames`) but `repositories` (objects). The name should reflect that.

homework/index.js
```diff
@@ -30,14 +36,102 @@
     return elem;
   }
 
+  function createOptions(repoNames, select) {
+    createAndAppend('option', select, { text: 'Select module', disabled: 'disabled' });
```
remarcmij: The assignment specified that you should display the details of the first repo in the (sorted) list at application startup. Therefore you should not add this placeholder option at the start of the list. (This in contrast to what we did in class).

At application startup the select element should display the name of the first repo (at present, `alumni`) and render the details of that repo in the page.

homework/index.html
```diff
@@ -1,23 +1,26 @@
...
-<body>
-  <div id="root"></div>
-  <script src="./index.js"></script>
-</body>
-
-</html>
\ No newline at end of file
+  <body>
+    <div id="root"></div>
+    <div id="bodyDiv"></div>
```
remarcmij: Do not add this extra div. Everything should be placed inside the root div. Hence the name `root`.

homework/index.js
```diff
@@ -30,14 +36,102 @@
...
+
+  function renderRepoInformation(repo, ul) {
+    const url = `https://api.github.com/repos/HackYourFuture/${repo}`;
+    fetchJSON(url, (err, repoInfos) => {
+      if (err) {
+        renderError(err);
+        return;
+      }
+
+      ul.innerHTML = ' ';
```
remarcmij: Is does not make sense to create a list with a single list item. This is the HTML you generate:

```html
<ul class="repoInfoList">
  <li>
    <table class="infoTable">
      <tbody>
        <tr>
          <td class="explanations">Repository:</td>
          <td class="list-item">
            <a href="https://github.com/HackYourFuture/alumni" target="blank">alumni</a>
          </td>
        </tr>
        <tr>
          <td class="explanations">Description:</td>
        </tr>
        <tr>
          <td class="explanations">Forks:</td>
          <td>9</td>
        </tr>
        <tr>
          <td class="explanations">Updated:</td>
          <td>2019-06-21T11:38:21Z</td>
        </tr>
      </tbody>
    </table>
  </li>
</ul>
```

It would make more sense to create a `div` or `section` to serve as a container for the `table`.

homework/index.js
```diff
@@ -30,14 +36,102 @@
...
+      const li = createAndAppend('li', ul);
+      const table = createAndAppend('table', li, { class: 'infoTable' });
```
remarcmij: CSS names should be in kebab-case by convention: `info-table`. This holds true for several places in your code.

homework/index.js
```diff
@@ -30,14 +36,102 @@
...
+      const tbody = createAndAppend('tbody', table);
+      const repoName = createAndAppend('tr', tbody);
```
remarcmij: You are repeating yourself here four times: create a `tr` with two `td` children. Can you make your code DRYer by creating a function to do this, e.g. a function named `addRow`? You could call it like this:

```js
```

For the first row containing the `a` tag, you would need to do something special, for instance:

```js
createAndAppend('a', firstRow.lastChild, { href: ...., text: ... });
```

homework/index.js
```diff
@@ -30,14 +36,102 @@
...
+      createAndAppend('td', repoName, { text: 'Repository:', class: 'explanations' });
+      const repoNameTd = createAndAppend('td', repoName, {
+        class: 'list-item',
+      });
+      createAndAppend('a', repoNameTd, {
+        href: `${repoInfos.html_url}`,
```
remarcmij: There is no need to use template literals here if they only contain something that is a string already (or is converted to string automatically):

```js
href: repoInfos.html_url,
````

There are several places in your code where you use unneeded template literals.

File: homework/index.js [hunk 10]
```diff
@@ -30,14 +36,102 @@
...
+      createAndAppend('td', fork, { text: 'Forks:', class: 'explanations' });
+      createAndAppend('td', fork, { text: `${repoInfos.forks}` });
+      const updateTime = createAndAppend('tr', tbody);
+      createAndAppend('td', updateTime, { text: 'Updated:', class: 'explanations' });
+      createAndAppend('td', updateTime, { text: `${repoInfos.updated_at}` });
+    });
+  }
+
+  function renderContributorsInformation(subjects, ul) {
+    const url = `https://api.github.com/repos/HackYourFuture/${subjects}/contributors`;
```
remarcmij: Don't construct a url like this yourself, because you are making assumptions about the internal format of the url. Use the `contributors_url` property of the repository object itself because that is guaranteed to be stable. GitHub may choose to change the internal format of the url and if they do it will break your code.

homework/index.js
```diff
@@ -30,14 +36,102 @@
...
+    fetchJSON(url, (error, contributorDetails) => {
+      if (error) {
+        renderError(error);
+        return;
+      }
+
+      ul.innerHTML = ' ';
+      createAndAppend('h4', ul, { text: 'Contributors' });
+      contributorDetails.forEach(contributor => {
+        const li = createAndAppend('li', ul);
```
remarcmij: It would be nice if you could click anywhere in the list item (not just on the login name) to go to the GtiHub page of the contributor. You can achieve that by wrapping the complete inner content of the `li` into an `a` tag with the correct `href` value. You then should not use an `a` tag anymore for the login name, but just a div or a span.

homework/index.js
```diff
@@ -16,6 +17,11 @@
     xhr.send();
   }
 
+  function renderError(error) {
+    const root = document.getElementById('root');
+    createAndAppend('h1', root, { text: error.message });
```
remarcmij: You have an ESLint error because you are referencing the `createAndAppend` function before it is defined. The solution is simple: just move this function definition after that of  `createAndAppend`.

homework/index.js
```diff
@@ -30,14 +30,105 @@
...
+      href: `${repositories.html_url}`,
+      target: 'blank',
+      text: `${repositories.name}`,
+    });
+    addRow(tbody, 'Description:', repositories.description, 'explanations');
+    addRow(tbody, 'Fork:', repositories.forks, 'explanations');
+    addRow(tbody, 'Updated:', `${repositories.updated_at}`.substring(0, 10), 'explanations');
+  }
+
+  function renderContributorsInformation(repositories, container) {
```
remarcmij: This function gets a single `repository` as its first argument. Your name `repositories` (implying plural or, in other words, an array) is misleading, to say the least or, more bluntly stated, completely incorrect.

homework/index.js
```diff
@@ -30,14 +30,105 @@
...
+    const contributors = repositories.contributors_url;
+    const url = contributors;
+    fetchJSON(url, (error, contributorDetails) => {
```
remarcmij: Lines 70 and 71 seem to unnecessarily juggle with the url value. Why not just get rid of them and change line 72 to:

```js
fetchJSON(repository.contributors_url, (error, contributors) => {
```

homework/index.js
```diff
@@ -30,14 +30,105 @@
...
+  }
+
+  function addRow(tbody, labelText, value, str) {
+    const tr = createAndAppend('tr', tbody);
+    createAndAppend('td', tr, { text: `${labelText}`, class: str });
+    createAndAppend('td', tr, { text: value });
+    return tr;
+  }
+
+  function renderRepoInformation(repositories, container) {
```
remarcmij: See my comment elsewhere on the parameter name `repositories`.

homework/index.js
```diff
@@ -30,14 +30,105 @@
...
+    const bodyDiv = createAndAppend('div', root, { id: 'body-div' });
+    const section = createAndAppend('section', bodyDiv, { class: 'repo-info-list' });
+    const contrDiv = createAndAppend('div', bodyDiv, {
+      class: 'contributors-div',
+      id: 'cont-ulist',
+    });
+
     fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
+      repositories.sort((a, b) => a.name.localeCompare(b.name));
```
remarcmij: If there is an error this `sort` will produce a runtime error in the console, as there is no data to sort. You better wait with sorting until you have established that there was no error.

homework/index.js
```diff
@@ -30,14 +30,105 @@
...
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        renderError(err);
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+
+      function showFirstRepo() {
```
remarcmij: This immediately invoked function looks a lot like what I have seen in the homework from @gulyener. As is it stands, this function does not add much value over simply executing lines 120-121 directly, e.g.:

```js
  if (err) {
    renderError(err);
    return;
  }

  renderRepoInformation(repositories[0], section);
  renderContributorsInformation(repositories[0], contrDiv);

  createOptions(repositories, select);
```

If you want to use a function, make it reusable like this:

```js
function showRepo(repo, repoContainer, contributorContainer) {
  renderRepoInformation(repo, repoContainer);
  renderContributorsInformation(repo, contributorContainer);
}

function main(url) {
   ...
    if (err) {
      renderError(err);
      return;
    }
    createOptions(repositories, select);

    showRepo(repositories[0], section, contrDiv);

      const repository = repositories[select.value];
      showRepo(repository, section, contrDiv);
    });
  });
}
```

elicarcar: So true, I don't know why I felt the need of using a non-reusable function. I've just realized that I don't need a specific function for showing the first repository, so I am just going to call my two functions instead of calling them inside of a new function. Is it alright?

PR #301, opened 2019-07-26
JS3 week1 homework submission

hnnhack:
best regards,..

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -32,6 +32,15 @@
     return elem;
   }
 
+  function renderError(error) {
+    const container = document.getElementsByClassName('container');
+    container.innerHTML = '';
+    createAndAppend('div', container, {
+      text: error.message,
+      class: 'alert',
+    });
+  }
+
   function showRepositoryInfo(repository) {
     const leftHand = document.getElementById('left-hand');
     leftHand.innerHTML = '';
@@ -120,9 +129,8 @@
     });
     const select = createAndAppend('select', navElement);
     fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        renderError(error);
         return;
       }
       createAndAppend('option', select, { text: 'Select repository', disabled: 'disabled' });
--- a/homework/index.js
+++ b/homework/index.js
@@ -33,11 +33,11 @@
   }
 
   function renderError(error) {
-    const container = document.getElementsByClassName('container');
+    const container = document.getElementById('container');
     container.innerHTML = '';
     createAndAppend('div', container, {
       text: error.message,
-      class: 'alert',
+      class: 'error',
     });
   }
 
@@ -126,10 +126,11 @@
     });
     const divElement = createAndAppend('div', rootElement, {
       class: 'container',
+      id: 'container',
     });
     const select = createAndAppend('select', navElement);
-    fetchJSON(url, (err, repositories) => {
-      if (err) {
+    fetchJSON(url, (error, repositories) => {
+      if (error) {
         renderError(error);
         return;
       }
--- a/homework/index.js
+++ b/homework/index.js
@@ -37,41 +37,35 @@
     container.innerHTML = '';
     createAndAppend('div', container, {
       text: error.message,
-      class: 'error',
+      class: 'alert-error',
     });
+  }
+
+  function addRow(table, text, value) {
+    const tableRow = createAndAppend('tr', table);
+    createAndAppend('th', tableRow, {
+      text: text,
+      class: 'table-header',
+    });
+    createAndAppend('td', tableRow, {
+      text: value,
+    });
+    return tableRow;
   }
 
   function showRepositoryInfo(repository) {
     const leftHand = document.getElementById('left-hand');
     leftHand.innerHTML = '';
     const tableElement = createAndAppend('table', leftHand);
-    const rowForName = createAndAppend('tr', tableElement);
-    createAndAppend('th', rowForName, {
-      text: 'Repository:',
-    });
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -30,17 +32,134 @@
...
-    fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+  function renderError(error) {
+    const container = document.getElementById('container');
+    container.innerHTML = '';
+    createAndAppend('div', container, {
+      text: error.message,
+      class: 'error',
```
remarcmij: The assignment asked for you to use the class `alert-error` (step 6 of the functional requirements https://github.com/HackYourFuture/JavaScript3/pull/296/files#r307708657)

My unit tests that I use to help me review the homework expects this. In general, you should meticulously follow the functional requirements. It is what the customer (in this case, me) expects.

homework/index.js
```diff
@@ -30,17 +32,134 @@
...
+      class: 'container',
+      id: 'container',
+    });
+    const select = createAndAppend('select', navElement);
+    fetchJSON(url, (error, repositories) => {
+      if (error) {
+        renderError(error);
+        return;
+      }
+      createAndAppend('option', select, { text: 'Select repository', disabled: 'disabled' });
```
remarcmij: The assignment specifies that, at application startup, the details for the first repo in the (sorted) list should be displayed. At present, this happens to be the **alumni** repo. This is in contrast with the other approach we took in the class where we used a placeholder option as you have here, with no details rendered in the page at startup.

Therefore, you should not add this placeholder option at the top of the list. At startup the select box should show the name of the first repo in the list and the page should show the details for that repo.

homework/index.js
```diff
@@ -30,17 +32,134 @@
...
+        });
+        createAndAppend('span', liElement, {
+          class: 'login',
+          text: elem.login,
+        });
+        createAndAppend('span', liElement, {
+          class: 'counter',
+          text: elem.contributions,
+        });
+        liElement.onclick = () => {
```
remarcmij: Responding to the click event works fine for users with normal vision, but visually impaired users need to able to navigate solely with the keyboard. They would need to be able to press Enter and go to the contributor's page while the list item has the focus. The best way to achieve is to use a regular `a` tag that wraps the entire inner content of the `li` element and has an `href` with the correct url. The default behaviour of the `a` tag meets all the requirements for ARIA compliance (part of the tab order, clickable and can be activated with the Enter key).

homework/index.js
```diff
@@ -30,17 +32,134 @@
...
+      text: error.message,
+      class: 'error',
+    });
+  }
+
+  function showRepositoryInfo(repository) {
+    const leftHand = document.getElementById('left-hand');
+    leftHand.innerHTML = '';
+    const tableElement = createAndAppend('table', leftHand);
+    const rowForName = createAndAppend('tr', tableElement);
```
remarcmij: You have repeated code here: create a `tr` with two `td` children. You make your code DRYer, by creating a function, say `addRow`, that you could call like this:

```js
```

For the first row (containing the `a` tag) you could do a 'fix up' after calling `addRow`:

```js
const firstRow = addRow(tbody, 'Repository', '');
createAndAppend('a', firstRow.lastChild, { href: ..., text: ... });
```

hnnhack: Hello Jim Cramer!
Thanks for your review and notes. I want summarize your notes, if it's true please confirm it. 
1. I have to change class: error; to class: alert-error; 
2. When application starts automatically first repository will be selected and details will be shown on the page.
3. On the eventlistener, i should use onselect instead of onclick.
4. I have to create and use addRow() function for creating table rows.
Is that correct?

homework/index.js
```diff
@@ -37,41 +37,35 @@
...
       text: error.message,
-      class: 'error',
+      class: 'alert-error',
     });
   }
 
+  function addRow(table, text, value) {
+    const tableRow = createAndAppend('tr', table);
+    createAndAppend('th', tableRow, {
+      text: text,
```
remarcmij: This produces an ESLint error and therefore a Travis error, but it is not your fault. In modern (ES6) JavaScript you can use 'short-hand' notation in cases where an object property name is the same as the variable name you assign to it. This short-hand notation is generally preferred and the (Airbnb) ESLint rules have been set up to complain about it if you don't use it. It is a **_style_** error, not a JavaScript error. 

With the short-hand notation, your code would look like this:

```js
createAndAppend('th', tableRow, {
  text,
  class: 'table-header',
});
```

homework/index.js
```diff
@@ -37,41 +37,35 @@
...
-    createAndAppend('th', rowForUpdated, {
-      text: 'Updated:',
-    });
-    createAndAppend('td', rowForUpdated, {
-      text: new Date(repository.updated_at).toLocaleString('en-GB'),
-    });
+    addRow(tableElement, 'Description:', repository.description);
+    addRow(tableElement, 'Forks:', repository.forks_count);
+    addRow(tableElement, 'Updated:', new Date(repository.updated_at).toLocaleString('en-GB'));
   }
```
remarcmij: DRY! 👍

homework/index.js
```diff
@@ -96,21 +90,23 @@
         const liElement = createAndAppend('li', ulElement, {
```
remarcmij: To be exact, what you get back from the GitHub API is an array `contributors`. And an element of that array is a `contributor`. Therefore, the most descriptive and intention-revealing form for your `.forEach` would be:

```js
```

Names like `element` are generic and do not reveal much about what type of data they contain.

hnnhack:
Hi Jim Cramer!
Thanks for you reviews, notes, suggestions and approves.

PR #302, opened 2019-07-27
mehmet bozdemir js3week1 homework

mehmet-bozdemir:


homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+  fetchJSON(url, (err, repositories) => {
+    const root = document.getElementById('root');
+    if (err) {
+      createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+      return;
+    }
+    const repNames = [];
+    createAndAppend('select', root, { id: 'main-select' });
+    createAndAppend('div', root, { class: 'portion-parent', id: 'parent' });
+    const parent = document.getElementById('parent');
```
remarcmij: You may have missed that the `createAndAppend` function actually returns a reference to the element it created (`return elem;` as its last statement). This means that can assign the reference to a variable which you can use directly, without having to use `getElementById()`. It also removes the need to define an `id` attribute. So lines 41 and 42 could be replaced with:

```js
const parent = createAndAppend('div', root, { class: 'portion-parent' });
```

This same principle applies to several other places in your code. Make sure that you also remove any `id` attributes that are no longer needed: _less is more_.

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+    createAndAppend('div', parent, { class: 'left-portion', id: 'lft' });
```
remarcmij: As I said, you don't need this `id` attribute, but the point I want to make here is that the names you choose should not be cryptic. What do you gain from naming something `lft` over `left` or `rt` over `right`? You should try and express yourself as clearly as possible. People reading your code will be thankful.

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-        return;
+    const leftPortion = document.getElementById('lft');
+    const rightPortion = document.getElementById('rt');
+    for (let i = 0; i < repositories.length; i++) {
+      const rep = repositories[i];
+      repNames.push(rep.name);
+    }
```
remarcmij: There is no need to create this separate `repNames` array. Just sort the `repositories` array itself on the `name` property, after fetching it.

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+    const selectItem = document.getElementById('main-select');
+
+    repNames.sort();
```
remarcmij: Use `localeCompare` to sort the `repositories` array on the `name` property. This simple sort that you do here on just the names places all repositories that start with an uppercase letter above all repositories that start with a lowercase letter. People will expect the list to be sorted without taking case into account. `localeCompare` does this for you, as well as taking into accounted accented letter, as in French.

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+    for (let i = 0; i < repNames.length; i++) {
```
remarcmij: Instead of a `for` loop, you could use `.forEach()` here.

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+      createAndAppend('option', selectItem, {
+        value: i,
+        text: repNames[i],
+      });
+    }
+    selectItem.selectedIndex = 0;
+    selectItem.addEventListener('change', () => {
+      const reponame = selectItem.options[selectItem.selectedIndex].text;
```
remarcmij: Instead of trying to get the repository name you can get an object with the full details of the selected repository by using `repositories[selectItem.selectedIndex]` or  `repositories[selectItem.value]` (because we have set the `value` attribute equal to the array index). Of course, this assume that the whole `repositories` array was sort by the `name` property and used in that sorted form to create the `<option>` elements.

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+      for (let i = 0; i < repositories.length; i++) {
+        if (repositories[i].name === reponame) {
+          const repo = repositories[i];
```
remarcmij: Because you can access the selected repository object directly by using the index there is no need to loop over all array elements to find the relevant element. You can remove line 64 and 65 and the corresponding closing brace and change line 66 into:

```js
```

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+          leftPortion.innerHTML = '';
```
remarcmij: The HTML structure you generate leaves room for improvement. If it gives you difficulty it will help to draw up a 'plan' on what HTML you want to generate, and then devise the JavaScript that is needed to achieve that. If you try it directly in JavaScript without a plan you have two problems to deal with at the same time. Therefore: _divide and conquer_.

Here is what you generate now (using the `alumni` repo as example):

```html
<div id="root">
  <select id="main-select">
    <option value="0">AngularJS</option>
    <!-- - more options -->
    <option value="46">ukulele</option>
  </select>
  <div class="portion-parent" id="parent">
    <div class="left-portion" id="lft">
      <p>alumni</p>
      <p>Alumni Guides, everything you need to land a job</p>
      <p>9</p>
      <p>2019-06-21T11:38:21Z</p>
    </div>
    <div class="right-portion" id="rt">
      <div class="rightParent" id="contList0">
        <img src="https://avatars0.githubusercontent.com/u/15912395?v=4" />
        <p>gijscor</p>
        <p>47</p>
      </div>
      <div class="rightParent" id="contList1">
        <img src="https://avatars1.githubusercontent.com/u/37488847?v=4" />
        <p>wouterkleijn</p>
        <p>11</p>
      </div>
    </div>
  </div>
</div>
```

This is what I generate in my solution to this homework:

```html
<div id="root">
  <header class="header">
    <p>HYF Repositories</p>
    <select class="repo-selector">
      <option value="0">alumni</option>
      <!-- - more options -->
      <option value="46">ukulele</option>
    </select>
  </header>
  <div id="container">
    <div class="left-div whiteframe">
      <table>
        <tbody>
          <tr>
            <td class="label">Repository :</td>
          </tr>
          <tr>
            <td class="label">Description :</td>
          </tr>
          <tr>
            <td class="label">Forks :</td>
            <td>9</td>
          </tr>
          <tr>
            <td class="label">Updated :</td>
          </tr>
        </tbody>
      </table>
    </div>
    <div class="right-div whiteframe">
      <p class="contributor-header">Contributions</p>
      <ul class="contributor-list">
        <li class="contributor-item" aria-label="gijscor" tabindex="0">
          <img
            height="48"
            class="contributor-avatar"
          />
          <div class="contributor-data">
            <div>gijscor</div>
            <div class="contributor-badge">47</div>
          </div>
        </li>
        <li class="contributor-item" aria-label="wouterkleijn" tabindex="0">
          <img
            height="48"
            class="contributor-avatar"
          />
          <div class="contributor-data">
            <div>wouterkleijn</div>
            <div class="contributor-badge">11</div>
          </div>
        </li>
      </ul>
    </div>
  </div>
</div>
```

Can you write JavaScript code, using `createAndAppend()` that creates something similar to what I have?

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+          createAndAppend('p', leftPortion, { text: repo.name });
+          createAndAppend('p', leftPortion, { text: repo.description });
+          createAndAppend('p', leftPortion, { text: repo.forks });
+          createAndAppend('p', leftPortion, { text: repo.updated_at });
+          const contUrl = repo.contributors_url;
+          fetchJSON(contUrl, (cerr, contList) => {
```
remarcmij: You are not handling errors here.

homework/index.js
```diff
@@ -1,46 +1,99 @@
...
+        }
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
     });
-  }
 
-  const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
-  window.onload = () => main(HYF_REPOS_URL);
+    // createAndAppend('pre', root, {text: JSON.stringify(repositories, null, 2)});
+  });
```
remarcmij: The assignment specifies that, at application startup, you should render the details for the first repo in the list (the one that is displayed in the select box). This does not happen in your case. Note that the `change` event is only fired if the user actively changes the selection. Therefore, at application startup, you must manually run the same code that the event listener runs. You now have a big arrow function as the event handler which makes it difficult to run the same code from another place. There you should try and create a separate functoin that contains this code that you can call both from the even t listener and from another place to render the first repo at startup.

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+        });
+        createAndAppend('div', tr, { text: name, class: 'col-30 description' });
+        const td3 = createAndAppend('div', tr, { class: 'col-30' });
+        createAndAppend('label', td3, { text: contributionsCount });
+      }
+    });
+  }
+  function bindRepositoryDetails(selectedRepository) {
+    const { description } = selectedRepository.description;
+    const { forks } = selectedRepository.forks;
```
remarcmij: You have tried 'object destructuring' here, but it doesn't quite work. You can see that the `description` and `forks` values remain blank on the page. You must either use regular JavaScript like this:

```js
const description = selectedRepository.description;
const forks = selectedRepository.forks;
```

Or object destructuring like this:

```js
const { description } = selectedRepository;
const { forks } = selectedRepository;
```

The nice thing with object destructuring is that you can initialize multiple variables in a single line like this:

```js
const { description, forks } = selectedRepository;
```

You have now defined two constants, `description` and `forks` initialized with the values of the corresponding object properties.

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+        });
+        return;
+      }
+
+      const div = createAndAppend('div', contributorsDiv);
+      createAndAppend('h4', div, { text: 'Contributions' });
+      createAndAppend('hr', div);
+      const table = createAndAppend('div', div, { class: 'grid' });
+      const contributors = response;
+      for (let index = 0; index < contributors.length; index++) {
```
remarcmij: While an old-style `for` loop will work fine, many developers, including myself, would prefer a `.forEach()` if you just want to loop through the elements of an array.

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+          text: error.message,
+          class: 'alert-error',
+        });
+        return;
+      }
+
+      const div = createAndAppend('div', contributorsDiv);
+      createAndAppend('h4', div, { text: 'Contributions' });
+      createAndAppend('hr', div);
+      const table = createAndAppend('div', div, { class: 'grid' });
```
remarcmij: The contributors list is best expressed as a list (for ARIA-compliance). Visually impaired users that use a screen reader will rely on your HTML to follow the semantic rules, e.g. a list must be presented by an HTML list, in this case a `ul`. Each item in the list should be an `li`.

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+  }
+  function bindRepositoryDetails(selectedRepository) {
+    const { description } = selectedRepository.description;
+    const { forks } = selectedRepository.forks;
+    const repositoryName = selectedRepository.name;
+    const updatedOn = new Date(selectedRepository.updated_at).toLocaleString();
+    const repoURL = selectedRepository.html_url;
+
+    const infoContainerDiv = document.getElementById('repo-info-container');
+    const repoInfodiv = createAndAppend('div', infoContainerDiv, {
```
remarcmij: You can avoid a spelling error being flagged here by using camelCase also for the `div` part: `repoInfoDiv`.

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+    const div = createAndAppend('div', topNavDiv, { class: 'header-container' });
+    const div1 = createAndAppend('div', div, { class: 'col-50' });
+    createAndAppend('a', div1, { text: 'HYF Repositories', href: '#' });
+    const repoDiv = createAndAppend('div', div, { class: 'col-50' });
+    const repoDropDown = createAndAppend('select', repoDiv, {
+      id: 'ddlRepository',
+      class: 'select-css',
+    });
+    createAndAppend('option', repoDropDown, { text: 'Select Repository' });
+    repositories.sort((a, b) => {
```
remarcmij: I suggested the use of `localeCompare` in my previous feedback. The problem of your custom implementation is that it may not handle language specific special characters such as accented letter in French etc. And I know that for instance Arabic doesn't have uppercase of lowercase.

The `.localeCompare()` method use the language settings from the user's computer configuration. For instance, on a computer configured to use the French language and regional settings, `localeCompare` will sort according to the French dictionary order.

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+      if (x < y) {
+        return -1;
+      }
+      if (x > y) {
+        return 1;
+      }
+      return 0;
+    });
+
+    for (let index = 0; index < repositories.length; index++) {
```
remarcmij: You can use `.forEach()`:

```js
repositories.forEach((repository, index) => {
```

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+    const topNavDiv = createAndAppend('div', parentElement, { class: 'topnav' });
+    const div = createAndAppend('div', topNavDiv, { class: 'header-container' });
+    const div1 = createAndAppend('div', div, { class: 'col-50' });
+    createAndAppend('a', div1, { text: 'HYF Repositories', href: '#' });
+    const repoDiv = createAndAppend('div', div, { class: 'col-50' });
+    const repoDropDown = createAndAppend('select', repoDiv, {
+      id: 'ddlRepository',
+      class: 'select-css',
+    });
+    createAndAppend('option', repoDropDown, { text: 'Select Repository' });
```
remarcmij: The homework assignment states that at application startup the details for the first repo should be displayed. You currently are not displaying details at startup.

So, the select box should show the name of the first repo in the (sorted) list, in this `alumni` and the page should show the details of that repo.

Note that this is different from what we did in class.

homework/index.js
```diff
@@ -1,6 +1,8 @@
 'use strict';
 
 {
+  let repositoryList = [];
```
remarcmij: Don't use global variables like this. It makes it difficult to find out where in your code references are made to it.

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+  }
+
+  function repositoryChanged(event) {
+    const infoElement = document.getElementById('repo-info-container');
+    infoElement.innerHTML = '';
+    const repositoryIndex = parseInt(event.target.selectedOptions[0].value, 10);
+    const selectedRepo = repositoryList[repositoryIndex];
+    bindRepositoryDetails(selectedRepo);
+    bindContributions(selectedRepo.contributors_url);
+  }
```
remarcmij: You can pass the selected repo directly as an argument:

```js
  function repositoryChanged(selectedRepo) {
    const infoElement = document.getElementById('repo-info-container');
    infoElement.innerHTML = '';
    // const repositoryIndex = parseInt(event.target.selectedOptions[0].value, 10);
    // const selectedRepo = repositoryList[repositoryIndex];
    bindRepositoryDetails(selectedRepo);
    bindContributions(selectedRepo.contributors_url);
  }
```

See how that works on line 155.

homework/index.js
```diff
@@ -30,14 +32,145 @@
...
+    });
+
+    for (let index = 0; index < repositories.length; index++) {
+      const element = repositories[index];
+      createAndAppend('option', repoDropDown, {
+        text: element.name,
+        value: index,
+      });
+    }
+    repoDropDown.onchange = event => repositoryChanged(event);
```
remarcmij: You can pass the selected repo directly:

```js
repoDropDown.onchange = () => repositoryChanged(repositories[repoDropDown.value]);
```

homework/index.js
mehmet-bozdemir: done

homework/index.js
mehmet-bozdemir: done

homework/index.js
mehmet-bozdemir: done

homework/index.js
mehmet-bozdemir: done

homework/index.js
mehmet-bozdemir: done

homework/index.js
mehmet-bozdemir: done

homework/index.js
mehmet-bozdemir: done

homework/index.js
mehmet-bozdemir: done

PR #303, opened 2019-07-27
finished js03-week01 homework

Laurelinduilwen:
I am sorry that I was this late, css is a bit unfinished. will fix everything asap and submit together with the week02 pull request

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -71,7 +71,7 @@
   function createContributors(rightColumn, requestURL) {
     fetchJSON(requestURL, (err, contributors) => {
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        createAndAppend('div', rightColumn, { text: err.message, class: 'alert-error' });
       }
       const rightUL = createAndAppend('ul', rightColumn, { id: 'main-list' });
       contributors.forEach(contributor => {
@@ -93,7 +93,7 @@
           href: contributor.html_url,
         });
         createAndAppend('div', contributorWrapper, {
-          text: contributor.contributions,
+          text: `Number of contributions: ${contributor.contributions}`,
           class: 'number-of-commits',
         });
       });
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,6 +1,9 @@
 'use strict';
 
 {
+  const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
+  const root = document.getElementById('root');
+
   function fetchJSON(url, cb) {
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -30,17 +33,124 @@
     return elem;
   }
 
+  function createDetails(leftColumn, repositories) {
+    const table = createAndAppend('table', leftColumn, { id: 'main-table' });
+    const tBody = createAndAppend('tbody', table);
+    const repoDetails = ['Repository', 'Description', 'Forks', 'Updated'];
```
rohni: Nice job. I like how you are looping through this rather than just copying and pasting. Shows a very good understanding of programming being about creative laziness. 😄

homework/index.js
```diff
@@ -30,17 +33,124 @@
...
+
+    // Functional requirement 4 -providing a link to the repository in the name property of the selected object
+
+    createAndAppend('a', repositoryName, {
+      href: repositories.html_url,
+      target: '_blank',
+      text: repositories.name,
+      class: 'repository-name',
+    });
+    if (repositories.description === null) {
```
rohni: Nice job handling the possibility that there is no description.

homework/index.js
```diff
@@ -30,17 +33,124 @@
...
+    }
+    document.querySelector('.forks-data').innerText = repositories.forks;
+    document.querySelector('.updated-data').innerText = new Date(
+      repositories.updated_at,
+    ).toLocaleString();
+  }
+
+  function createContributors(rightColumn, requestURL) {
+    fetchJSON(requestURL, (err, contributors) => {
+      if (err) {
```
rohni: I would put the error message for getting contributors in the rightColumn rather than in root. If this query goes wrong, only the contributor data is missing, while the data over the repository can still show. As we don't know what the user is looking for, this repository data might be all they need (e.g. they want to see how many forks etc.).

Laurelinduilwen:
thank you for the feedback :D I have improved the css and implemented the suggestions ^^

homework/index.js
Laurelinduilwen: fixed this, I think I missed it due to my heavy use of ctrl + c and ctrl + v 😅

homework/index.js
Laurelinduilwen: laziness is the mother of  inventions I guess :p

homework/style.css
```diff
@@ -1,3 +1,78 @@
+/* will update the css ASAP */
```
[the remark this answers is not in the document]
Laurelinduilwen: fixed this one, thanks for the detailed feedback ^^

PR #305, opened 2019-07-30
js3-week1/Adil

musafir928:


homework/index.js
```diff
@@ -30,14 +30,129 @@
...
+  function renderRepositoryBasicInfo(repository, table) {
+    table.innerHTML = '';
+    addRow(repository, table, 'name');
+    addRow(repository, table, 'description');
+    addRow(repository, table, 'forks');
+    addRow(repository, table, 'updated_at');
+    const upDate = document.getElementById('updated_at');
+    upDate.innerHTML = '';
+    const dateArray = repository.updated_at.split('T');
+    upDate.innerHTML = `<em>${dateArray[0]}</em>, ${dateArray[1].substring(
```
remarcmij: This is a style break with the rest of the code. We are creating all elements with `createAndAppend`. For consistency, you should do the same here. You could an `<em>` element like this:

```js
const datePart = createAndAppend('em', update, {text: dateArray[0]});
```

But then again, I would not bother trying to emphasize the date part at all.

homework/index.js
```diff
@@ -30,14 +30,129 @@
...
+      text: `${property.toUpperCase()}: `,
+    });
+    if (property !== 'name') {
+      createAndAppend('td', tr, {
+        text: `${repository[property]}`,
+        id: property,
+      });
+    } else {
+      const firstRow = createAndAppend('td', tr);
+      createAndAppend('a', firstRow, {
```
remarcmij: The assignment states that the associated GitHub repo page should be opened in a **new browser tab**.

homework/index.js
```diff
@@ -30,14 +30,129 @@
...
+      contributions.forEach(contributor => {
+        const li = createAndAppend('li', ul);
+        const section = createAndAppend('section', li);
+        createAndAppend('img', section, {
+          src: contributor.avatar_url,
+          alt: contributor.login,
+        });
+        const sectionName = createAndAppend('section', li);
+        const h3 = createAndAppend('h3', sectionName);
+        createAndAppend('a', h3, {
```
remarcmij: The assignment states that the associated GitHub repo page should be opened in a **new browser tab**.

homework/index.js
```diff
@@ -30,14 +30,129 @@
...
   function main(url) {
     fetchJSON(url, (err, repositories) => {
       const root = document.getElementById('root');
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        renderError(err);
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+      // create header
```
remarcmij: I can already conclude that you are creating a reader by reading the code. What is the added value of this comment?

homework/index.js
```diff
@@ -30,14 +30,129 @@
...
+      const header = createAndAppend('div', root, {
+        class: 'header',
+      });
+      createAndAppend('h1', header, {
+        text: 'HYF Repositories: ',
+        class: 'title',
+      });
+      const select = createAndAppend('select', header);
+      // sort repositories
```
remarcmij: This comment restates what the code already clearly indicates. Don't state the obvious.

homework/index.js
```diff
@@ -30,14 +30,129 @@
...
+      repositories.sort((a, b) => a.name.localeCompare(b.name));
+      createOptions(repositories, select);
+      // left-side: info div
+      const left = createAndAppend('div', root, {
```
remarcmij: Don't use comments to beef up an inadequate variable name. If you want to indicate that this variable holds a reference to a div for the left side containing 'info', reflect that in the name and get rid of the comment: let the code speak for itself.

```js
const leftInfoDiv = createAndAppend(...);
```

homework/style.css
```diff
@@ -1,3 +1,114 @@
+* {
```
remarcmij: You probably will not win first prize for your UI design (nor will I, for that matter). On a mobile, your select box seems to run off the right margin:

image

PR #306, opened 2019-07-30
Add Week2 Homework - Oguzhan

oguzhanturgut:
- Add Promise, introduce `initializePage` function

- Add ARIA-compliance, make contributor `li`s focusable, add `keyup` eventListener

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,19 +1,21 @@
 'use strict';
 
 {
-  function fetchJSON(url, cb) {
-    const xhr = new XMLHttpRequest();
-    xhr.open('GET', url);
-    xhr.responseType = 'json';
-    xhr.onload = () => {
-      if (xhr.status >= 200 && xhr.status <= 299) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
+      const xhr = new XMLHttpRequest();
+      xhr.open('GET', url);
+      xhr.responseType = 'json';
+      xhr.onload = () => {
+        if (xhr.status >= 200 && xhr.status <= 299) {
+          resolve(xhr.response);
+        } else {
+          reject(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
+        }
+      };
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -30,17 +32,152 @@
...
+          contributorsTitle.textContent = 'No Contributor So Far';
+          return;
+        }
+        const contributorsList = createAndAppend('ul', rightSide, {
+          class: 'contributors-list',
+        });
+        contributors.forEach(contributor => {
+          const listItem = createAndAppend('li', contributorsList, {
+            class: 'contributor',
+            tabindex: 0,
```
remarcmij: 👍 To include the `li` in the tab order so it can be navigated to with the keyboard.

homework/index.js
```diff
@@ -30,17 +32,152 @@
...
+            text: contributor.contributions,
+          });
+          listItem.addEventListener('click', () => {
+            // Go to new page on click
+            window.open(contributor.html_url, '_blank');
+          });
+          listItem.addEventListener('keyup', event => {
+            // Open new page with contributor url when pressed Enter on focused contributor
+            if (event.key === 'Enter') window.open(contributor.html_url, '_blank');
+          });
```
remarcmij: Adding these event listeners in combination with setting the tabindex allows the list item to be activated with both the mouse and the keyboard. Alternatively, you could use an `a` tag that implements this behaviour by default.

homework/index.js
```diff
@@ -30,17 +32,152 @@
...
+    });
+    createAndAppend('select', header, {
+      id: 'select',
+      'aria-label': 'HYF Repositories',
+    });
+
+    fetchJSON(url)
+      .then(repositories => initializePage(repositories))
+      .then(repository => fetchAndRender(repository))
+      .catch(error => renderError(error));
```
remarcmij: `initiazePage` does not return a promise by itself. But this still works because the first `.then()` return a promise that is immediately resolved to the return value of `initializePage`. Alternatively, you could code it likes this:

```js
fetchJSON(url)
  .then(repositories => {
    const repository = initializePage(repositories);
    return fetchAndRender(repository);
  })
  .catch(error => renderError(error));
```

On the plus side, your implementation reads nicely. On the downside, it implies something that does not reflect reality.

oguzhanturgut: Changed the code following your advice. Actually i trusted the `then`'s promise wrapping mechanism if the result is not a promise.

oguzhanturgut:
Thanks for your time making review for the homework

PR #307, opened 2019-07-30
Homework/week2

hnnhack:
Last changes from week1 and homework week2.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -32,6 +32,15 @@
     return elem;
   }
 
+  function renderError(error) {
+    const container = document.getElementsByClassName('container');
+    container.innerHTML = '';
+    createAndAppend('div', container, {
+      text: error.message,
+      class: 'alert',
+    });
+  }
+
   function showRepositoryInfo(repository) {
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -30,17 +34,127 @@
...
+        });
+      })
+      .catch(error => renderError(error));
+  }
+
+  function main() {
+    getAndAppend(HYF_REPOS_URL);
+  }
+
+  window.onload = () => main();
```
remarcmij: It is not absolutely necessary to have a function main. In your case, it just calls another function.

```js
  // function main() {
  //   getAndAppend(HYF_REPOS_URL);
  // }

  window.onload = () => getAndAppend(HYF_REPOS_URL);
}
```

Alternatively, you can directly assign your function main to `window.onload`, as long as you leave out the parentheses (otherwise you would be calling main).

```js
  function main() {
    getAndAppend(HYF_REPOS_URL);
  }

  window.onload = main;
}
```

hnnhack: Hi Jim!
Thank you very much for time to review!

PR #309, opened 2019-07-31
Js3_week1_Homework

Farzanna:
Hi @remarcmij this is my week1 homework, I apologize for being late

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,5 +1,3 @@
-'use strict';
-
 {
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
@@ -30,14 +28,120 @@
     return elem;
   }
 
+  function clearContainer(container) {
+    while (container.firstChild) {
+      container.removeChild(container.firstChild);
+    }
+  }
+
+  function renderError(err) {
+    const listContainer = document.getElementById('list-container');
+    clearContainer(listContainer);
+    const root = document.getElementById('root');
+    createAndAppend('div', root, { text: err.message, class: 'alert alert-error' });
+  }
+
+  function addRow(tbody, label, repository) {
+    const tr = createAndAppend('tr', tbody);
+    createAndAppend('td', tr, { text: `${label}:`, class: 'label' });
+    if (label === 'Repository:') {
+      const urlRepo = repository.html_url;
+      const td = createAndAppend('td', tr);
+      createAndAppend('a', td, {
+        text: repository.name,
+        href: urlRepo,
+        target: '_blank',
+      });
+    }
+    if (label === 'Description:') {
+      createAndAppend('td', tr, { text: repository.description });
+      console.log(repository.description);
+    }
+    if (label === 'Forks:') {
+      createAndAppend('td', tr, { text: repository.forks });
+    }
+
+    if (label === 'Updated:') {
+      createAndAppend('td', tr, { text: repository.updated_at });
+    }
+  }
+
+  function createInfoList(repository, container) {
+    clearContainer(container);
+    const div = createAndAppend('div', container, { id: 'info_div' });
+    const table = createAndAppend('table', div);
+    const tbody = createAndAppend('tbody', table);
+    createAndAppend('tr', table);
+    createAndAppend('tr', table, {});
+    addRow(tbody, 'Repository:', repository);
+    addRow(tbody, 'Description:', repository);
+    addRow(tbody, 'Forks:', repository);
+    addRow(tbody, 'Updated:', repository);
+  }
+
+  function createContributorsList(repositoryName, ul) {
+    clearContainer(ul);
+    const url = `https://api.github.com/repos/HackYourFuture/${repositoryName}/contributors`;
+    fetchJSON(url, (err, data) => {
+      if (err) {
+        renderError(err);
+        return;
+      }
+
+      createAndAppend('h2', ul, {
+        text: 'contributions',
+      });
+      data.map(item => {
+        const li = createAndAppend('li', ul);
+        const div = createAndAppend('div', li);
+
+        createAndAppend('img', div, {
+          class: 'contributor_avatar',
+          src: item.avatar_url,
+          alt: item.login,
+        });
+        const p1 = createAndAppend('p', div, { id: 'p1' });
+        const p2 = createAndAppend('p', div, { id: 'p2' });
+        p1.textContent = item.login;
+        p2.innerHTML = item.contributions;
+        return item;
+      });
+    });
+  }
+
+  function startPage(data, listContainer, ul) {
+    createContributorsList(data.name, listContainer);
+    createInfoList(data, ul);
+  }
+
   function main(url) {
-    fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root);
+    const section = createAndAppend('section', root);
+    createAndAppend('h1', header, { text: 'HYF Repositories' });
+    const ul = createAndAppend('ul', section, { id: 'list-container' });
+    const select = createAndAppend('select', header);
+    const listContainer = createAndAppend('ul', section, { id: 'Contributors_list' });
+    fetchJSON(url, (err, data) => {
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+      data
+        .sort((a, b) => a.name.localeCompare(b.name))
+        .forEach((item, index) => {
+          createAndAppend('option', select, { text: item.name, value: index });
+        });
+      startPage(data[0], listContainer, ul);
+
+      select.addEventListener('change', () => {
+        clearContainer(ul);
+        clearContainer(listContainer);
+        const repositoryData = data[select.value];
+        const repositoryName = data[select.value].name;
+        createContributorsList(repositoryName, listContainer);
+        createInfoList(repositoryData, ul);
+      });
     });
   }
 
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,3 +1,5 @@
+'use strict';
+
 {
   function fetchJSON(url, cb) {
     const xhr = new XMLHttpRequest();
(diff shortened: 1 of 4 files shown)


homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+  }
+
+  function renderError(err) {
+    const listContainer = document.getElementById('list-container');
+    clearContainer(listContainer);
+    const root = document.getElementById('root');
+    createAndAppend('div', root, { text: err.message, class: 'alert alert-error' });
+  }
+
+  function addRow(tbody, label, repository) {
```
remarcmij: Excellent that you tried to make your code DRY by adding this function. It is a good first try, however, in your implementation, you do not gain much as it needs detailed knowledge about the structure of the repository object. Alternatively, the `addRow` function can be as simple as this:

```js
function addRow(tbody, label, value = '') {
  const row = createAndAppend('tr', tbody);
  createAndAppend('td', row, { text: `${label} :`, class: 'label' });
  createAndAppend('td', row, { text: value });
  return row;
}
```

In this implementation, the `addRow` function is completely detached from any knowledge about where the data comes from. It just needs a parent `tbody` element and two string arguments, one for the label text and one for the text value to be displayed. You can call use it like this (note how the first row get some extra treatment to add the required `<a>` tag):

```js
const firstRow = addRow(tbody, 'Repository');
  href: repo.html_url,
  target: '_blank',
  text: repo.name,
});

addRow(tbody, 'Description', repo.description);
addRow(tbody, 'Forks', repo.forks);
addRow(tbody, 'Updated', new Date(repo.updated_at).toLocaleString());
```

This `addRow` function is reusable whereas your version is highly specialized to work only with repository objects.

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+      const td = createAndAppend('td', tr);
+      createAndAppend('a', td, {
+        text: repository.name,
+        href: urlRepo,
+        target: '_blank',
+      });
+    }
+    if (label === 'Description:') {
+      createAndAppend('td', tr, { text: repository.description });
+      console.log(repository.description);
```
remarcmij: Always remove console.log statements that you have used for debugging purposes before you make a pull request.

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+    }
+
+    if (label === 'Updated:') {
+      createAndAppend('td', tr, { text: repository.updated_at });
+    }
+  }
+
+  function createInfoList(repository, container) {
+    clearContainer(container);
+    const div = createAndAppend('div', container, { id: 'info_div' });
```
remarcmij: You are generating some invalid HTML here. First of all, you don't need a `ul` here as you are not creating a list. Moreover, a `ul` can only have `li` elements as its children, and not a `div` as you have here.

Inside your `tbody` you also have two empty `tr` elements which serve no purpose at all:

```html
<ul id="list-container">
  <div id="info_div">
    <table>
      <tbody>
        <tr>
          <td class="label">Repository::</td>
        </tr>
        <tr>
          <td class="label">Description::</td>
        </tr>
        <tr>
          <td class="label">Forks::</td>
          <td>9</td>
        </tr>
        <tr>
          <td class="label">Updated::</td>
          <td>2019-06-21T11:38:21Z</td>
        </tr>
      </tbody>
      <tr></tr>
      <tr></tr>
    </table>
  </div>
</ul>
```

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+  function createContributorsList(repositoryName, ul) {
+    clearContainer(ul);
+    const url = `https://api.github.com/repos/HackYourFuture/${repositoryName}/contributors`;
+    fetchJSON(url, (err, data) => {
+      if (err) {
+        renderError(err);
+        return;
+      }
+
+      createAndAppend('h2', ul, {
```
remarcmij: Here is some invalid HTML too. As I said, a `ul` can only have `li` children, and note an `h2` as you have here:

```html
<ul id="Contributors_list">
  <h2>contributions</h2>
  <li>
    <div>
      <img
        class="contributor_avatar"
        alt="gijscor"
      />
      <p id="p1">gijscor</p>
      <p id="p2">47</p>
    </div>
  </li>
  <li>
    <div>
      <img
        class="contributor_avatar"
        alt="wouterkleijn"
      />
      <p id="p1">wouterkleijn</p>
      <p id="p2">11</p>
    </div>
  </li>
</ul>
```

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+        text: 'contributions',
+      });
+      data.map(item => {
```
remarcmij: 1. You should only use `.map()` if you want to create a new array by transforming the elements of the source array, for example as you did in the JS2 homework, e.g.:

    ```js
    const doubledNumbers = numbers.map(number => number * 2);
    ````

    In this case, where you just one to loop through the elements to create DOM elements, you should use a `.forEach()`.

2. Try and avoid generic names such as `data` and `item` if you have more specific knowledge about the data you are dealing with. In this case, the `data` concerns an array of `contributors`. An element of that array is a `contributor`. Reveal your intentions by using these names explicitly:

```js
```

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+    createAndAppend('tr', table, {});
+    addRow(tbody, 'Repository:', repository);
+    addRow(tbody, 'Description:', repository);
+    addRow(tbody, 'Forks:', repository);
+    addRow(tbody, 'Updated:', repository);
+  }
+
+  function createContributorsList(repositoryName, ul) {
+    clearContainer(ul);
+    const url = `https://api.github.com/repos/HackYourFuture/${repositoryName}/contributors`;
```
remarcmij: Do not construct this url yourself. Instead use the `.contributors_url` property of the repository object, as per the hint from the assignment. The reason for this is that GitHub may choose to change the internal structure of the url in the future. However, it will continue to support the documented `.contributors_url` property.

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+        const div = createAndAppend('div', li);
+
+        createAndAppend('img', div, {
+          class: 'contributor_avatar',
+          src: item.avatar_url,
+          alt: item.login,
+        });
+        const p1 = createAndAppend('p', div, { id: 'p1' });
+        const p2 = createAndAppend('p', div, { id: 'p2' });
+        p1.textContent = item.login;
```
remarcmij: 1. It is not necessary to use `.textContent` or `.innerHTML` properties here. Instead, you can use the `text` property of `createAndAppend()`. As a bonus it also allows you to get rid of the `p1` and `p2` variables.

    ```js
    createAndAppend('p', div, { id: 'p1', text: item.login });
    createAndAppend('p', div, { id: 'p2', text: item.contributions });
    ```

2. This assignment requires that you should be able to visit the contributors GitHub page in a new tab when you click on the list item (or navigate to it with the tab key and press Enter, for visually impaired users).

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+        p2.innerHTML = item.contributions;
+        return item;
```
remarcmij: There is no need to return `item`. There is no code that makes use of it.

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
-    fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root);
+    const section = createAndAppend('section', root);
+    createAndAppend('h1', header, { text: 'HYF Repositories' });
+    const ul = createAndAppend('ul', section, { id: 'list-container' });
+    const select = createAndAppend('select', header);
+    const listContainer = createAndAppend('ul', section, { id: 'Contributors_list' });
+    fetchJSON(url, (err, data) => {
```
remarcmij: `data` -> `repositories`

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
       if (err) {
         createAndAppend('div', root, { text: err.message, class: 'alert-error' });
         return;
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+      data
+        .sort((a, b) => a.name.localeCompare(b.name))
+        .forEach((item, index) => {
```
remarcmij:
```js
.forEach((repository, index) => {
```

homework/index.js
```diff
@@ -30,14 +30,120 @@
...
+        });
+      startPage(data[0], listContainer, ul);
+
+      select.addEventListener('change', () => {
+        clearContainer(ul);
+        clearContainer(listContainer);
+        const repositoryData = data[select.value];
+        const repositoryName = data[select.value].name;
+        createContributorsList(repositoryName, listContainer);
+        createInfoList(repositoryData, ul);
```
remarcmij: You already have a `startPage` function that does what lines 144-145 do. Why not reuse that function and make your code DRY?

homework/style.css
```diff
@@ -1,3 +1,105 @@
 .alert-error {
```
remarcmij: Your app is not responsive.

image

Farzanna:
Hi Jim thank you for all your comments and your time for reviewing my code and giving valuable remarks, I am working hard to compensate my absence
And going to correct my code based on your comments or at least explain why I have done that way

remarcmij:
Hi @Farzanna, I'm still waiting for your updates and also for your week 2 homework. Or an explanation please why it is not here yet.

Farzanna:
hi @remarcmij , I was watching the video of lesson two times, but it seems still have trouble with promise, I read and watch lots of resources and it makes me more confused

remarcmij:
In that case, you should ask for help on slack. I'm happy to do an online session using slack with you later this afternoon if that suits you, to go over promises. You should not just let time pass by like this without making progress. Otherwise you will not make it through the curriculum.

homework/index.js
```diff
@@ -30,14 +30,103 @@
...
+  function addRow(tbody, label, value = '') {
+    const row = createAndAppend('tr', tbody);
+    createAndAppend('td', row, { text: `${label} :`, class: 'label' });
+    createAndAppend('td', row, { text: value });
+    return row;
+  }
+
+  function createInfoList(repository, container) {
+    clearContainer(container);
+    const tbody = createAndAppend('tbody', container);
```
remarcmij: You are missing a `table` tag. A `tbody` should be the child of a `table` tag, it cannot exist on itself.

homework/index.js
```diff
@@ -30,14 +30,103 @@
...
+          src: contributor.avatar_url,
+          alt: contributor.login,
+        });
+        createAndAppend('p', div, { id: 'p1', text: contributor.login });
+        createAndAppend('p', div, { id: 'p2', text: contributor.contributions });
+      });
+    });
+  }
+
+  function startPage(repositories, listContainer, ul) {
```
remarcmij: The name `repositories` (plural) implies an array. But you passing a single repository as the first argument. Therefore the argument name should be singular: `repository`.

homework/index.js
```diff
@@ -30,14 +30,104 @@
...
+          id: 'link_contributions',
+        });
+        const div = createAndAppend('div', a);
+
+        createAndAppend('img', div, {
+          class: 'contributor_avatar',
+          src: contributor.avatar_url,
+          alt: contributor.login,
+        });
+        createAndAppend('p', div, { id: 'p1', text: contributor.login });
```
remarcmij: I saw a new issue: because you are creating this in a loop you end up with multiple occurrences of  **id** `p1` and `p2`. In HTML ids must be unique (occur once only). So in this case, you need to use CSS classes instead of ids.

Farzanna: applied

PR #313, opened 2019-07-31
JavaScript3-week2/Shyar

ShyarK:


homework/index.js
```diff
@@ -30,17 +15,134 @@
...
+    fetchJSON(url)
+      .then(contributors => {
+        contributesDiv.innerHTML = '';
+        createAndAppend('h1', contributesDiv, {
+          class: 'contributes-header',
+          text: 'Contributions',
+        });
+        const ul = createAndAppend('ul', contributesDiv);
+
+        Object.keys(contributors).forEach(contributor => {
```
remarcmij: This is something I overlooked in the week 1 homework: the `contributors` obtained from `fetchJSON` is an array. While an array is also an object, its keys are simply the index values from zero up to `length-1`. So `Object.keys()` is not really useful here. Better use `.forEach()`:

```js
```

`contributor` is no longer a key or an index, but an element of the `contributors` array, which is a contributor object.

PR #317, opened 2019-08-01
Week2/Adil

musafir928:


musafir928:
CSS - section overflow solved,
addRow - function simplified as required,
tbody issues solved,
comments deleted,
week2 homework added;

homework/index.js
```diff
@@ -30,15 +32,119 @@
...
+      class: label,
+    });
+    createAndAppend(
+      'td',
+      tr,
+      value
+        ? {
+            text: value,
+          }
+        : undefined,
```
remarcmij: There is a slight issue here. I would like to know if a repo has zero forks. However, `0` is also falsy. Therefore the number zero will not be displayed with your code. This will work:

```js
createAndAppend('td', tr, value == null ? undefined : { text: value });
```

`value == null` with be true if value is `null` or `undefined`. Comparing to `null` like this is the only time you can use a non-strict comparison with `==`.

musafir928: Thank you for your feedback, I have just tried it and solved it.

PR #318, opened 2019-08-01
Adham js3 homework week2

Adham-Elias-Botours:
Dear @remarcmij , this is my homework js3 week02. Also with my own review about my work.
Thanks in advance.

**In this version:**

- The information about a repository are displayed in table.
- If the repository has no description or contributions, then a sentence will be displayed to notify that.
- The page's style and layout is adjusted.
- Replace Promise object instead CallBack functions to make the XML request.

**what could be better:**

- The code structure: if I want to display data from server, then it is better to make 3 function. 
The first to fetch the data and it contains the XML request as a Promise object (depending on the Promise's result the second function or the third function will be executed ). 
The second  is to render the result form the request in the page. 
The third to render the error if the request has error.
- Displaying the distributions in list instead of DIV element. Then it will be much easier to sort the list alphabetically.
- Another technique to notify that the repository does not have the wanted information. 
js3-week2

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -108,7 +108,7 @@
     }
     fetchJSON(urlText)
       .then(goodResult => renderContributions(goodResult))
-      .then(notifyNoData())
+      .then(notifyNoData)
       .catch(somethingWrong => renderError(somethingWrong));
   }
 
--- a/homework/index.js
+++ b/homework/index.js
@@ -41,7 +41,7 @@
     const url = intext.innerHTML;
     const infoContainer = document.getElementById('info_container');
     const infoRight = createAndAppend('div', infoContainer, { class: 'info_right' });
-    const title = createAndAppend('h4', infoRight, { text: 'info_right' });
+    createAndAppend('h4', infoRight, { text: 'info_right' });
     fetchJSON(url, (err, data) => {
       if (err) {
         renderError(err);
--- a/homework/index.js
+++ b/homework/index.js
@@ -117,14 +117,13 @@
     }
   }
 
-  const clearInfoContainer = element => {
-    while (element.firstChild) {
-      element.removeChild(element.firstChild);
+  const deleteContents = container => {
+    while (container.firstChild) {
+      container.removeChild(container.firstChild);
     }
   };
 
   function starter(url) {
-    // ******
     const root = document.getElementById('root');
     const header = createAndAppend('header', root, {
       class: 'header',
@@ -141,7 +140,7 @@
       id: 'info_container',
       class: 'info-container',
     });
-    // *******
+
     function renderResult(repositories) {
       repositories.sort((a, b) => a.name.localeCompare(b.name));
       repositories.forEach((repo, index) => {
@@ -150,9 +149,8 @@
       showRepoInfo(repositories[selectEl.value], infoContainer);
       showContributions(repositories[selectEl.value].contributors_url, infoContainer);
       selectEl.addEventListener('change', () => {
-        clearInfoContainer(infoContainer);
+        deleteContents(infoContainer);
         showRepoInfo(repositories[selectEl.value], infoContainer);
-        console.log(repositories[selectEl.value].contributors_url);
         setTimeout(
           showContributions(repositories[selectEl.value].contributors_url, infoContainer),
           0,
--- a/homework/index.js
+++ b/homework/index.js
@@ -112,7 +112,7 @@
     if (secondRow.lastChild.textContent === '') {
       createAndAppend('span', secondRow.lastChild, {
         class: 'no-content',
-        text: 'There is NO description',
+        text: 'There is NO description.',
       });
     }
   }
--- a/homework/index.js
+++ b/homework/index.js
@@ -112,14 +112,8 @@
     addRow(tbody, `Description`, wantedRepo.description);
     addRow(tbody, `Forks`, wantedRepo.forks);
     addRow(tbody, `Last update`, new Date(wantedRepo.updated_at).toLocaleString());
-    const toBeLink = tbody.childNodes[0].childNodes[1];
-    toBeLink.addEventListener('mouseover', () => {
-      toBeLink.style.textDecoration = 'underline';
-    });
-    toBeLink.addEventListener('click', () => window.open(wantedRepo.html_url));
-    toBeLink.addEventListener('mouseout', () => {
-      toBeLink.style.textDecoration = 'none';
-    });
+    const toBeLink = tbody.firstChild.childNodes[1];
+    toBeLink.innerHTML = `<a href="${wantedRepo.html_url}" target="_blank">${wantedRepo.name}</a>`;
   }
   // ************************************
   const clearinfoContainer = () => {
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,5 +1,6 @@
 'use strict';
 
+// modifying
 {
   const HYF_REPOS_URL = 'https://api.github.com/orgs/HackYourFuture/repos?per_page=100';
 
@@ -77,7 +78,7 @@
     });
   }
 
-  function viewRepInfo() {
+  function viewRepoInf() {
     fetchJSON(HYF_REPOS_URL, (err, data) => {
       if (err) {
         renderError(err);
@@ -135,7 +136,6 @@
       const infoContainer = createAndAppend('div', document.body, {
         id: 'info_container',
         class: 'info_container',
-        text: 'You can choose a repository name from the list',
       });
       if (err) {
         infoContainer.innerHTML = '';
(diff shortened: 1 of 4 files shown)


Adham-Elias-Botours:
to view the work online:
https://adham-elias-botours.github.io/my-js3/week02/

homework/index.js
```diff
@@ -23,24 +10,151 @@
...
-    fetchJSON(url, (err, repositories) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
-        return;
+  function addRow(tbody, label, value) {
+    const tr = createAndAppend('tr', tbody);
+    createAndAppend('td', tr, { text: `${label}:` });
+    createAndAppend('td', tr, { text: value });
+  }
```
remarcmij: This function becomes more useful if you return `tr` as the return value, similar to how it is done in `createAndAppend()`.

```js
return tr;
```

homework/index.js
```diff
@@ -23,24 +10,151 @@
...
+
+  function showRepoInfo(wantedRepo) {
+    const infoLeft = createAndAppend('div', infoContainer, {
+      class: 'info-left',
+      text: 'Repository information:',
     });
+    const table = createAndAppend('table', infoLeft);
+    const tbody = createAndAppend('tbody', table);
+    addRow(tbody, `Name`, wantedRepo.name);
```
remarcmij: With `addRow()` returning the `tr` it created, you can now assign its return value to a variable, e.g.:

```js
const firstRow = addRow(tbody, `Name`, wantedRepo.name);
```

homework/index.js
```diff
@@ -23,24 +10,151 @@
...
+    addRow(tbody, `Description`, wantedRepo.description);
+    addRow(tbody, `Forks`, wantedRepo.forks);
+    addRow(tbody, `Last update`, new Date(wantedRepo.updated_at).toLocaleString());
+    const toBeLink = tbody.firstChild.childNodes[1];
+    toBeLink.innerHTML = `<a href="${wantedRepo.html_url}" target="_blank">${wantedRepo.name}</a>`;
```
remarcmij: This is ugly and inconsistent with the rest of the code. You can use `createAndAppend` instead:

```js
createAndAppend('a', firstRow, { href: ..., target: ..., text: ... });
``

File: homework/index.js [hunk 4]
```diff
@@ -23,24 +10,151 @@
...
+      showContributions(repositories[selectEl.value].contributors_url);
+      selectEl.addEventListener('change', () => {
+        clearinfoContainer();
+        showRepoInfo(repositories[selectEl.value]);
+        setTimeout(showContributions(repositories[selectEl.value].contributors_url), 0);
+      });
+    }
+    fetchJSON(url)
+      .then(goodResult => renderResult(goodResult))
+      .catch(somethingWrong => renderError(somethingWrong));
```
remarcmij: Funny, but not the best names... Describe the content of the data: `repositories`, `error`.

homework/index.js
```diff
@@ -23,24 +10,151 @@
...
+          class: 'no-content',
+          text: 'There are no contributions yet.',
+        });
       }
-      createAndAppend('pre', root, { text: JSON.stringify(repositories, null, 2) });
+    }
+
+    fetchJSON(urlText)
+      .then(goodResult => renderContributions(goodResult))
+      .catch(somethingWrong => renderError(somethingWrong));
```
remarcmij: Bad choice of names. See my comment for the same in the `starter())` function.

homework/index.js
Adham-Elias-Botours: I wrote so in order not te be confused with line 141. They are parameters with the same name. But now I got the logic .

homework/index.js
Adham-Elias-Botours: I need to embed the anchor tag in the cell, exactly the second cell. So I have just wrote like: 
`createAndAppend('a', firstRow.childNodes[1], {
      text: ....,
      href: .....,
      target: ....',
    });`
Is that good? or doe you mean that I do not need to use childNodes[]?

remarcmij: Ah yes, the easiest is to use firstRow.lastChild

homework/index.js
remarcmij: I made a mistake here. The third argument should just be an empty string ‘’ because you are going to add a child anchor tag.

Adham-Elias-Botours: Thanks

Adham-Elias-Botours:
Dear @remarcmij ;
This is my last work before the next lesson.
For now I think that all the requested changes are covered.
And sure in case there are more requested changes then I will work on them next week.
I have also another question (in the photo), it is not urgent but the answer could be helpful in the future. 
empty-repo

For online testing and viewing , this is the online link:
https://adham-elias-botours.github.io/my-js3/week02/

Best regards

homework/index.js
```diff
@@ -23,24 +10,157 @@
...
+    });
+  }
+
+  function showContributions(urlText, container) {
+    const infoRight = createAndAppend('div', container, {
+      class: 'info-right',
+      text: 'Contributions:',
+    });
+
+    function renderContributions(contributors) {
```
remarcmij: For the same reason that you should avoid global variables, you should also avoid nested named functions. It is tempting to use them, because you have access to the variables and parameters of the outer function, but again it can become difficult to track how this is all interconnected. You can place this function at the file level by supplying the value it needs from the outer function through the parameter list.

Also shown here is how you can deal with no contributors (`contributors == null`)

```js
function renderContributions(contributors, infoRight) {
  if (contributors == null) {
    return;
  }
  ...
}

Then, where you call your function, supply the extra parameter:

```js
function showContributions(urlText, container) {
  const infoRight = createAndAppend('div', container, {
    class: 'info-right',
    text: 'Contributions:',
  });

  fetchJSON(urlText)
    .then(contributors => renderContributions(contributors, infoRight))  // <- extra param
    .catch(error => renderError(error, infoRight));
}
```

homework/index.js
```diff
@@ -23,24 +10,157 @@
...
+    const selectEl = createAndAppend('select', header, {
+      class: 'select-list',
+      id: 'repositories_list',
+    });
+    const infoContainer = createAndAppend('div', root, {
+      id: 'info_container',
+      class: 'info-container',
+    });
+
+    function renderResult(repositories) {
```
remarcmij: Again, void this nested function. Place it at the file level and supply extra parameters:

```js
function renderResult(repositories, selectEl, infoContainer) {
```

```js
function starter(url) {
  ...
  fetchJSON(url)
    .then(repositories => renderResult(repositories, selectEl, infoContainer)) // <- extra params
    .catch(error => renderError(error, infoContainer));
}

File: homework/index.js [hunk 8]
```diff
@@ -23,24 +10,157 @@
...
+      repositories.sort((a, b) => a.name.localeCompare(b.name));
+      repositories.forEach((repo, index) => {
+        createAndAppend('option', selectEl, { text: repo.name, value: index });
+      });
+      showRepoInfo(repositories[selectEl.value], infoContainer);
+      showContributions(repositories[selectEl.value].contributors_url, infoContainer);
+      selectEl.addEventListener('change', () => {
+        deleteContents(infoContainer);
+        showRepoInfo(repositories[selectEl.value], infoContainer);
+        setTimeout(
```
remarcmij: Why have you used a `setTimeout` here?

homework/index.js
```diff
@@ -23,24 +10,157 @@
...
+
+  function showRepoInfo(wantedRepo, container) {
+    const infoLeft = createAndAppend('div', container, {
+      class: 'info-left',
+      text: 'Repository information:',
+    });
+    const table = createAndAppend('table', infoLeft);
+    const tbody = createAndAppend('tbody', table);
+    const firstRow = addRow(tbody, `Name`);
+    const secondRow = addRow(tbody, `Description`, wantedRepo.description);
```
remarcmij: To use a default value for `description` you can use this trick (google for it):

```js
addRow(tbody, `Description`, wantedRepo.description || 'There is NO description.');
```

This removes the need for lines 112-116. You also no longer need the `secondRow` variable.

homework/index.js
Adham-Elias-Botours: I forget why! Now I am asking me the same question.
But I noticed that the contributions column has a faster rendering on the screen  with using `setTimeout `. 
I have to think about that again.  It seems that I am going to delete it.

homework/index.js
Adham-Elias-Botours: Great .

Adham-Elias-Botours:
> Hi @Adham-Elias-Botours, thanks for your final updates. I still have some recommendations but I happy for you to implement them in week 3.

Thanks for all the tips and the recommendation. They are all considered.

PR #320, opened 2019-08-02
Js03 week02 homework

Laurelinduilwen:
sorry for being a few hours late 🐢

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -71,7 +71,7 @@
   function createContributors(rightColumn, requestURL) {
     fetchJSON(requestURL, (err, contributors) => {
       if (err) {
-        createAndAppend('div', root, { text: err.message, class: 'alert-error' });
+        createAndAppend('div', rightColumn, { text: err.message, class: 'alert-error' });
       }
       const rightUL = createAndAppend('ul', rightColumn, { id: 'main-list' });
       contributors.forEach(contributor => {
@@ -93,7 +93,7 @@
           href: contributor.html_url,
         });
         createAndAppend('div', contributorWrapper, {
-          text: contributor.contributions,
+          text: `Number of contributions: ${contributor.contributions}`,
           class: 'number-of-commits',
         });
       });
--- a/homework/index.js
+++ b/homework/index.js
@@ -35,39 +35,36 @@
     return elem;
   }
 
-  function createDetails(leftColumn, repositories) {
+  function addRow(tBody, label, value) {
+    const row = createAndAppend('tr', tBody);
+    if (value === null) {
+      createAndAppend('td', row, { text: `${label} :`, class: 'label' });
+      createAndAppend('td', row, {
+        text: `No ${label.toLowerCase()} is provided for this repository.`,
+      });
+    } else {
+      createAndAppend('td', row, { text: `${label} :`, class: 'label' });
+      createAndAppend('td', row, { text: value });
+    }
+    return row;
+  }
+
+  function jsUcfirst(string) {
+    return string.charAt(0).toUpperCase() + string.slice(1);
+  }
+
+  function createDetails(leftColumn, repository) {
     const table = createAndAppend('table', leftColumn, { id: 'main-table' });
     const tBody = createAndAppend('tbody', table);
-    const repoDetails = ['Repository', 'Description', 'Forks', 'Updated'];
-    repoDetails.forEach(detail => {
-      const tr = createAndAppend('tr', tBody, { class: `${detail.toLowerCase()}-row` });
-      createAndAppend('td', tr, {
-        text: `${detail}: `,
-        class: `${detail.toLowerCase()}-header`,
-      });
-      createAndAppend('td', tr, { class: `${detail.toLowerCase()}-data` });
-    });
-    const repositoryName = document.querySelector('.repository-data');
-
-    // Functional requirement 4 -providing a link to the repository in the name property of the selected object
-
-    createAndAppend('a', repositoryName, {
-      href: repositories.html_url,
+    const firstRow = addRow(tBody, 'Repository', '');
+    createAndAppend('a', firstRow.lastChild, {
+      href: repository.html_url,
       target: '_blank',
-      text: repositories.name,
-      class: 'repository-name',
+      text: jsUcfirst(repository.name),
     });
-    if (repositories.description === null) {
-      document.querySelector(
-        '.description-data',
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -30,17 +35,125 @@
     return elem;
   }
 
+  function createDetails(leftColumn, repositories) {
```
remarcmij: This function receives a single repository object. Therefore the second argument should be `repository`. The name `repositories` (plural) implies an array.

homework/index.js
```diff
@@ -30,17 +35,125 @@
...
+    const table = createAndAppend('table', leftColumn, { id: 'main-table' });
+    const tBody = createAndAppend('tbody', table);
+    const repoDetails = ['Repository', 'Description', 'Forks', 'Updated'];
+    repoDetails.forEach(detail => {
+      const tr = createAndAppend('tr', tBody, { class: `${detail.toLowerCase()}-row` });
```
remarcmij: You are using CSS classes to get a reference to them later in your code, querying the DOM with `querySelector`. While this works it would be better and more efficient to stay as much as possible in 'JavaScript land'. You could extend the approach you have chosen using a data structure (the `repoDetails` array), but a functional approach results in simpler code.

Basically, what needs to be done for each row in the table is to create a `tr` (parent - with two `td` children. The first one to hold label text and the second one to hold a property value from the repository object. So the fiunction could like this:

```js
function addRow(tbody, label, value) {
  const row = createAndAppend('tr', tbody);
  createAndAppend('td', row, { text: `${label} :`, class: 'label' });
  createAndAppend('td', row, { text: value });
  return row;
}
```

Now it is easy to add a row, e.g.:

```js
addRow(tbody, 'Forks', repo.forks);
```

For the first row, you would need to do something special, as it does not use simple text but an `<a>` tag:

```js
const firstRow = addRow(tbody, 'Repository');
  href: repo.html_url,
  target: '_blank',
  text: repo.name,
});
```

homework/style.css
```diff
@@ -1,3 +1,140 @@
+/* will improve css further in the next commit */
```
remarcmij: On a mobile, the fonts get tiny: better not scale them.

image

Laurelinduilwen:
Thank you for the feedback Jim, I am so sorry for being this late again but I will power through the week03 after this. 💪

homework/index.js
Laurelinduilwen: fixed 👍

homework/index.js
Laurelinduilwen: ahh I get it now :) another function for the "toolbox" ^^ thank you for showing me the correct way to do this, I have included an if statement here for repositories without a description but I wonder if there is a more elegant way of handling it since it kinda doubles the code 🤷‍♀

homework/style.css
Laurelinduilwen: fixed and got rid of the logo since it created more problems while scaling

homework/index.js
```diff
@@ -30,17 +35,122 @@
...
+      createAndAppend('td', row, { text: `${label} :`, class: 'label' });
+      createAndAppend('td', row, {
+        text: `No ${label.toLowerCase()} is provided for this repository.`,
+      });
+    } else {
+      createAndAppend('td', row, { text: `${label} :`, class: 'label' });
+      createAndAppend('td', row, { text: value });
+    }
+    return row;
+  }
```
remarcmij: This function can indeed be made simpler. As a first approach, you could move the common part of the `if` and `else` clause out of the `if` statement:

```js
function addRow(tBody, label, value) {
  const row = createAndAppend('tr', tBody);
  createAndAppend('td', row, { text: `${label} :`, class: 'label' });
  if (value === null) {
    createAndAppend('td', row, {
      text: `No ${label.toLowerCase()} is provided for this repository.`,
    });
  } else {
    createAndAppend('td', row, { text: value });
  }
  return row;
}
```

A further  improvement is to create an intermediate variable and use a ternary operator to check for `null`:

```js
function addRow(tBody, label, value) {
  const text =
    value != null ? value : `No ${label.toLowerCase()} is provided for this repository.`;
  const row = createAndAppend('tr', tBody);
  createAndAppend('td', row, { text: `${label} :`, class: 'label' });
  createAndAppend('td', row, { text });
  return row;
}
```

Notes:
1. When comparing to `null` you can use the non-strict operators `==` and `!=`. In this way, `null` and `undefined` will both meet the comparison condition. This is the only situation where it is recommended to use the non-strict versions.
2. Note the syntax `{ text }`. This is ES6 shorthand notation for `{ text: text }`.

homework/index.js
```diff
@@ -30,17 +35,122 @@
...
+
+  function createDetails(leftColumn, repository) {
+    const table = createAndAppend('table', leftColumn, { id: 'main-table' });
+    const tBody = createAndAppend('tbody', table);
+    const firstRow = addRow(tBody, 'Repository', '');
+    createAndAppend('a', firstRow.lastChild, {
+      href: repository.html_url,
+      target: '_blank',
+      text: jsUcfirst(repository.name),
```
remarcmij: While it looks user-friendly to capitalize the repository name, with this app supposedly targetting developers who use GitHub, it is important to display the information exactly as it is returned by GitHub. It can become confusing if there is a discrepancy between the case of the displayed name and the actual GitHub repo name. 

Having said that, 👍 for creating a separate function to capitalize the word (although I would call it `capitalize`).

homework/index.js
```diff
@@ -30,17 +35,122 @@
...
+          const contributorWrapper = createAndAppend('div', contributorItem, {
+            class: 'contributor-wrapper',
+          });
+          createAndAppend('img', contributorWrapper, {
+            src: contributor.avatar_url,
+            class: 'contributor-avatar',
+            alt: `Avatar of ${contributor.login}`,
+          });
+
+          createAndAppend('a', contributorWrapper, {
```
remarcmij: It would be nice if clicking anywhere on the list item would take you to the contributor's page. You can achieve this by wrapping the entire inner content of the `li` into an anchor tag with an appropriate `href` and replacing your current anchor tag with just a `div` or `p` or something like that.

homework/style.css
remarcmij: Your layout still leaves room for improvement.

Mobile:

image

On a wide-screen it would be better to limit the content width rather than stretching it out to the full width:

image

PR #324, opened 2019-08-05
Js3 weeek 3 part 1

elicarcar:
In some parts, instead of assigning a new variable, I preferred to combine try with .then, just for practicing both styles.


homework/index.js
```diff
@@ -30,15 +24,98 @@
...
+  }
+
+  async function main(url) {
+    const root = document.getElementById('root');
+    const menuSection = createAndAppend('section', root, { id: 'menu-section' });
+    createAndAppend('p', menuSection, { text: 'HYF Repositories', id: 'hyf-text' });
+    const select = createAndAppend('select', menuSection, { id: 'select-button' });
+    const bodyDiv = createAndAppend('div', root, { id: 'body-div' });
+    const section = createAndAppend('section', bodyDiv, { class: 'repo-info-list' });
+    const contrDiv = createAndAppend('div', bodyDiv, {
```
remarcmij: The semantics of your HTML tags seems somewhat disorganized. You have `section` for the menu, a `div` for the `body-div`, then as its children, another `section` and a `div`:

```html
<div id="root">
  <section id="menu-section"><!-- ... --></section>
  <div id="body-div">
    <section class="repo-info-list"><!-- ... --></section>
    <div class="contributors-div" id="cont-ulist"><!-- ... --></div>
  </div>
</div>
```

Instead, consider this:

```html
<div id="root">
  <header id="menu-section"
    <!-- ... -->
  </header>
  <main id="body-div">
    <section class="repo-info-list">
      <!-- ... -->
    </section>
    <section class="contributors-div" id="cont-ulist">
      <!-- ... -->
    </section>
  </main>
</div>
```

homework/style.css
```diff
@@ -1,3 +1,191 @@
 .alert-error {
```
remarcmij: Your CSS still needs some tweaking. When I start your application I get this:

image

When I switch to mobile emulation in Chrome Developer tools I get this:

image

When I switch back again to non-mobile, the number of contributions seems to be in the right place. Looks like some timing / animation issue.

image

PR #329, opened 2019-08-06
JavaScript3/Week3/Shyar

ShyarK:


homework/index.js
```diff
@@ -30,17 +15,135 @@
...
+
+  async function renderDefaultPage(url, select, tableDiv, contributesDiv) {
+    let repositories;
+    try {
+      const data = await fetchJSON(url);
+      repositories = data.sort((a, b) => a.name.localeCompare(b.name));
+      createOptionElements(repositories, select);
+      renderRepository(tableDiv, repositories[select.value]);
+      renderContributions(repositories[select.value].contributors_url, contributesDiv);
+      renderEventListener(repositories, select, tableDiv, contributesDiv);
```
remarcmij: The verb `render` in IT is generally understood as visualizing data on a medium. In this case, where you are not visualizing anything, a better name would be `addEventListener`, were it not that this name is too easily confused with `element.addEventListener`.

I think I would create a function, say `createHeader`, in which I would do all the header related stuff, including adding the event listener to the `select` element. Then I would call `createHeader()` from `main()`.

homework-classes/Contributor.js
```diff
@@ -14,6 +14,26 @@ class Contributor {
    */
   render(container) {
     // TODO: replace the next line with your code.
```
remarcmij: Remove comments lines that have outlived their purpose.

homework/index.js
ShyarK: Thanks a lot for taking the time and effort in reviewing my homework! And I am very happy that you accept my PR as-is!

PR #333, opened 2019-08-08
Submission Js3 week3 Part1 -- Mehmet E.

Mehmet-Ergorgec:
hi @remarcmij ;
I have already done first part. But I have learned for second part  what the classes are, extends, super functions. and I tried to understand why we need to use classes. I hope I will do part 2 ASAP. I spent this week try to understand previous topics especially Js3 module topics for exam.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -6,6 +6,7 @@
   "editor.tabSize": 2,
   "cSpell.words": [
     "networkidle",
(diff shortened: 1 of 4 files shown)


Mehmet-Ergorgec:
Thank you for your review. I am going to learn about  data encapsulation. I heard about it but did not pay attention before.

PR #336, opened 2019-08-10
very late homework

Remzi1:


Remzi1:
Hi Jim,It is very late, I know, may be it is the last homework in my hyf life. But I want to show that I try much, at least this week. I only look at only Mehmet s assignment but I don't just copy, I can find solutions there. Excuse me for late replyRemzi Sent from my Huawei phone-------- Original Message --------Subject: Re: [HackYourFuture/JavaScript3] very late homework (#336)From: Jim Cramer To: HackYourFuture/JavaScript3 CC: Remzi ,Mention @remarcmij commented on this pull request.

Hi @Remzi1, this homework is indeed late and actually too late as far as I am concerned. Furthermore, it looks like you have been borrowing heavily from other peoples work and the feedback I have been giving them, so I can hardly see this as your own original creation. If you think my assessment is wrong, please explain where you think I'm wrong.

PR #340, opened 2019-09-17
week 1 - homework

msnmz:


msnmz:
Thank you very much @remarcmij for your detailed explanation and also great lesson. 
Yes, I have prior experience, but also I learned a lot from your class and also I watched the 'Uncle Bob' s video that you posted. It is also great, thank you very much for your intensive effort on us.

PR #341, opened 2019-09-17
js3 w1 homework

rabrad:
For your review

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -30,12 +30,33 @@
   }
 
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const li = createAndAppend('li', ul, { class: 'list-info' });
+    const table = createAndAppend('table', li);
+    const trName = createAndAppend('tr', table);
+    createAndAppend('th', trName, { text: 'Repository: ' });
+    const tdName = createAndAppend('td', trName);
+    createAndAppend('a', tdName, {
+      text: repo.name,
+      href: repo.html_url,
+      target: '_blank',
+    });
+    const trDesc = createAndAppend('tr', table);
+    createAndAppend('th', trDesc, { text: 'Description: ' });
+    createAndAppend('td', trDesc, { text: repo.description });
+    const trFork = createAndAppend('tr', table);
+    createAndAppend('th', trFork, { text: 'Forks: ' });
+    createAndAppend('td', trFork, { text: repo.forks_count });
+    const trUpdate = createAndAppend('tr', table);
+    createAndAppend('th', trUpdate, { text: 'Last Update: ' });
+    createAndAppend('td', trUpdate, { text: repo.updated_at });
   }
 
   function main(url) {
     fetchJSON(url, (err, repos) => {
       const root = document.getElementById('root');
+      const header = createAndAppend('header', root, { id: 'header' });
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -30,20 +30,41 @@
...
+    });
+    const trDesc = createAndAppend('tr', table);
+    createAndAppend('th', trDesc, { text: 'Description: ' });
+    createAndAppend('td', trDesc, { text: repo.description });
+    const trFork = createAndAppend('tr', table);
+    createAndAppend('th', trFork, { text: 'Forks: ' });
+    createAndAppend('td', trFork, { text: repo.forks_count });
+    const trUpdate = createAndAppend('tr', table);
+    createAndAppend('th', trUpdate, { text: 'Last Update: ' });
+    createAndAppend('td', trUpdate, { text: repo.updated_at });
```
remarcmij: I see some repetition of code here. Do you think you could create a function to make your code DRY? Of course, you would need to do something special for the first row.

homework/index.js
```diff
@@ -30,20 +30,41 @@
...
       if (err) {
         createAndAppend('div', root, {
           text: err.message,
           class: 'alert-error',
         });
         return;
       }
-      const ul = createAndAppend('ul', root);
+      const ul = createAndAppend('ul', root, { class: 'repo-containers' });
       repos.forEach(repo => renderRepoDetails(repo, ul));
```
remarcmij: The assignment requires that the repositories are displayed sorted by name. Perhaps you missed that?

homework/style.css
```diff
@@ -1,3 +1,63 @@
 .alert-error {
```
remarcmij: Your app is responsive. 👍

homework/index.js
rabrad: Thank you for the feedback. I will continue to work to meet the required assignment.

homework/index.js
remarcmij: The repetition I see is:

1. Create and append a `<tr>`.
2. Create and append a `<th>` and set its text to `labelText`.
3. Create and append a `<td`> and set its text to `value`.

You do this for the 2nd, 3rd and 4th rows. The only thing that is different is the actual text for the `<th>` and the `<td>`. Whenever you see repetition like this, you have an opportunity to implement the DRY principle. You just need to identify what is different and introduce parameters to the function through which you pass on the variable parts.   

So you could write a function that takes three parameters:

1. the `table` to which to add the row
2. the label text
3. the repository value

```js
  createAndAppend('th', tr, ...);
  createAndAppend('td', tr, ...);
  return tr;
}
```

For instance, for the second row you could call this function like this:

```js
addRow(table, 'Description: ', repo.description);
```

Even though the first row requires an `<a>` instead of simple text, you can still use that function, but you need to do some fixing up afterwards:

```js
const firstRow = addRow(table,  'Repository', '');
createAndAppend('a', firstRow.lastChild, ...);
```

The trick here is to pass an empty string for `value` which we will fix later. This works, because we return the created `tr` as the return value from `addRow()`. This trick with dealing with the first row in a special way will usually only pop up in your head when you look at your code again the next day or so. At least, that's how it usually happens for me.

rabrad: Thank you so much for your feedback, I will try to update my code accordingly.

PR #342, opened 2019-09-18
Homework

ahmthsrvsmz:


remarcmij:
I will wait until you have fixed the ESLint errors before reviewing. Let me know if you run into issues with this.

homework/index.js
```diff
@@ -30,21 +30,53 @@
   }
 
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    createAndAppend('p', ul, { text: `Repository:` });
+    createAndAppend('p', ul, { text: `Description:` });
+    createAndAppend('p', ul, { text: `Forks:` });
+    createAndAppend('p', ul, { text: `Updated:` });
+  }
```
remarcmij: You are rendering the text labels and values in two separate, side-by-side divs. This works as if each row fits on one line. But is if they take more than one line, things get messy, as in the screenshot below, for a small screen device. Do you notice that the values for **Forks** and **Updated** for the **Git** repo are no longer aligned with their labels? Can you find another way of doing this that fixes this problem?

image

homework/index.js
```diff
@@ -30,21 +30,53 @@
...
+  function sortName(a, b) {
+    return a.name.localeCompare(b.name);
   }
 
   function main(url) {
     fetchJSON(url, (err, repos) => {
       const root = document.getElementById('root');
+      createAndAppend('div', root, {
+        text: 'HYF Repositories',
+        class: 'hyfTitle',
```
remarcmij: By convention, CSS identifiers should be **kebab-case**: `hyf-title`.

homework/index.js
```diff
@@ -30,21 +30,53 @@
...
         return;
       }
-      const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+
+      repos.sort(sortName);
+      for (let i = 0; i < repos.length; i++) {
+        const div = createAndAppend('div', root);
+        const insideDiv = createAndAppend('div', div);
+        const insideDiv2 = createAndAppend('div', div);
```
remarcmij: Avoid names that only differ by a meaningless number suffix. Try and invent more descriptive names.

homework/index.js
```diff
@@ -30,21 +30,53 @@
...
+        div.classList.add('rowFlex');
+        insideDiv.classList.add('insideDiv');
```
remarcmij: If line 56 you are using the `options` parameter of `createAndAppend()` to add a `class`. Why not here?

homework/index.js
```diff
@@ -30,21 +30,53 @@
...
           text: err.message,
           class: 'alert-error',
         });
         return;
       }
-      const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+
+      repos.sort(sortName);
+      for (let i = 0; i < repos.length; i++) {
```
remarcmij: Consider using a chained `.forEach()` method here:

```js
repos
  .sort(sortName)
  .forEach(repo => ...
```

homework/style.css
```diff
@@ -1,3 +1,56 @@
+.root {
```
remarcmij: On a wide screen it would better to set a maximum width of your items.

image

homework/index.js
```diff
@@ -30,21 +30,53 @@
...
+
+  function renderRepoDetailsTitle(repo, ul) {
+    createAndAppend('a', ul, {
+      text: repo.name,
+      href: repo.clone_url,
+      target: '_blank',
+    });
+    createAndAppend('p', ul, { text: repo.description });
+    createAndAppend('p', ul, { text: repo.forks });
+    createAndAppend('p', ul, { text: repo.updated_at });
```
remarcmij: Can you think of a way to format the date in a more user-friendly format?

homework/index.js
ahmthsrvsmz: Hi Jim,
Thank you for your feedback. 
I tried to fix problems that you text in your feedback. I changed some codes and now I can't find a solution. It is so complicated. I need more time to work and understand the logic of this code.

remarcmij: Hi @ahmthsrvsmz, I think the complication lies in the HTML structure you have chosen. Trying this fix using JavaScript while keeping the underlying HTML is complicated if not impossible.

The best way to handle this is to plan the HTML you want to generate ahead of actually writing the corresponding JavaScript code.

The most straightforward way to render the repository details is to use an HTML `<table>`, like this:

```html
<ul>
  <li>
    <table>
      <tbody>
        <tr>
          <th class="label">Repository:</th>
          <td>
            <a href="https://github.com/HackYourFuture/alumni" target="_blank"
              >alumni</a
            >
          </td>
        </tr>
        <tr>
          <th class="label">Description:</th>
        </tr>
        <tr>
          <th class="label">Forks:</th>
          <td>10</td>
        </tr>
        <tr>
          <th class="label">Updated:</th>
        </tr>
      </tbody>
    </table>
  </li>
  <li>
    <!-- etc-->
  </li>
</ul>
```

Once you have that plan, you can now write the JavaScript required to generate this HTML.

Does this help to get you on your way?

ahmthsrvsmz: I think this is a real hint. I will try again. Thank you @remarcmij

homework/index.js
```diff
@@ -28,23 +28,56 @@
     });
     return elem;
   }
+  function renderRepoDetails(table, title, repoText, article, repoHref = '') {
```
remarcmij: You have changed the purpose of this function. Its original purpose was to render the complete details of a repository. Now it renders only a table row of those details. I suggest you rename this function `addRow()` and reintroduce a function `renderRepoDetails()`. I give some further hint on that below.

This function for adding a row has become a bit complicated because of the first row that deviates from the other rows, in that it requires an `<a>` tag instead of simple text content. I would advise that you let the caller of this function deal with the exception for the first row and simplify this function to:

```js
function addRow(table, title, repoText) {
  createAndAppend('th', tr, {
    text: title,
  });
  createAndAppend('rd', tr, { text: repoText });
  return tr;
}
```

Notice that this function return the `tr` that was created so that the caller of this function can access this `tr`.

Next, all the code you had in the `.forEach()` inside function `main()` can now go to the original function `renderRepoDetails()` like this:

```js
  const li = createAndAppend('li', ul, { class: 'li-root' });
  const table = createAndAppend('table', li);
  const date = new Date(repo.updated_at);
    href: repo.html_url,
    text: repo.name,
    target: '_blank',
  });
  addRow(table, 'Description:', repo.description, 'td');
  addRow(table, 'Forks:', repo.forks, 'td');
  addRow(table, 'Updated:', date.toLocaleDateString(), 'td');
}
```

As you can see, this function now deals with adding the `<a>` after creating the first row with an empty text content. In this way, the `addRow()` function can be kept simple: it no longer has to deal with the exception for the first row.

As you can see, we let this function add the

homework/index.js
```diff
@@ -28,23 +28,56 @@
...
+      createAndAppend('th', tr, {
+        text: title,
+      });
+      createAndAppend(article, tr, { text: repoText });
+    }
+  }
+
+  function sortName(a, b) {
+    return a.name.localeCompare(b.name);
   }
```
remarcmij: The single-line function is not really needed. You can use an arrow as shown below.

homework/index.js
```diff
@@ -28,23 +28,56 @@
...
           class: 'alert-error',
         });
         return;
       }
-      const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      const ul = createAndAppend('ul', root, {
+        class: 'ul-root',
+      });
+      repos.sort(sortName).forEach(repo => {
```
remarcmij: With the modified `repoDetails()` function, you can now replace lines 72-80 with these three simple lines:

```js
repos
  .forEach(repo => renderRepoDetails(repo, ul));
```

homework/style.css
```diff
@@ -1,3 +1,62 @@
...
+  padding: 0;
+}
+
+table {
+  text-align: left;
+}
+
+table tr th {
+  font-weight: bold;
+  padding-right: 20px;
```
remarcmij: On a mobile device long repo descriptions look better if you align them at top. Try adding this here:

```css
vertical-align: top;
```

homework/index.js
ahmthsrvsmz: My real problem is to create new and hoisting function. This is a really good example. Thank you. I am working on it. 👍

PR #343, opened 2019-09-18
Add JS3-Week1-Homework

EsraZengn:
Here is my JS3-week1 homework.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -29,28 +29,16 @@
     return elem;
   }
 
-  function createTableRows(table, optionsHeader = {}, optionsValue = {}) {
+  function createTableRow(table, header, optionsValue = {}) {
     const tr = createAndAppend('tr', table);
-    createAndAppend('td', tr, optionsHeader);
-    if (optionsHeader.text === 'Repository:') {
-      const td = createAndAppend('td', tr);
-      createAndAppend('a', td, optionsValue);
-    } else {
-      createAndAppend('td', tr, optionsValue);
-    }
+    createAndAppend('td', tr, { text: header, class: 'td-header' });
+    createAndAppend('td', tr, optionsValue);
+    return tr;
   }
 
   function formatDate(dateString) {
-    const date = new Date(dateString);
-    let hours = date.getHours();
-    const ampm = hours >= 12 ? 'PM' : 'AM';
-    hours %= 12;
-    hours = hours === 0 ? 12 : hours;
-    let seconds = date.getSeconds();
-    seconds = seconds < 10 ? `0${seconds}` : seconds;
-    const formatedDate = `${date.getMonth() +
-      1}/${date.getDate()}/${date.getFullYear()}, ${hours}:${date.getMinutes()}:${seconds} ${ampm}`;
-    return formatedDate;
+    const dateTime = new Date(dateString);
+    return dateTime.toLocaleString();
   }
 
   function renderRepoDetails(repo, ul) {
@@ -62,31 +50,16 @@
     } else {
       description = 'No description';
     }
-    const headerClass = 'td-header';
-    createTableRows(
-      table,
-      { text: 'Repository:', class: headerClass },
-      {
-        text: repo.name,
-        href: repo.html_url,
-        target: '_blank',
-      },
-    );
-    createTableRows(
-      table,
-      { text: 'Description:', class: headerClass },
-      { text: description },
-    );
-    createTableRows(
-      table,
-      { text: 'Forks:', class: headerClass },
-      { text: repo.forks },
-    );
-    createTableRows(
-      table,
-      { text: 'Updated', class: headerClass },
-      { text: formatDate(repo.updated_at) },
-    );
+    const firstRow = createTableRow(table, 'Repository:');
+    createAndAppend('a', firstRow.lastChild, {
+      text: repo.name,
+      href: repo.html_url,
+      target: '_blank',
+    });
+
+    createTableRow(table, 'Description:', { text: description });
+    createTableRow(table, 'Forks:', { text: repo.forks });
+    createTableRow(table, 'Updated', { text: formatDate(repo.updated_at) });
   }
 
   function main(url) {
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -29,22 +29,97 @@
...
+  function createTableRows(table, optionsHeader = {}, optionsValue = {}) {
+    const tr = createAndAppend('tr', table);
+    createAndAppend('td', tr, optionsHeader);
+    if (optionsHeader.text === 'Repository:') {
+      const td = createAndAppend('td', tr);
+      createAndAppend('a', td, optionsValue);
+    } else {
+      createAndAppend('td', tr, optionsValue);
+    }
+  }
```
remarcmij: Excellent that you saw an opportunity to make your code DRY by extraction the code repetition into a function that you can call repeatedly. This is a good first version, that I could imagine having written myself. But then, pondering this code the next day, I would probably set about doing another round of refactoring.

The idea of writing a function like this is to maximise the work that you can do inside the function and minimise the work that the caller still needs to do. For instance, I observe that all header text gets assigned the CSS class `td-header`. So you could hard-code that inside the function body and simply pass the text value for the header rather than an object:

```js
function createTableRows(table, header, optionsValue = {}) {
  createAndAppend('td', tr, { text: header, class: 'td-header' });
  if (header === 'Repository:') {
    const td = createAndAppend('td', tr);
    createAndAppend('a', td, optionsValue);
  } else {
    createAndAppend('td', tr, optionsValue);
  }
}
```

Then, because you need to do some special for the first row, you test whether the header `Repository:` is passed as the header text. This looks a bit awkward to me. It would probably take me another day to think of a more elegant solution. So the next day we could do a further refactoring. We can simplify the function to this (note the `return` statement): 

```js
function createTableRows(table, header, value) {
  createAndAppend('td', tr, { text: header, class: 'td-header' });
  createAndAppend('td', tr, { text: value });
  return tr;
}
```

But how then would we deal with that special first row? Answer: we can create the header and an empty value and stuff in the `<a>` tag afterwards:

```js
const firstRow = createTableRows(table, 'Repository:', '');
  text: repo.name,
  href: repo.html_url,
  target: '_blank',
});
createTableRows(table, 'Description:', description);
createTableRows(table, 'Forks:', repo.fork);
createTableRows(table, 'Updated', formatDate(repo.updated_at));
```

Finally, since this function creates a single row (although you call it more than once), I would rename this function to `createTableRow` (drop the plural 's').

When I did this homework myself it took me several iterations over a couple of days to land at this final version.

homework/index.js
```diff
@@ -29,22 +29,97 @@
...
+    let hours = date.getHours();
+    const ampm = hours >= 12 ? 'PM' : 'AM';
+    hours %= 12;
+    hours = hours === 0 ? 12 : hours;
+    let seconds = date.getSeconds();
+    seconds = seconds < 10 ? `0${seconds}` : seconds;
+    const formatedDate = `${date.getMonth() +
+      1}/${date.getDate()}/${date.getFullYear()}, ${hours}:${date.getMinutes()}:${seconds} ${ampm}`;
+    return formatedDate;
+  }
```
remarcmij: There is no need to do this formatting yourself. The `Date` object has many built-in ways to format a date. Another student posed a question about this in slack:

https://hackyourfuture.slack.com/archives/CL795TNFQ/p1568802816003000

I would just use `.localLocaleString()`. This uses the date format that the user has chosen for his/her computer.

homework/index.js
```diff
@@ -29,22 +29,97 @@
...
           class: 'alert-error',
         });
         return;
       }
+
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+
+      repos
+        .sort((repo, nextRepo) => {
```
remarcmij: In the assignment we gave the hint to use `.localeCompare`. Your method works fine if all names just use the regular letters. However, as soon as they include letters with diacritics, such as in French, German, Dutch or Turkish the names are not sorted in the (dictionary) order you would expect.

You may want to try this code, comparing your sort method agains `.localeCompare()`. I have taken the surnames of the students in class23 but added another name, viz. **Öbul** which I would expect to be sorted just after **Obul**. With `.localeCompare()` (which takes into account the 'locale' that is set up in your computer) this works fine. Your sort method however places **Öbul** after **Zengin**.

```js
'use strict';

{
  const users = [
    'Hüsrev',
    'Raad',
    'Kanatsiz',
    'Ekinci',
    'Ince',
    'Ajaj',
    'Obul',
    'Öbul',
    'Zengin',
    'Karatas',
    'Ersaru',
    'Tercan',
    'Alnakib',
    'Grabetskaia',
    'Sönmez',
    'Akkuz',
  ];

  const esraSorted = users.slice().sort((a, b) => {
    const aUpper = a.toUpperCase();
    const bUpper = b.toUpperCase();
    if (aUpper < bUpper) {
      return -1;
    }
    if (aUpper > bUpper) {
      return 1;
    }
    return 0;
  });

  const localeSorted = users.slice().sort((a, b) => a.localeCompare(b));

  console.log(esraSorted);
  console.log(localeSorted);
}
```

homework/style.css
```diff
@@ -1,3 +1,60 @@
+* {
```
remarcmij: On a mobile phone you are wasting a lot of screen real estate. Better trim down on the left and right margins.

image

Also, your error message box is a bit cramped:

image

homework/index.js
EsraZengn: Well, my first version was also like this `createTableRows(table, header, options={});` but in order to add a tag I had to change it, which I see now was not very efficient at all. I guess I should have spend more time on it before sending the homework. It's a very nice solution to append `a` to `firstRow.lastChild`. I improved and implemented my code as you suggested. Thank you very much for the awesome feedback.

homework/index.js
EsraZengn: Very efficient indeed! I can't believe I missed this.. Thank you very much.

homework/index.js
EsraZengn: I just realized I missed the tip and thank you very much for explaining why my function is not very useful. I will try to be more careful reading instructions and writing better functions. Many thanks.

homework/style.css
EsraZengn: I made the necessary corrections as you suggested. Thanks a lot.

homework/index.js
remarcmij: Your solution was acceptable to me. But sometimes taking a little more time, if available, will bring new insights. Or somebody else looking at your code may trigger that. Don't worry if it's not 'perfect' this first time around.

PR #345, opened 2019-09-19
JavaScript-3-Week-1-homework-Ann-Grabetski

Ann-Grabetski:


homework/index.js
```diff
@@ -29,8 +29,42 @@
     return elem;
   }
 
-  function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+  function renderRepoDetails(repo) {
```
remarcmij: This is a rather large function. It is difficult to understand what is happening here. It would be better if you break this function up into smaller function, each with a descriptive name to indicate what it is doing.

homework/index.js
```diff
@@ -29,8 +29,42 @@
...
+    const reposList = document.getElementById('reposList');
+    const repoItem = createAndAppend('li', reposList);
+    const commonBlock = createAndAppend('div', repoItem, {
+      class: 'commonBlock',
```
remarcmij: By convention, CSS classed should be in **kebab-case**: `common-block`.  This holds true for a couple of places in your code (classes, ids).

homework/index.js
```diff
@@ -29,8 +29,42 @@
...
+    });
+    const dataTypeList = createAndAppend('ul', commonBlock, {
```
remarcmij: Your UI uses two unordered lists, placed side by side, one containing labels and the other containing the repository values. When viewed on a mobile phone the long repository values are clipped:

image

This is undesired. But if you somehow manage to let the repository values wrap over multiple lines you run into another problem because of your chosen organisation. Wrapped lines for the description will push the values for **Forks** and **Updated** down and they will no longer align with their labels.

Can you try another structure of your HTML that solves these problems, perhaps an HTML table?

homework/index.js
```diff
@@ -29,8 +29,42 @@
...
+    if (repo.description === null) {
+      createAndAppend('li', dataList, {
+        text: 'No Description',
+        class: 'alert-no-description',
+      });
+    } else {
+      createAndAppend('li', dataList, { text: repo.description });
+    }
+    createAndAppend('li', dataList, { text: repo.forks });
+    createAndAppend('li', dataList, { text: repo.updated_at });
```
remarcmij: Can you format this date in a more user-friendly format?

homework/index.js
```diff
@@ -29,8 +29,42 @@
...
+      class: 'commonBlock',
+    });
+    const dataTypeList = createAndAppend('ul', commonBlock, {
+      class: 'dataTypeList',
+    });
+    createAndAppend('li', dataTypeList, { text: 'Repository:' });
+    createAndAppend('li', dataTypeList, { text: 'Description:' });
+    createAndAppend('li', dataTypeList, { text: 'Forks:' });
+    createAndAppend('li', dataTypeList, { text: 'Updated:' });
+    const dataList = createAndAppend('ul', commonBlock, { class: 'dataList' });
```
remarcmij: I couldn't find a `dataList` class in your CSS file.

homework/index.js
```diff
@@ -43,8 +77,13 @@
...
       }
-      const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      createAndAppend('div', root, {
+        id: 'title',
+        text: 'HYF Repositories',
+      });
+      createAndAppend('ul', root, { id: 'reposList' });
+      sortRepos(repos);
+      repos.forEach(repo => renderRepoDetails(repo));
```
remarcmij: You do the sort in-line and then chain with `forEach()`:

```js
repos
  .sort((a, b) => a.name.localeCompare(b.name));
  .forEach(repo => ...);
```

Ann-Grabetski: I did not quite understand this remark. Do you mean that It does not need to make a separate function for sorting and just sort in a line is enough?

homework/index.js
Ann-Grabetski: I am not using it for now, but I suggest that this class may be useful in future. But anyway, with a new structure it will be no sense in this class.

homework/index.js
remarcmij: I have the unfortunate habit of sometimes leaving out words when I type. This is a typical case:

> You **can** do the sort in-line and then chain with `.forEach()`:

homework/index.js
remarcmij: There is a principle in software development that goes by the acronym YAGNI:

>Yagni originally is an acronym that stands for "You Aren't Gonna Need It". It is a mantra from ExtremeProgramming that's often used generally in agile software teams. It's a statement that some capability we presume our software needs in the future should not be built now because "you aren't gonna need it".

Read more about it here:

https://martinfowler.com/bliki/Yagni.html

homework/index.js
Ann-Grabetski: Ah, ok, I got it.

homework/index.js
Ann-Grabetski: Okay. Thank you for the explanation!

homework/index.js
Ann-Grabetski: I did my best in UpDate.

homework/index.js
Ann-Grabetski: Good to know! Thank you

homework/index.js
Ann-Grabetski: I have changed all of it.

homework/index.js
Ann-Grabetski: Done

homework/index.js
```diff
@@ -29,22 +29,74 @@
...
+    const row = createAndAppend(`tr`, parent);
+    createAndAppend(`th`, row, { text: `${title}:` });
+    let data;
+    if (typeof content === `string` || typeof content === `number`) {
+      createAndAppend(`td`, row, { text: `${content}` });
+    } else {
+      data = createAndAppend(`td`, row);
+      data.appendChild(content);
+    }
+  }
```
remarcmij: This function and well as the previous (`createTag`) function is a good attempt at making your code DRY and extract code segments from a larger function into a couple of smaller functions. However, the resulting code is unnecessarily complicated, because of the `if` statement. In the `buildTableRow` function (good name 👍), you are trying to deal with exceptions, where content is not a text string or a number but an object. This makes this function difficult to understand. It is better to let the caller of this function handle the exceptions. You could then simplify this function to:

```js
function buildTableRow(title, parent, content) {
  const row = createAndAppend(`tr`, parent);
  createAndAppend(`th`, row, { text: `${title}:` });
  createAndAppend(`td`, row, { text: `${content}` });
  return row;
}
```

Note the addition of `return row;`. This allows the caller of the function to further work with the returned `tr`.

homework/index.js
```diff
@@ -29,22 +29,74 @@
...
+  function renderRepoDetails(repo, parent) {
+    const detailsDiv = createAndAppend(`div`, parent, { class: `details` });
+    const table = createAndAppend(`table`, detailsDiv);
+
+    const repoLink = createTag(`a`, {
+      href: repo.html_url,
+      target: '_blank',
+      text: repo.name,
+    });
+    buildTableRow(`Repository`, table, repoLink);
```
remarcmij: Instead of passing a pre-made `<a>` tag to the `buildTableRow` function, you can add it afterwards (using the simplified `buildTableRow` function), like this:

```js
const firstRow = buildTableRow(`Repository`, table, '');
createAndAppend(`a`, firstRow.lastChild, {
  href: repo.html_url,
  target: '_blank',
  text: repo.name,
});
```

homework/index.js
```diff
@@ -29,22 +29,74 @@
...
+
+    const description = repo.description
+      ? { text: repo.description }
+      : { text: 'No Description', class: 'alert-no-description' };
+
+    const repoDescription = createTag(`span`, description);
+    buildTableRow(`Description`, table, repoDescription);
```
remarcmij: To deal with the Description you could use a so-called _short-circuit_ expression:

```js
buildTableRow(`Description`, table, repo.description || 'No Description');
```

Of course, this doesn't use the extra CSS class you were adding in the case of a missing description. If you would still like to add that, you can do the same trick as we did for the first row:

```js
const descriptionRow = buildTableRow(`Description`, table, '');
const description = repo.description
  ? { text: repo.description }
  : { text: 'No Description', class: 'alert-no-description' };
createAndAppend('span', descriptionRow.lastChild, description);
```

Note that we no longer need your `createTag` function, which looked like a stripped-down version of `createAndAppend`.

homework/index.js
Ann-Grabetski: Wow! I didn't think about this.

PR #346, opened 2019-09-19
JS3 Week1 Hw

FarukTercan:


homework/index.js
```diff
@@ -23,28 +23,73 @@
...
   }
 
+  function changeDateTimeFormat(dateTime) {
+    const timeFormat = new Date(dateTime);
+    return timeFormat.toLocaleString();
+  }
+
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const list = createAndAppend('li', ul, { class: 'topClass' });
```
remarcmij: By convention, CSS identifiers should be in **kebab-case**: `top-class`.

homework/index.js
```diff
@@ -23,28 +23,73 @@
...
+    const name = createAndAppend('li', list, {
```
remarcmij: You can not have an `li` inside another `li`. This is invalid HTML (although the browser will try and make the best of it). An `li` can only be a child of an `ol` or `ul` (or the less used `menu`). See:

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/li

This is the HTML you have generated:

image

homework/index.js
```diff
@@ -23,28 +23,73 @@
...
+      text: 'Updated :',
+      class: 'keys',
+    });
+
+    createAndAppend('a', name, {
+      text: repo.name,
+      href: repo.html_url,
+      class: 'values',
+    });
+    createAndAppend('a', description, {
```
remarcmij: You should only use `<a>` tags if you have a corresponding URL in an `href` attribute. Without it, it doesn't link to anything.

homework/index.js
```diff
@@ -23,28 +23,73 @@
...
+      text: repo.description,
+      class: 'values',
+    });
+    createAndAppend('a', fork, { text: repo.forks, class: 'values' });
+    createAndAppend('a', time, {
+      text: changeDateTimeFormat(repo.updated_at),
```
remarcmij: The date has a user-friendly format: 👍

homework/index.js
```diff
@@ -23,28 +23,73 @@
...
         createAndAppend('div', root, {
           text: err.message,
           class: 'alert-error',
         });
         return;
       }
+
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      repos.sort(sortAlpha).forEach(repo => renderRepoDetails(repo, ul));
```
remarcmij: Repositories are sorted by name: 👍

homework/style.css
```diff
@@ -1,3 +1,48 @@
 .alert-error {
```
remarcmij: Your UI layout looks a bit sloppy. The repository values are not ligned out vertically, the colons (`:`) need proper spacing and long repository values wrap to the extreme left margin, especially visible on a mobile device. Maybe you can solve a couple of layout issues by using an HTML table for each repo?

image

On wide-screen monitor your UI looks too streched out. Perhaps set a max width?

image

homework/index.js
FarukTercan: Hi @remarcmij, thanks for your feedback. 
I used a table as you suggested. Thus, I removed the invalid li tags and also unlinked <a> tags.  To minimize table creating function, I tried to use your hint from other comments about the first row and <a> tag connection.

homework/style.css
FarukTercan: About the stretch issue, I gave percentage value to widths and margins.

homework/index.js
```diff
@@ -29,22 +29,51 @@
...
+  function makeTable(table, header, value) {
+    const tr = createAndAppend('tr', table, { class: 'tr' });
+    createAndAppend('th', tr, { text: header, class: 'keys' });
+    createAndAppend('td', tr, { text: value, class: 'values' });
+    return tr;
+  }
+
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const table = createAndAppend('table', ul, { class: 'table' });
```
remarcmij: This generates invalid HTML. A `ul` can only have  `li` tags as its children. You need to create an `li` tag and append it to the `ul`. Then, create the `table` tag and append it to the `li`.

homework/index.js
```diff
@@ -29,22 +29,51 @@
     return elem;
   }
 
+  function changeDateTimeFormat(dateTime) {
+    const timeFormat = new Date(dateTime);
+    return timeFormat.toLocaleString();
+  }
+  function makeTable(table, header, value) {
```
remarcmij: You are not making a table here. You are creating or adding a table row. This should be reflected in the function name.

(Also, place a blank line before a function name to make it stand out in the code. This is like placing blanks lines between paragraphs in a piece of text.)

homework/index.js
remarcmij: You can also combine `.sort()1` and `.forEach()` in a chain and do away with the separate `sortAlpha` function:

```js
repos
```

PR #347, opened 2019-09-19
add Js3 Week1 Homework

mehmetekinci:


homework/index.js
```diff
@@ -29,13 +29,54 @@
     return elem;
   }
 
+  function makeText(stringDateAndTime) {
```
remarcmij: The name `makeText` is a bit vague and does not accurately describe what this function does. Perhaps `formatDate` would come closer?

homework/index.js
```diff
@@ -29,13 +29,54 @@
...
+    const dt = new Date(stringDateAndTime);
+    return dt.toLocaleString();
+  }
+
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const li = createAndAppend('li', ul, { class: 'repos' });
+    const table = createAndAppend('table', li, { class: 'table' });
```
remarcmij: An HTML table is an appropriate choice. But there some repetition of code here that begs to be put in a separate, reusable function. The repetition I see is: create a `tr` with a `td` child with some label text and another `td` child with a repository property text value. So I could imagine a function, say `addRow`, as follows:

```js
  const tr = ...
  ...
  return tr;
}
```

The first row requires some special handling as it should contain a `<a>` tag instead of simple text. I'll give you a hint how this can be done:

```js
createAndAppend('a', firstRow.lastChild, ...));
```

homework/index.js
```diff
@@ -29,13 +29,54 @@
...
+    const tr4 = createAndAppend('tr', table);
+
+    createAndAppend('td', tr1, { text: 'Repository :', class: 'tableTitle' });
+    const link = createAndAppend('td', tr1, { class: 'tableText' });
+    createAndAppend('a', link, { text: repo.name, href: repo.html_url });
+    createAndAppend('td', tr2, { text: 'Description :', class: 'tableTitle' });
+    if (repo.description === null) {
+      createAndAppend('td', tr2, {
+        text: 'No Information',
+        class: 'tableText',
```
remarcmij: By convention, CSS classes should be in **kebab-case**: `table-text`

homework/index.js
```diff
@@ -29,13 +29,54 @@
...
+      createAndAppend('td', tr2, {
+        text: repo.description,
+        class: 'tableText',
+      });
+    }
+    createAndAppend('td', tr3, { text: 'Forks :', class: 'tableTitle' });
+    createAndAppend('td', tr3, { text: repo.forks, class: 'tableText' });
+    createAndAppend('td', tr4, { text: 'Updated :', class: 'tableTitle' });
+    createAndAppend('td', tr4, {
+      text: makeText(repo.updated_at),
```
remarcmij: Date presented in a user-friendly format: 👍

homework/index.js
```diff
@@ -44,7 +85,9 @@
         return;
       }
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      repos.sort(sortReposByNameProp).forEach(repo => {
```
remarcmij: Repositories sorted by name: 👍

homework/style.css
```diff
@@ -1,3 +1,32 @@
...
+  padding: 20px 10px;
+}
+
+.header {
+  background-color: rgb(63, 81, 181);
+  border: 2px solid rgb(229, 229, 229);
+  padding: 20px 10px;
+  color: white;
+}
+#root {
```
remarcmij: On a wide-screen monitor you use the full width of the screen, which makes your UI look too stretched out. Perhaps set a max width?

image

mehmetekinci:
Hi @remarcmij I am sorry for late respond. Yesterday I was busy with my car in garage. first of all I appreciate for your helpful feedback. I learn a lot in last class.I hope I can make more use of your experiences. I  improved my codes. I know it is not the best but my best. Thnaks

homework/index.js
```diff
@@ -29,41 +29,37 @@
...
+    });
+
+    if (link) {
+      createAndAppend('a', secondTdElm, { href: link, text: value });
+    } else {
+      secondTdElm.innerText = value;
+    }
+
+    return trElm;
+  }
```
remarcmij: This is a nice improvement which makes your code DRY. An alternative to adding a fourth parameter for the link and using an `if` statement inside the function body is to take the exception for the first row out of this function and make it a responsibility of the caller. That was in my hint in my initial feedback. Your function could then simply look like this:

```js
function addRepoDetailsInfo(listElm, title, value, link) {
  const trElm = createAndAppend('tr', listElm);
  createAndAppend('td', trElm, {
    class: 'titles',
    text: title,
  });
  createAndAppend('td', trElm, {
    class: 'values',
    text: value,
  });
  return trElm;
}
```

And you would call it like this:

```js
const firstRow = addRepoDetailsInfo(table, 'Repository :', '');
  text: repo.name,
  href: repo.html_url,
});
addRepoDetailsInfo(table, 'Description :', repo.description);
addRepoDetailsInfo(table, 'Forks :', repo.forks);
addRepoDetailsInfo(table, 'Updated :', formatDate(repo.updated_at));
```

Having said that, your solution looks fine too.

homework/index.js
```diff
@@ -29,20 +29,23 @@
...
+        class: 'values',
+        text: 'No Information',
+      });
     } else {
-      secondTdElm.innerText = value;
+      createAndAppend('td', trElm, {
+        class: 'values',
+        text: value,
+      });
     }
```
remarcmij: Instead of an `if` statement you can use a ternary expression to further simplify your code:

```js
function addRepoDetailsInfo(listElm, title, value) {
  const trElm = createAndAppend('tr', listElm);
  createAndAppend('td', trElm, {
    class: 'titles',
    text: title,
  });
  createAndAppend('td', trElm, {
    class: 'values',
    text: value !== null ? value : 'No Information',
  });
  return trElm;
}
```

mehmetekinci:
Again hi @remarcmij i am really feeling good with your feedbacks when I make some mistake in my homework. Thank you for all useful and helpful feedbacks

homework/style.css
```diff
@@ -1,3 +1,35 @@
...
+
+.header {
+  background-color: rgb(63, 81, 181);
+  border: 2px solid rgb(229, 229, 229);
+  padding: 20px 10px;
+  color: white;
+}
+
+#root {
+  max-width: 800px;
```
remarcmij: Your app would look better if the content was centered horizontally. You only need to add this to `#root`:

```css
margin: 0 auto;
```

homework/index.js
mehmetekinci: Thank you so much @remarcmij . I also made last changes.

PR #349, opened 2019-09-19
JS3-WEEK1

malnaqeeb:
Hi @remarcmij, Actually I'm not fully satisfied with what I have done, but I submitted it to match the deadline and also to get feedback to improve it.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -28,48 +28,43 @@
     });
     return elem;
   }
+
+  // 1- create function to add row
+  function createRow(table, label, content) {
+    const row = createAndAppend('tr', table);
+    createAndAppend('th', row, { text: label });
+    createAndAppend('td', row, { text: content });
+    return row;
+  }
+
+  // 2-  append details of repositories to list
   function renderRepoDetails(repo, ul) {
     const li = createAndAppend('li', ul, { class: 'item' });
     const table = createAndAppend('table', li);
-
-    // add Repository name
-    const row1 = createAndAppend('tr', table);
-    const repoName = createAndAppend('th', row1);
-    repoName.innerText = 'Repository:';
-    const repoNameContent = createAndAppend('td', row1);
-    createAndAppend('a', repoNameContent, {
+    // 2-1 add repository name
+    const repoName = createRow(table, 'Repository:');
+    createAndAppend('a', repoName.lastChild, {
       href: repo.html_url,
       text: repo.name,
+      target: '_blank',
     });
-
-    // add description
-    const row2 = createAndAppend('tr', table);
-    const repoDescription = createAndAppend('th', row2);
-    repoDescription.innerText = 'Description:';
-    const repoDescriptionContent = createAndAppend('td', row2);
+    // 2-2 add description
     if (repo.description) {
-      repoDescriptionContent.innerText = repo.description;
+      createRow(table, 'Description:', repo.description);
     } else {
-      repoDescriptionContent.innerText = 'N/A';
+      createRow(table, 'Description:', 'N/A');
     }
-
-    // add Forks
-    const row3 = createAndAppend('tr', table);
-    const repoForks = createAndAppend('th', row3);
-    repoForks.innerText = 'Forks:';
-    const repoForksContent = createAndAppend('td', row3);
-    repoForksContent.innerText = repo.forks_count;
-
-    // add updated
-    const row4 = createAndAppend('tr', table);
-    const repoUpdate = createAndAppend('th', row4);
-    repoUpdate.innerText = 'Updated:';
-    const repoUpdateContent = createAndAppend('td', row4);
-    repoUpdateContent.innerText = new Date(repo.updated_at).toLocaleString();
+    // 2-3 add Forks
+    createRow(table, 'Forks:', repo.forks);
+    // 2-4 add updated
+    createRow(table, 'Updated:', new Date(repo.updated_at).toLocaleString());
   }
+
+  // 3- Create function to sort the repositories
   function sorting(part1, part2) {
     return part1.name.localeCompare(part2.name);
   }
+
   function main(url) {
     fetchJSON(url, (err, repos) => {
       const root = document.getElementById('root');
@@ -83,7 +78,6 @@
         return;
       }
       const ul = createAndAppend('ul', root);
-
       repos.sort(sorting).forEach(repo => renderRepoDetails(repo, ul));
     });
   }
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,3 +1,5 @@
+/* eslint-disable prefer-template */
+
 'use strict';
 
 {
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -28,9 +30,37 @@
     });
     return elem;
   }
-
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    // eslint-disable-next-line prefer-template
+    const li = createAndAppend('li', ul, { id: 'item' });
+    const repoName = createAndAppend('p', li);
+    repoName.innerHTML = "<span class = 'nameItem'>Repository:</span> ";
```
remarcmij: Using `.innerHTML` to create HTML element this way is inconsistent with the rest of the code that uses `createAndAppend()`. Please use only `createAndAppend()` to create elements in this homework.

homework/index.js
```diff
@@ -28,9 +30,37 @@
     });
     return elem;
   }
-
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    // eslint-disable-next-line prefer-template
+    const li = createAndAppend('li', ul, { id: 'item' });
```
remarcmij: Your generated HTML contains multiple occurrences of the `item` **id**. This violates the HTML rule that **id** attributes should be unique, i.e. occurs once only in an HTML document.

homework/index.js
```diff
@@ -28,9 +30,37 @@
...
+        "<span class = 'nameItem'>Description: </span> N/A";
+    }
+    const repoForks = createAndAppend('p', li, { class: 'paragraph' });
+    repoForks.innerHTML =
+      "<span class = 'nameItem'>Fork: </span>" + repo.forks_count;
+    const repoDate = createAndAppend('p', li);
+    repoDate.innerHTML =
+      "<span class = 'nameItem'>Updated: </span> " +
+      new Date(repo.updated_at).toLocaleString();
+  }
```
remarcmij: I think using an HTML table would produce a nice UI. It would vertically align all repository values and also ensure that longer values do not wrap to the extreme left margin as happens in your case if you decrease the screen width:

image

homework/index.js
```diff
@@ -28,9 +30,37 @@
...
+
+  const createTitle = document.getElementById('root');
+  const title = createAndAppend('h2', createTitle, { class: 'paragraph' });
+  title.innerText = 'HYF Repositories';
+  function sorting(part1, part2) {
```
remarcmij: Use a blank line before a function definition to make them stand out from the rest of the code. This is similar to separating paragraphs in a piece of text with a blank line to improve readability.

homework/index.js
```diff
@@ -44,6 +74,7 @@
         return;
       }
       const ul = createAndAppend('ul', root);
+      repos.sort(sorting);
       repos.forEach(repo => renderRepoDetails(repo, ul));
```
remarcmij: You can also do the sort in-line using an arrow function and then chain the `.forEach()`:

```js
repos
  .forEach(repo =>  renderRepoDetails(repo, ul));
```

homework/index.js
```diff
@@ -28,9 +30,37 @@
...
+    } else {
+      repoDescription.innerHTML =
+        "<span class = 'nameItem'>Description: </span> N/A";
+    }
+    const repoForks = createAndAppend('p', li, { class: 'paragraph' });
+    repoForks.innerHTML =
+      "<span class = 'nameItem'>Fork: </span>" + repo.forks_count;
+    const repoDate = createAndAppend('p', li);
+    repoDate.innerHTML =
+      "<span class = 'nameItem'>Updated: </span> " +
```
remarcmij: By convention, CSS identifiers should be **kebab-case**: `name-item`

malnaqeeb:
Hi @remarcmij , I have almost re-build the whole PR  :)

homework/index.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable prefer-template */
```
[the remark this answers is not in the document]
malnaqeeb: I don't know why that didn't work, mabye i'm gonna discuss that with u tommorow

homework/index.js
malnaqeeb: I was looking for a way to make change the font to be bold, but now i don't need it any more

homework/index.js
malnaqeeb: Yes you are right, I was confused about the whole homework and forget this point

homework/index.js
malnaqeeb: I did :+1:

homework/index.js
malnaqeeb: done :+1:

homework/index.js
malnaqeeb: done :+1:

homework/index.js
malnaqeeb: Don't know what I was thinking :(

homework/index.js
```diff
@@ -28,14 +28,53 @@
...
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
-  }
+    const li = createAndAppend('li', ul, { class: 'item' });
+    const table = createAndAppend('table', li);
+
+    // add Repository name
+    const row1 = createAndAppend('tr', table);
+    const repoName = createAndAppend('th', row1);
+    repoName.innerText = 'Repository:';
```
remarcmij: You can set the text content directly with the `createAndAppend()` function like this (it also allows you to get rid of the variable):

```js
createAndAppend('th', row, { text: 'Repository:' });
```

homework/index.js
```diff
@@ -28,14 +28,53 @@
...
+    const repoForksContent = createAndAppend('td', row3);
+    repoForksContent.innerText = repo.forks_count;
+
+    // add updated
+    const row4 = createAndAppend('tr', table);
+    const repoUpdate = createAndAppend('th', row4);
+    repoUpdate.innerText = 'Updated:';
+    const repoUpdateContent = createAndAppend('td', row4);
+    repoUpdateContent.innerText = new Date(repo.updated_at).toLocaleString();
+  }
```
remarcmij: I see some repetition here: create a `tr`, add a `th` child with some label text and a `td` child with a repository property value. Can you write a function for this and call it for each row to be created?

The function could look like this:

```js
  const tr = ...;
  createAndAppend('th', table, ....);
  createAndAppend('td, table, ... );
  return tr;
}
```

You could use this function to create the table rows, instead of repeating the code four times.

For the first (that is to contain an `<a>` tag instead of simple text) you would need to do something special after calling the `addRow()` function:

```js
  href: repo.html_url,
  text: repo.name,
  target: "_blank",
});
```

Note that the third parameter for `addRow` is an empty string in this case, because we don't actually want to set the text content.

homework/index.js
remarcmij: I still see function definitions that are not preceded by a blank line, so it's not done.

homework/index.js
malnaqeeb: Thanks alot for this function, it's very hulpful :)

homework/index.js
malnaqeeb: done :+1:

malnaqeeb:
Thanks a lot @remarcmij, I appreciate these comments.

homework/index.js
```diff
@@ -29,13 +29,47 @@
...
+      href: repo.html_url,
+      text: repo.name,
+      target: '_blank',
+    });
+    // 2-2 add description
+    if (repo.description) {
+      createRow(table, 'Description:', repo.description);
+    } else {
+      createRow(table, 'Description:', 'N/A');
+    }
```
remarcmij: Final tip: no need for an `if` statement.

```js
// 2-2 add description
createRow(table, 'Description:', repo.description || 'N/A');
```

The value of the expression ` repo.description || 'N/A'` is the value of `repo.description` if that value is truthy, or `N/A` if it is falsy (which includes `null`. `undefined` and the empty string `''`).

PR #350, opened 2019-09-19
Week1

haha1906:
Week 1 Homework

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -26,11 +26,52 @@
         elem.setAttribute(key, value);
       }
     });
+
     return elem;
   }
 
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const li = createAndAppend('li', ul, {
+      class: 'container',
+    });
+
+    if (repo.name != null) {
+      createAndAppend('li', li, {
+        text: 'Repository :',
+        class: 'leftHandSide',
+      });
+      const repositoryName = createAndAppend('li', li, {
+        class: 'rightHandSide',
+      });
+      createAndAppend('a', repositoryName, {
+        href: repo.html_url,
+        target: '_blank',
+        text: repo.name,
+      });
+    }
+    function keyToRead(key) {
+      let newValue = key;
+      for (let i = 0; i < key.length; i++) {
+        if (key[i] === '_') {
+          newValue = newValue.substr(0, i) + ' ' + newValue.substr(i + 1);
+        }
+      }
+      return newValue[0].toUpperCase() + newValue.slice(1);
+    }
+
+    const renderKeys = ['description', 'forks', 'updated_at'];
+    renderKeys.forEach(key => {
+      if (repo[key] != null) {
+        createAndAppend('li', li, {
+          text: keyToRead(key) + ' :',
+          class: 'leftHandSide',
+        });
+        createAndAppend('li', li, {
+          text: repo[key],
+          class: 'rightHandSide',
+        });
+      }
+    });
   }
 
   function main(url) {
@@ -43,8 +84,12 @@
         });
         return;
       }
+      createAndAppend('h2', root, { text: 'HYF Repositories' });
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      const reposSorted = repos.sort(function(a, b) {
+        return a.name.localeCompare(b.name);
+      });
+      reposSorted.forEach(repo => renderRepoDetails(repo, ul));
     });
   }
 
--- a/homework/index.js
+++ b/homework/index.js
@@ -1,98 +1,81 @@
 'use strict';
 
-{
-  function fetchJSON(url, cb) {
-    const xhr = new XMLHttpRequest();
-    xhr.open('GET', url);
-    xhr.responseType = 'json';
-    xhr.onload = () => {
-      if (xhr.status >= 200 && xhr.status <= 299) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
-  }
-
-  function createAndAppend(name, parent, options = {}) {
-    const elem = document.createElement(name);
-    parent.appendChild(elem);
-    Object.entries(options).forEach(([key, value]) => {
-      if (key === 'text') {
-        elem.textContent = value;
-      } else {
-        elem.setAttribute(key, value);
-      }
-    });
-    return elem;
-  }
-
-  function renderRepoDetails(repo, ul) {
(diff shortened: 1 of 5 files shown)


haha1906:
I have faced some 'fatal error's when I was trying to push my homework. Finally I can see on the HackYourFuture repository. But I am not sure that I could upload it yet.

haha1906:
Sir, I have examined some other homeworks and see some mistakes of mine. I can say some of them before you maybe.

firstly, I should have begun this work before and spent more time discussing. I opened the index.js everyday and looked sadly and long. I apologize for that.

'Updated at' should be more meaningfull value using some easy functions such as `.localLocaleString()` 

I have some errors with ESLint. Because I wanted to put space between 'Updated' and 'at' words, I had to use two `' '` that contains one space. It's below:

```
        for (let i = 0; i < key.length; i++) {
        if (key[i] === '_') {
          newValue = newValue.substr(0, i) + ' ' + newValue.substr(i + 1);
        }
```

I used one for loop to look into texts. I couldn't do it with forEach(). As far as I know we could use forEach() for array, can we use it for strings either?

homework/index.js
```diff
@@ -30,7 +30,47 @@
   }
 
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const li = createAndAppend('li', ul, {
+      class: 'container',
+    });
+
+    if (repo.name != null) {
+      createAndAppend('li', li, {
```
remarcmij: An `li` tag can only have a `ul`, `ol` or `menu` tag as its parent. You can not define as a child of another `li`. This will result in invalid HTML. The browser will try and make the best of it anyway, but it is still an error. This is the HTML you generate:

```html
<div id="root">
  <h2>HYF Repositories</h2>
  <ul>
    <li class="container">
      <li class="leftHandSide">Repository :</li>
      <li class="rightHandSide"><a href="https://github.com/HackYourFuture/alumni" target="_blank">alumni</a></li>
      <li class="leftHandSide">Description :</li>
      <li class="rightHandSide">Alumni Guides, everything you need to land a job</li>
      <li class="leftHandSide">Forks :</li>
      <li class="rightHandSide">10</li>
      <li class="leftHandSide">Updated at :</li>
      <li class="rightHandSide">2019-06-21T11:38:21Z</li>
    </li>
    <li class="container">
      <li class="leftHandSide">Repository :</li>
      <li class="rightHandSide"><a href="https://github.com/HackYourFuture/angular" target="_blank">angular</a></li>
      <li class="leftHandSide">Description :</li>
      <li class="rightHandSide">Angular</li>
      <li class="leftHandSide">Forks :</li>
      <li class="rightHandSide">3</li>
      <li class="leftHandSide">Updated at :</li>
      <li class="rightHandSide">2018-04-06T17:17:51Z</li>
    </li>
    ...
  </ul>
</div>
```

Perhaps I can suggest an HTML `table` instead?

homework/index.js
```diff
@@ -30,7 +30,47 @@
...
+        createAndAppend('li', li, {
+          text: keyToRead(key) + ' :',
+          class: 'leftHandSide',
+        });
+        createAndAppend('li', li, {
+          text: repo[key],
+          class: 'rightHandSide',
+        });
+      }
+    });
```
remarcmij: I'm looking at the lines 51-73. I gather you attempted to make your code DRYer by using a loop rather than repeating the same code three times. However, using a loop makes your code unnecessary complicated. And your attempt to convert the property names of the repository object into label text that you can display further complicates the code.

The standard tool for making your code DRY is to use functions. For instance, let's assume that you use an HTML `table` for rendering the repository details. Then you can write an `addRow()` function that you could call like this:

```js
addRow(table, 'Description', repo.description);
addRow(table, 'Forks', repo.forks);
addRow(table, 'Updated at', repo.updated_at);
```

Voila: no need for a loop or a function to convert property names into label text.

homework/index.js
```diff
@@ -43,8 +83,12 @@
         });
         return;
       }
+      createAndAppend('h2', root, { text: 'HYF Repositories' });
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      const reposSorted = repos.sort(function(a, b) {
```
remarcmij: The ESLint error that is reported for this line is because it prefers you use an arrow function here.

homework/index.js
```diff
@@ -43,8 +83,12 @@
...
+        return a.name.localeCompare(b.name);
+      });
+      reposSorted.forEach(repo => renderRepoDetails(repo, ul));
```
remarcmij: You can combine the `.sort()` and `.forEach()` in an elegant chain like this:

```js
repos
  .forEach(repo => renderRepoDetails(repo, ul));
```

homework/style.css
```diff
@@ -1,3 +1,53 @@
 .alert-error {
```
remarcmij: On a wide-screen monitor your UI looks a bit stretched out. Perhaps set a max width and center the content horizontally?

image

haha1906:
Hi Jim,

I have completed all the mistakes and commited again.

Made a table and added rows through an `addRow()` function.
Instead of big complicated loops, I preferred smaller funtions.
Made some codes DRYer.
Arranged stylesheet for extra wide screens with a maximum width limit.

I have learned a lot in this homework.
And I noticed that I can write more understandable codes. That makes me happy.
If you want make anymore criticism, I would be happy.

homework/index.js
```diff
@@ -1,54 +1,81 @@
 'use strict';
 
-{
```
remarcmij: You have removed the curly braces which surround all the code. Without these braces (that create a block scope), all your functions are now "polluting" the global scope. It is best to keep them of global scope.

homework/index.js
```diff
@@ -1,54 +1,81 @@
...
+    const td = createAndAppend('td', tr);
+    createAndAppend('a', td, {
+      href: link,
+      target: '_blank',
+      text: key,
     });
+  } else {
+    createAndAppend('td', tr, { text: key });
   }
+}
```
remarcmij: You have understood well how this function can make your code DRY. Note however how this function includes code to deal with the exception for the first row, which should contain a link instead of simple text. A better strategy would be to let the caller of this function deal with the exception so that this function can be simplified to:.
```js
function addRow(property, key, parent) {
  const tr = createAndAppend('tr', parent);
  createAndAppend('th', tr, {
    text: `${property} :`,
  });
  createAndAppend('td', tr, { text: key });
  return tr;
}
```

Note that in the last line we are returning the created `tr` so that the caller of this function can pick it up.

Now, in your `renderRepoDetails()` function, you can deal with the first row as follows:

```js
const firstRow = addRow('Repository', '', table);
  href: repo.html_url,
  target: '_blank',
  text: repo.name,
});
addRow('Description', ...);
```

Note that we pass an empty string for the `key` parameter and save the return value of the function into the `firstRow` variable. Then we add the link to the last child of the `tr`, which is, of course, the `td` where the link should go.

PR #351, opened 2019-09-20
JS3 Week1 Homework

ihk2018:


homework/index.js
```diff
@@ -28,14 +28,75 @@
     });
     return elem;
   }
+  // convert date time
+  function convertDataTime(dateTimeTxt) {
```
remarcmij: What is the added value of the comment preceding this function? It just repeats what the function name already indicates (although there is a typo in the function name: it is not `data` but `date`).

homework/index.js
```diff
@@ -28,14 +28,75 @@
...
+    const dateTime = new Date(dateTimeTxt);
+    return dateTime.toLocaleString();
+  }
+  // sort repositories by name
+  function sortElements(a, b) {
```
remarcmij: This is a case where you seem to need a comment because the function name is not clear enough. If you renamed the function to `sortRepositoriesByName` you could get rid of the comment. It is always better to let the code speak for itself.

homework/index.js
```diff
@@ -28,14 +28,75 @@
...
+    return a.name.localeCompare(b.name);
+  }
 
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const repoList = createAndAppend('li', ul, { class: 'rep-list' });
```
remarcmij: You have two other CSS class names, viz. `repo-header` and `repo-details`. For consistency, it is best to use `repo-list` here.

homework/index.js
```diff
@@ -28,14 +28,75 @@
...
+    createAndAppend('span', forkedNumbers, {
+      text: repo.forks,
+      class: 'texts',
+    });
+
+    createAndAppend('span', updatedTime, {
+      text: convertDataTime(repo.updated_at),
+      class: 'texts',
+    });
   }
```
remarcmij: While your chosen HTML works, the information to be rendered is basically a table with two columns and four rows. Using an HTML table for this would be the most natural, intuitive representation. This is what your UI looks like on a mobile phone (perhaps also reduce the font size?):

image

Using an HTML table it could look like this:

image

If this was an app in the App Store, which one would you prefer?

homework/index.js
```diff
@@ -44,7 +105,7 @@
         return;
       }
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      repos.sort(sortElements).forEach(repo => renderRepoDetails(repo, ul));
```
remarcmij: Since the function body of  `sortElements` function is just a single line, you might as well code it in-line using an arrow function:

```js
repos
```

homework/index.js
```diff
@@ -28,14 +28,75 @@
...
+      class: 'repo-details',
+    });
+
+    createAndAppend('a', repositoryName, {
+      text: repo.name,
+      href: repo.html_url,
+      target: '_blank',
+    });
+
+    if (repo.description === null) {
```
remarcmij: Nice check 👍

homework/index.js
ihk2018: Dear @remarcmij  firts of all, thank you so much for the valuable feedbacks.

actually you are right there is no added value for a comment here, function name is clesar enough to tell about the code.  I will be more careful when commenting the codes.

homework/index.js
ihk2018: Thank you, it would be nicer to choose a clear name.

homework/index.js
ihk2018: thank you

PR #352, opened 2019-09-20
submit js3 week1 solutions- Abdullah

abdullahkanatsiz7:
my js3 week1 homework

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -28,53 +28,32 @@
     });
     return elem;
   }
-  function NewDate(dateString) {
+  function formatDate(dateString) {
     const dateTime = new Date(dateString);
     return dateTime.toLocaleString();
   }
 
-  function renderRepoDetails(repo, section) {
-    const Mix = createAndAppend('ul', section);
-    const name = createAndAppend('li', Mix, {
-      text: 'Repository:',
-      class: 'left-side',
-    });
-    const description = createAndAppend('li', Mix, {
-      text: 'Description:',
-      class: 'left-side',
-    });
-    const fork = createAndAppend('li', Mix, {
-      text: 'Forks:',
-      class: 'left-side',
-    });
-    const time = createAndAppend('li', Mix, {
-      text: 'Updated:',
-      class: 'left-side',
-    });
+  function addTable(table, body, value) {
+    const tr = createAndAppend('tr', table);
+    createAndAppend('th', tr, { text: body});
+    createAndAppend('td', tr, { text: value });
+    return tr;
+  }
 
-    // now I'm going to use that name,description,fork,time in the below function
-
-    createAndAppend('a', name, {
-      href: repo.clone_url,
-      target: '_blank',
-      class: 'right-side-1',
+  function renderRepoDetails(repo, li) {
+    const table = createAndAppend('table', li,);
+    const tr1 = addTable(table, 'Repository:', '');
+    createAndAppend('a', tr1.lastChild, {
+      href: repo.html_url,
       text: repo.name,
     });
-    createAndAppend('p', description, {
-      text: repo.description,
-      class: 'right-side-2',
-    });
-    createAndAppend('p', fork, { text: repo.forks, class: 'right-side-3' });
-    createAndAppend('p', time, {
-      text: NewDate(repo.updated_at),
-      class: 'right-side-4',
-    });
+    addTable(table, 'Description:', repo.description);
+    addTable(table, 'Fork: ', repo.forks);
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -28,27 +28,74 @@
     });
     return elem;
   }
+  function NewDate(dateString) {
```
remarcmij: Function, variable and parameter names should be in `camelCase`: `newDate`. Moreover, `newDate` is not the best possible name. You are not creating a new date, you are just formatting an existing date for presentation purposes. Therefore, the name `formatDate` would seem a better fit.

homework/index.js
```diff
@@ -28,27 +28,74 @@
...
+    const dateTime = new Date(dateString);
+    return dateTime.toLocaleString();
+  }
+
+  function renderRepoDetails(repo, section) {
+    const Mix = createAndAppend('ul', section);
```
remarcmij: This variable name should start with a lowercase letter. What does `mix` mean in this context?

homework/index.js
```diff
@@ -28,27 +28,74 @@
...
+    createAndAppend('p', description, {
+      text: repo.description,
+      class: 'right-side-2',
+    });
+    createAndAppend('p', fork, { text: repo.forks, class: 'right-side-3' });
+    createAndAppend('p', time, {
+      text: NewDate(repo.updated_at),
+      class: 'right-side-4',
+    });
+  }
```
remarcmij: The repository details that you render look like a table with two columns and four rows. Therefor an HTML table would be a natural choice. I noticed that you have manually fine-tuned the left padding in the classes `right-side-n` to vertically line up the values in the second column. However, this is not the recommended way as it is depended on the chosen font etc., which can lead to maintenance difficulties in the future when somebody decides the change the chosen font.

If you look closely on the UI as displayed on a mobile phone you can already see that the second column is not accurately aligned vertically:

image

Using an HTML table will solve this:

image

If this was an app on the App Store, which one would you install?

homework/index.js
```diff
@@ -28,27 +28,74 @@
...
           text: err.message,
           class: 'alert-error',
         });
         return;
       }
-      const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      const section = createAndAppend('section', root);
+
+      repos.sort(sortName).forEach(repo => renderRepoDetails(repo, section));
```
remarcmij: Since the function body of `sortName` is only a single line, I would prefer to code this function in-line as an arrow function:

```js
repos
```

homework/index.js
```diff
@@ -28,27 +28,74 @@
...
     });
   }
-
-  const HYF_REPOS_URL =
+  const HYF_REPO_URL =
```
remarcmij: Was this change really necessary?

homework/index.js
abdullahkanatsiz7: Firstly Thank you for this feedback. Yes, you right about that. I'm going to change its name

homework/index.js
abdullahkanatsiz7: I think list name will be better than mix

homework/index.js
abdullahkanatsiz7: thank you for this information because this is more short way

homework/index.js
abdullahkanatsiz7: of course not, I only forget that make change

homework/index.js
abdullahkanatsiz7: now I changed my function

homework/index.js
```diff
@@ -44,10 +70,14 @@
         return;
       }
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      const li = createAndAppend('li', ul);
+
+      repos
+  .sort((a, b) => a.name.localeCompare(b.name))
+  .forEach(repo => renderRepoDetails(repo, li));
```
remarcmij: You are using a single list item that contains all the repositories. I would expect one list item per repository. So each `table` element should be wrapped in an `li`.

abdullahkanatsiz7: Yes I reviewed another project for find table solution. I know this is not good for me. If you want  I  work to find another solution

remarcmij: No, you don't need to find another solution as far as I'm concerned. But for a next time, please try for yourself first and ask help if you get stuck.

abdullahkanatsiz7: yes you right I need to do like your said. Next time I will use only my solution. Thank you for this good feedback

PR #353, opened 2019-09-20
N.Ince-JS3-HW1

NesibeInce:


homework/index.js
```diff
@@ -43,8 +65,20 @@
...
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      createAndAppend('h1', root, {
+        text: 'HYF Repository',
+        class: 'title',
+      });
+      repos.sort((a, b) => {
+        return a.name.localeCompare(b.name);
+      });
+
+      repos.forEach(repo => {
```
remarcmij: Optionally, you can combine the sorting and `.forEach()` in a chain:

```js
repos
```

NesibeInce:
Hi Mr Jim,

Thank you for the feedback. I will certainly try and focus  to implement
it.

Because this coding and IT and everything are completely new to me, I am
trying to do all my best.

Nesibe Ince

homework/index.js
NesibeInce: Thank you for making it shorter. I implemented it.

homework/style.css
```diff
@@ -1,3 +1,45 @@
+#root {
```
remarcmij: I had made more comments on your homework, but somehow they got lost. If I look at your app on a mobile device then the repository values no longer line up vertically with their corresponding labels. Basically, the repository information to be displayed consists of a table with two columns and four rows. So to fix the alignment problem,  I suggest you try an HTML table to replace what you have now.

Your UI:

image


What it could look like using HTML tables:

image

NesibeInce: Sorry, your comment about being responsive is still there, i did not miss it, however i was working on it.
Since i applied the other one, i immediately wrote you back. Maybe I had to complete all and have sent you both together, sorry for that.

NesibeInce: Hi, goedemorgen.
I did my best and implemented the feedback about responsiveness.
Hope it works well now.
thanks in advance.

homework/index.js
```diff
@@ -29,8 +29,32 @@
...
+      text: 'Repository:',
+    });
+    const RepoNameTh = createAndAppend('td', RepositoryTr);
+    createAndAppend('a', RepoNameTh, {
+      text: repo.name,
+      href: repo.html_url,
+      target: '_blank',
+    });
+
+    function renderRepoTable(name, path) {
```
remarcmij: Avoid embedded named functions like this. Place at the file level (i.e. at the same level as, for instance, the `renderRepoDetails()` function. Of course, you will then need to add an extra parameter: `table`. And I would s

```js

function renderRepoTable(table, name, path) {
  createAndAppend('th', tr, {
    text: name,
  });
  createAndAppend('td', tr, {
    text: path,
  });
  return tr;
}
```

I also added a statement at the end: `return tr`. You can use the return value to handle the exception for the first row that requires a `<a>` instead of simple text content. Your `renderRepoDetails()` function could now look like this:

```js
function renderRepoDetails(repo, table) {
  const firstRow = renderRepoTable(table, 'Repository', '');
    text: repo.name,
    href: repo.html_url,
    target: '_blank',
  });
  renderRepoTable(table, 'Description:', repo.description);
  renderRepoTable(table, 'Forks:', repo.forks);
  const date = new Date(repo.updated_at).toLocaleString('en-US');
  renderRepoTable(table, 'Updated:', date);
}
```

homework/index.js
```diff
@@ -29,8 +29,32 @@
     return elem;
   }
 
-  function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+  function renderRepoDetails(repo, table) {
+    const RepositoryTr = createAndAppend('tr', table);
```
remarcmij: Remember that variable names should be in camelCase (start with a lowercase letter).

NesibeInce:
Hi@remarcmij Thank you for approving my pull request. I will follow your
suggestions.

PR #354, opened 2019-09-21
week1/homework

lamaajaj:


homework/index.js
```diff
@@ -28,14 +28,42 @@
...
     return elem;
   }
+  function createTable(table, header, value) {
+    const tr = createAndAppend('tr', table);
+    createAndAppend('th', tr, { text: header, class: 'keys' });
+    createAndAppend('td', tr, { text: value, class: 'values' });
+    return tr;
+  }
+
+  function changDate(dateTime) {
```
remarcmij: This is not the best possible name. You are not changing the date, you are just formatting an existing date for presentation purposes. So `formatDate` would be a better name.

homework/index.js
```diff
@@ -28,14 +28,42 @@
     });
     return elem;
   }
+  function createTable(table, header, value) {
```
remarcmij: You are not creating a table here, you are creating or adding a table row. So perhaps `addRow()` would be a better name?

homework/index.js
```diff
@@ -44,7 +72,7 @@
         return;
       }
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      repos.sort(sortFun).forEach(repos => renderRepoDetails(repos, ul));
```
remarcmij: Since the function body of your `sortFun()` consist of a single line only, you could just use an inline arrow function like this:

```js
repos
```

homework/style.css
```diff
@@ -1,3 +1,45 @@
 .alert-error {
```
remarcmij: Your styling is correct but will not win first prize in a beauty contest, I'm afraid 😉.

image

lamaajaj:
> Hi @lamaajaj , I see you took inspiration from other pull requests and my feedback on them. I would like to suggest that this is not an effective method of learning. It is best to try things on your own strength. I am not looking for a perfect solution. The end result is not what counts, it is the journey towards that result that I'm interested in.
> 
> I will approve this PR because it meets the functional requirements of the assignment, but please take my comment above to heart for a next time.

lamaajaj:
Hello @remarcmij , 
Actually the homework was unclear to me.
I watched a lot of videos even your old class records, but I didn't get it, so yes you are absolutely right. Some of my colleagues helped me, and I took a look at other homeworks but I understood the code afterwards. It was actually the most homework which I put effort and time in.
Thanks a lot for your feedback.

PR #355, opened 2019-09-22
Js3 W1 Homework(resubmission)

ersarue:


ersarue:
Dear Jim, as you might remember there was an error on my 'too early' cloned repo. I followed your instructions and uploaded the fresh state here. I was planning to add table for showing the sorted repos, but it took a while and didn't go well. I am on it and will add table as soon as possible. For now, I am submitting as to have your guiding feedbacks as early as possible.

homework/index.js
```diff
@@ -29,10 +29,68 @@
...
+    const timeFormat = new Date(date);
+    return timeFormat.toLocaleString();
+  }
+
+  function sortRepos(a, b) {
+    const first = a.name.toLowerCase();
+    const second = b.name.toLowerCase();
+    const key = first < second ? -1 : first > second ? 1 : 0;
+    return key;
   }
```
remarcmij: On this sort function, I'm giving you the same feedback as I gave Esra:

In the assignment we gave the hint to use `.localeCompare`. Your method works fine if all names just use the regular letters. However, as soon as they include letters with diacritics, such as in French, German, Dutch or Turkish the names are not sorted in the (dictionary) order you would expect.

You may want to try this code, comparing your sort method agains `.localeCompare()`. I have taken the surnames of the students in class23 but added another name, viz. **Öbul** which I would expect to be sorted just after **Obul**. With `.localeCompare()` (which takes into account the 'locale' that is set up in your computer) this works fine. Your sort method however places **Öbul** after **Zengin**.

```js
'use strict';

{
  const users = [
    'Hüsrev',
    'Raad',
    'Kanatsiz',
    'Ekinci',
    'Ince',
    'Ajaj',
    'Obul',
    'Öbul',
    'Zengin',
    'Karatas',
    'Ersaru',
    'Tercan',
    'Alnakib',
    'Grabetskaia',
    'Sönmez',
    'Akkuz',
  ];

  const esraSorted = users.slice().sort((a, b) => {
    const aUpper = a.toUpperCase();
    const bUpper = b.toUpperCase();
    if (aUpper < bUpper) {
      return -1;
    }
    if (aUpper > bUpper) {
      return 1;
    }
    return 0;
  });

  const localeSorted = users.slice().sort((a, b) => a.localeCompare(b));

  console.log(esraSorted);
  console.log(localeSorted);
}
```

homework/index.js
```diff
@@ -29,10 +29,68 @@
...
 
+  function renderRepoDetails(repo, ul) {
+    const grouping = createAndAppend('li', ul, { class: 'listItem' });
+    const name = createAndAppend('li', grouping, {
```
remarcmij: `grouping` is a `<li>`; this can't be the parent of another `<li>`. This would be invalid HTML. You need an intermediate `<ul>` as shown here:

image

homework/index.js
```diff
@@ -29,10 +29,68 @@
...
+    createAndAppend('span', forkCount, { text: repo.forks, class: 'content' });
+    createAndAppend('span', createTime, {
+      text: changeDateFormat(repo.created_at),
+      class: 'content',
+    });
+    createAndAppend('span', updateTime, {
+      text: changeDateFormat(repo.updated_at),
+      class: 'content',
+    });
+  }
```
remarcmij: This function contains a lot of repetition: create an `li` with text content and append a `span` to it with more text content. To make your code DRY, you can create a function, for instance:

```js
function addDetailLine(ul, title, content) {
  const li = createAndAppend('li', ul, {
    text: `${title}: `,
    class: 'title',
  });
  createAndAppend('span', li, {
    text: content,
    class: 'content',
  });
}
```

Now, in your `renderRepoDetails()` function you can use that function, with the exception of the **Repository name**, which needs an `<a>` tag instead of simple text:

```js
  const li = createAndAppend('li', ul, { class: 'listItem' });
  const ulDetail = createAndAppend('ul', li);
  const name = createAndAppend('li', ulDetail, {
    text: 'Repository name: ',
    class: 'title',
  });
  createAndAppend('a', name, {
    text: repo.name,
    href: repo.html_url,
    class: 'content',
  });

  addDetailLine(ulDetail, 'Repository language', repo.language);
  ...
}
```

homework/index.js
```diff
@@ -29,10 +29,68 @@
...
   function main(url) {
```
remarcmij: For readability, place a blank line before and after each function definition. It's like separating paragraphs with a blank line in a piece of text: it improves readability.

homework/index.js
```diff
@@ -44,10 +102,13 @@
         return;
       }
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      repos.sort(sortRepos).forEach(repo => renderRepoDetails(repo, ul));
     });
+    const mainTitle = document.createElement('header');
```
remarcmij: Why are you not using `createAndAppend()` here?

homework/index.js
```diff
@@ -19,20 +19,49 @@
   function createAndAppend(name, parent, options = {}) {
     const elem = document.createElement(name);
     parent.appendChild(elem);
-    Object.entries(options).forEach(([key, value]) => {
+    Object.entries(options).forEach(([key, content]) => {
```
remarcmij: In general, you should not make trivial changes such as this in a PR. Every change you make gets incorporated in the pull request and requires time from the reviewer to review. In this case it would be wasted time.

homework/index.js
```diff
@@ -19,20 +19,49 @@
...
+
+  function sortRepos(a, b) {
+    const first = a.name.toLowerCase();
+    const second = b.name.toLowerCase();
+    const key = first < second ? -1 : first > second ? 1 : 0;
+    return key;
   }
 
+  function renderRepoDetails(repo, ul) {
+    function createTable(table, heading, content) {
```
remarcmij: This function should stand on itself, i.e. avoid embedded function definition like this unless there is a good reason to embed it.

homework/index.js
ersarue: I've actually seen your feedback on this. Seems, I've missed that while being busy with something else.

homework/index.js
ersarue: I got it, I'll be having this sorted out on the next commit.

homework/index.js
ersarue: Right..

homework/index.js
ersarue: Well, no answer on this for myself also :) I'll make the change.

ersarue:
Dear Jim, first of all thank you for your feedbacks! That's right, since I was unlucky for this repo I was in hurry to get the table used repos application reviewed. 
About trivial changes, I'll avoid to do so, thanks for raising the issue. And using HTML table, you didn't give me such advice, rather it was a suggestion of friends.

PR #357, opened 2019-09-24
Week2

rabrad:
For review.

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -91,11 +91,9 @@
   }
 
   function renderOptionElements(repos, select) {
-    repos
-      .sort((a, b) => a.name.localeCompare(b.name))
-      .forEach((repo, index) => {
-        createAndAppend('option', select, { text: repo.name, value: index });
-      });
+    repos.forEach((repo, index) => {
+      createAndAppend('option', select, { text: repo.name, value: index });
+    });
   }
 
   function main(url) {
--- a/homework/index.js
+++ b/homework/index.js
@@ -58,34 +58,25 @@
 
   const root = document.getElementById('root');
 
-  function renderContributors(repo, ul) {
+  function renderContributors(repo, div) {
     fetchJSON(repo.contributors_url)
       .then(contributors => {
-        ul.innerHTML = '';
+        div.innerHTML = '';
 
         contributors.forEach(contributor => {
-          const li = createAndAppend('li', ul);
-          const table = createAndAppend('table', li, {
+          const contributorDiv = createAndAppend('div', div, {
             class: 'contributor-details',
           });
-          const tableBody = createAndAppend('tbody', table);
-          const tableRow = createAndAppend('tr', tableBody, {
-            class: 'table-row',
+          createAndAppend('img', contributorDiv, {
+            src: contributor.avatar_url,
           });
-
-          createAndAppend('img', tableRow, { src: contributor.avatar_url });
-
-          const contributorName = createAndAppend('td', tableRow, {
-            class: 'user-name',
-          });
-          const contributorNameP = createAndAppend('p', contributorName);
-          createAndAppend('a', contributorNameP, {
+          createAndAppend('a', contributorDiv, {
             text: contributor.login,
             href: contributor.html_url,
             target: '_blank',
+            class: 'user-name',
           });
-
-          createAndAppend('td', tableRow, {
+          createAndAppend('div', contributorDiv, {
             text: contributor.contributions,
             class: 'contributions-count',
           });
@@ -99,11 +90,12 @@
       });
   }
 
-  // I think i forgot to delete .sort form this function. it been used already in main:
   function renderOptionElements(repos, select) {
-    repos.forEach((repo, index) => {
-      createAndAppend('option', select, { text: repo.name, value: index });
-    });
+    repos
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .forEach((repo, index) => {
+        createAndAppend('option', select, { text: repo.name, value: index });
+      });
   }
 
   function main(url) {
@@ -126,7 +118,7 @@
       text: `Contributors:`,
       id: 'contributors-title',
     });
-    const ul = createAndAppend('ul', contributorsContainer, {
+    const div = createAndAppend('div', contributorsContainer, {
       id: 'list-contributions',
     });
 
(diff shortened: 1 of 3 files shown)


homework/index.js
```diff
@@ -29,23 +31,117 @@
...
+  }
+
+  /* must declare variable: root, before the next function. */
+  const root = document.getElementById('root');
+
+  function renderContributors(repo, ul) {
+    fetchJSON(repo.contributors_url)
+      .then(contributors => {
+        ul.innerHTML = '';
+        createAndAppend('p', ul, {
```
remarcmij: A `<p>` tag can't be a child of a `<ul>` tag. This would be invalid HTML.

homework/index.js
```diff
@@ -29,23 +31,117 @@
...
+          text: `Contributors:`,
+          id: 'contributors-title',
+        });
+        contributors.forEach(contributor => {
+          const li = createAndAppend('li', ul);
+          const table = createAndAppend('table', li, {
+            class: 'contributor-details',
+          });
+          const a = createAndAppend('a', table, {
```
remarcmij: An `<a>` can't be the child of a `<table>` tag. That would be invalid HTML.

homework/index.js
```diff
@@ -29,23 +31,117 @@
...
+    fetchJSON(repo.contributors_url)
+      .then(contributors => {
+        ul.innerHTML = '';
+        createAndAppend('p', ul, {
+          text: `Contributors:`,
+          id: 'contributors-title',
+        });
+        contributors.forEach(contributor => {
+          const li = createAndAppend('li', ul);
+          const table = createAndAppend('table', li, {
```
remarcmij: This is a table with a single row. It works, but a would a couple of divs work here too? It would give you more CSS styling flexibility.

homework/index.js
```diff
@@ -29,23 +31,117 @@
...
-        return;
-      }
-      const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      });
+  }
+
+  function createAndSortElements(repos, select) {
+    repos
+      .sort((a, b) => a.name.localeCompare(b.name))
```
remarcmij: This function produces a side-effect because it changes one of the parameters that was passed to it: `.sort()` sorts "in place", i.e. it mutates the array you are sorting. What's more, the caller of this function is actually relying on this side-effect, because in `main()` you expect the first element to be in the sorted order. Side-effects like this make your code harder to understand. I would advise doing the sorting in your main function:

```js
fetchJSON(url)
  .then(repos => repos.sort((a, b) => a.name.localeCompare(b.name)))
  .then(repos => {
    createAndSortElements(repos, select);
    renderRepoDetails(repos[0], reposContainer);
    renderContributors(repos[0], ul);
```

Of course, your function name `createAndSortElements` is then no longer applicable. You could rename it to, say, `renderOptionElements()`.

homework/index.js
```diff
@@ -29,23 +31,117 @@
...
+    });
+    const ul = createAndAppend('ul', contributorsContainer, {
+      id: 'list-contributions',
+    });
+
+    fetchJSON(url)
+      .then(repos => {
+        createAndSortElements(repos, select);
+        renderRepoDetails(repos[0], reposContainer);
+        renderContributors(repos[0], ul);
```
remarcmij: Render details for repo at start-up: 👍

homework/style.css
```diff
@@ -1,3 +1,153 @@
...
+.table-row {
+  display: grid;
+  grid-template-columns: 1fr 4fr 1fr;
+  margin-bottom: 10px;
+}
+
+.contributions-count {
+  text-align: right;
+  color: #000000;
+  margin-top: 43px;
```
remarcmij: You can vertically center the contributions count by replacing `margin-top` with `margin: auto 0;`

homework/style.css
```diff
@@ -1,3 +1,153 @@
...
+}
+
+.user-name {
+  margin-top: 43px;
```
remarcmij: You can vertically center the user name by replacing `margin-top` with `margin: auto 0;`

homework/style.css
```diff
@@ -1,3 +1,153 @@
...
+  border: 0;
+  border-radius: 2px;
+  background: url(http://cdn1.iconfinder.com/data/icons/cc_mono_icon_set/blacks/16x16/br_down.png)
+    no-repeat right #ffffff;
+  -webkit-appearance: none;
+  background-position-x: 216px;
+}
+
+#main-container {
+  display: grid;
```
remarcmij: Ah, grid. I must look into that one myself one of these days 😉.

homework/index.js
```diff
@@ -1,19 +1,21 @@
...
-      if (xhr.status >= 200 && xhr.status <= 299) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
```
remarcmij: Promise: 👍

rabrad: Thanks for your helpful feedback, I will edit my files accordingly

rabrad:
I have tried carefully to follow and amend according to your recommendations. 
I'm sorry If i accidentally missed something. Could you please specify your point?

homework/index.js
remarcmij: Hi Rabie. this is one.

homework/index.js
remarcmij: And another one. I would not use a table here, but I'm fine with you leaving it like it is.

homework/index.js
rabrad: I forgot to remove the sorting from renderOptionsElements function. It has been already used in main function.

homework/index.js
rabrad: Oh, I see. 
I have replaced the contributors table with divs.

homework/index.js
```diff
@@ -99,12 +99,11 @@
...
   }
 
+  // I think i forgot to delete .sort form this function. it been used already in main:
   function renderOptionElements(repos, select) {
-    repos
-      .sort((a, b) => a.name.localeCompare(b.name))
-      .forEach((repo, index) => {
-        createAndAppend('option', select, { text: repo.name, value: index });
-      });
+    repos.forEach((repo, index) => {
```
remarcmij: I don't know what happened here. You seem to have removed the sorting here, but in the final PR it seems to reappear.

rabrad: Yes, it is wrong.
It seems like I forgot to update my other local copy of the project where I test the amendments before transferring them to the main project folder.
I should be more organized on my work.

PR #358, opened 2019-09-24
Week2

haha1906:
I hope I could make with the correct branch

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -26,11 +26,52 @@
         elem.setAttribute(key, value);
       }
     });
+
     return elem;
   }
 
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const li = createAndAppend('li', ul, {
+      class: 'container',
+    });
+
+    if (repo.name != null) {
+      createAndAppend('li', li, {
+        text: 'Repository :',
+        class: 'leftHandSide',
+      });
+      const repositoryName = createAndAppend('li', li, {
+        class: 'rightHandSide',
+      });
+      createAndAppend('a', repositoryName, {
+        href: repo.html_url,
+        target: '_blank',
+        text: repo.name,
+      });
+    }
+    function keyToRead(key) {
+      let newValue = key;
+      for (let i = 0; i < key.length; i++) {
+        if (key[i] === '_') {
+          newValue = newValue.substr(0, i) + ' ' + newValue.substr(i + 1);
+        }
+      }
+      return newValue[0].toUpperCase() + newValue.slice(1);
+    }
+
+    const renderKeys = ['description', 'forks', 'updated_at'];
+    renderKeys.forEach(key => {
+      if (repo[key] != null) {
+        createAndAppend('li', li, {
+          text: keyToRead(key) + ' :',
+          class: 'leftHandSide',
+        });
+        createAndAppend('li', li, {
+          text: repo[key],
+          class: 'rightHandSide',
+        });
+      }
+    });
   }
 
   function main(url) {
@@ -43,8 +84,12 @@
         });
         return;
       }
+      createAndAppend('h2', root, { text: 'HYF Repositories' });
       const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      const reposSorted = repos.sort(function(a, b) {
+        return a.name.localeCompare(b.name);
+      });
+      reposSorted.forEach(repo => renderRepoDetails(repo, ul));
     });
   }
 
--- a/homework/index.js
+++ b/homework/index.js
@@ -96,6 +96,12 @@
   }
 
   function deleteContainers() {
+    const selectButton = document.getElementById('selectButton');
+    let firstOption = selectButton.firstChild;
+    while (firstOption) {
+      selectButton.removeChild(firstOption);
+      firstOption = selectButton.firstChild;
+    }
     document.getElementById('flex-container').remove();
   }
 
@@ -134,7 +140,9 @@
     const promise = fetchJSON(url);
     promise
       .then(repos => {
-        const selectButton = createAndAppend('select', header);
+        const selectButton = createAndAppend('select', header, {
+          id: 'selectButton',
+        });
         selectButton.addEventListener('change', () => {
           deleteContainers();
           makeFlexContainer(root, repos, selectButton);
--- a/homework/index.js
+++ b/homework/index.js
@@ -31,23 +31,23 @@
     return elem;
   }
 
-  function addCont(login, img, contributeNumber, parent, link) {
+  function addContributor(contributor, parent) {
     const tr = createAndAppend('tr', parent);
     const imageColumn = createAndAppend('th', tr);
     createAndAppend('img', imageColumn, {
-      src: img,
-      class: 'contributorImg',
+      src: contributor.avatar_url,
+      class: 'contributor-img',
     });
     const name = createAndAppend('td', tr);
     createAndAppend('a', name, {
-      href: link,
+      href: contributor.html_url,
       target: '_blank',
-      text: login,
+      text: contributor.login,
     });
     const contributeNamePlace = createAndAppend('td', tr);
     createAndAppend('span', contributeNamePlace, {
-      text: contributeNumber,
-      class: 'contributeNumber',
+      text: contributor.contributions,
+      class: 'contribute-number',
     });
   }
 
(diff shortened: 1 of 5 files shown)


haha1906:
> Hi @haha1906, at first glance your app looks nice, however I'm missing one of the requirements of the assignment which was actually the topic of week 2: convert the callback into a promise. Before I continue with the review, can I ask you to implement that?
> 
> By the way, the branch and pull request look fine.

I don't know how to do exactly, but I will try and upload in some hours

remarcmij:
I showed how to convert a callback to a promise in the second part of the lecture (check the recording). We also did it as an exercise in class.

You can also check the JavaScript3 examples. Compare **2-async** (using a callback) with **7-promise** from this link:

https://github.com/HackYourFuture/JavaScript3_examples/tree/master/src/week2

If you need further help, let me know.

haha1906:
Hi Jim,
I made callbacks promise. You can go on checking.

homework/index.js
```diff
@@ -29,12 +31,115 @@
     return elem;
   }
 
-  function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+  function addCont(login, img, contributeNumber, parent, link) {
```
remarcmij: What do you mean with `addCont`? Is it `addContributor`? How much effort is it to type those few more characters to take away any doubt on what you mean?

You are passing a log of parameters to this function. Except for `parent` they are all properties of a contributor object. Therefore this function would look less complicated and easier to understand if you just pass two parameters:

```js
function addContributor(contributor, parent) {
  ...
}
```

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+    const tr = createAndAppend('tr', parent);
+    const imageColumn = createAndAppend('th', tr);
+    createAndAppend('img', imageColumn, {
+      src: img,
+      class: 'contributorImg',
```
remarcmij: By convention, CSS identifiers (class names, ids) should be in **kebab-case**, i.e. all lowercase and words separated by hyphens: `contributor-img`.  Please check the rest of your code for similar issues.

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+  function addRow(property, key, parent) {
+    const tr = createAndAppend('tr', parent);
+    createAndAppend('th', tr, {
+      text: `${property} :`,
+    });
+    createAndAppend('td', tr, { text: key });
+    return tr;
+  }
+
+  let optionNumber = 0;
```
remarcmij: You should not define variables at the global level (i.e. outside of a function).

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+  function madeOption(repo, selectButton) {
```
remarcmij: Do not use the past tense for a verb when describing actions. Always use the present tense: make...

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+    createAndAppend('option', selectButton, {
+      text: repo.name,
+      value: optionNumber,
+    });
+    optionNumber++;
+  }
+
+  function renderContributorDetails(conts, contributorsContainer) {
```
remarcmij: Please do not use cryptic abbreviations such as `conts`. It only takes a few more characters to take away any doubt that you mean `constributors`.

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+    const table = createAndAppend('table', contributorsContainer);
```
remarcmij: An HTML table is not the best way to represent a list of contributors. Please take into account that your application (this one, or a future one) may need to be used by users that are visually impaired and will require on a screen reader to use your application. Therefore, for a list of contributors you should use an HTML element that semantically represents a list: in this case an unordered would be the best choice.

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+    conts.forEach(cont => {
+      addCont(
+        cont.login,
+        cont.avatar_url,
+        cont.contributions,
+        table,
+        cont.html_url,
+      );
```
remarcmij: As per my previous comment, this function call would now become:

```
addContributor(contributor, ul);
```

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+    });
+    addRow('Description', repo.description, table);
+    addRow('Forks', repo.forks, table);
+    addRow('Updated', repo.updated_at, table);
+  }
+
+  function fetchContributors(url, contributorsContainer) {
+    const promiseContributor = fetchJSON(url);
+    promiseContributor.then((conts, err) => {
+      if (err) {
```
remarcmij: Error handling with promises doesn't work this way. Please review again how to `.catch()` errors when using promises.

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+      text: repo.name,
+    });
+    addRow('Description', repo.description, table);
+    addRow('Forks', repo.forks, table);
+    addRow('Updated', repo.updated_at, table);
+  }
+
+  function fetchContributors(url, contributorsContainer) {
+    const promiseContributor = fetchJSON(url);
+    promiseContributor.then((conts, err) => {
```
remarcmij: `conts` -> `contributors`

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+          class: 'alert-error',
+        });
+        return;
+      }
+      renderContributorDetails(conts, contributorsContainer);
+    });
+  }
+
+  function deleteContainers() {
+    document.getElementById('flexContainer').remove();
```
remarcmij: **kebab-case**: `flex-container`

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+    const contributorsContainer = createAndAppend('section', flexContainer, {
+      class: 'container',
+    });
+    createAndAppend('h5', contributorsContainer, { text: 'Contributions' });
+    // eslint-disable-next-line func-names
+    repos
+      .sort((a, b) => a.name.localeCompare(b.name))
+      .forEach(repo => {
+        madeOption(repo, selectButton);
+      });
```
remarcmij: You are re-adding `<option>` elements each time a user changes the selected repository. For instance, I changed the selection from the first repo to the second repo, and now the select element contains these repeated option elements:

```html
<select>
  <option value="0">alumni</option>
  <option value="1">angular</option>
  <!-- etc -->
  <option value="45">TicTacToeTDD</option>
  <option value="46">ukulele</option>
  <option value="0">alumni</option>
  <option value="1">angular</option>
  <!-- etc -->
  <option value="45">TicTacToeTDD</option>
  <option value="46">ukulele</option>
</select>
```

Instead of creating the option elements here, you need to do it once only, directly after you have received the repository information from the GitHub API.

Furthermore, when you do that, there is not much to be gained from putting the code to create an `<option>` element in a separate function. In addition, you can use the `index` parameter to set the value of each option element like this:

```js
.forEach((repo, index) => {
  createAndAppend('option', selectButton, {
    text: repo.name,
    value: index,
  });
});
```

homework/index.js
```diff
@@ -29,12 +31,115 @@
...
+      id: 'flexContainer',
+    });
+    const repoContainer = createAndAppend('section', flexContainer, {
+      class: 'container',
+    });
+    const contributorsContainer = createAndAppend('section', flexContainer, {
+      class: 'container',
+    });
+    createAndAppend('h5', contributorsContainer, { text: 'Contributions' });
+    // eslint-disable-next-line func-names
```
remarcmij: Do not disable ESLint rules.

haha1906:
Hi Jim,

I have made all the variable names, promise errors properly. 
I still don't understand how to manage 'ul's, 'li's, alongside the 'table's. So I don't make any change on the main DOM plan. If it's ok, could I change it next week?

homework/index.js
remarcmij: This problem is still not solved: every time a user changes the selection, all repository names are again added to the select element, which thereby grows each time, as you can see in the figure below. Note that after `46 ukelele` it starts again with `0 alumni`.

image

homework/index.js
```diff
@@ -96,6 +96,12 @@
   }
 
   function deleteContainers() {
+    const selectButton = document.getElementById('selectButton');
+    let firstOption = selectButton.firstChild;
+    while (firstOption) {
+      selectButton.removeChild(firstOption);
+      firstOption = selectButton.firstChild;
+    }
```
remarcmij: This is an inefficient way to solve the problem of an ever-growing list of option elements. Basically, you are re-creating each time the very same option elements, each time the user changes the selection. To avoid growing the existing list you erase the previous content. The proper solution would be to create these option elements **once only**, in your main function after fetching the repo data, sorting it and creating the select element.

homework/index.js
remarcmij: Please follow up on this point, or indicate why not. See also message in the class23 channel with respect to ARIA-compliance.

PR #359, opened 2019-09-25
Week2

abdullahkanatsiz7:
js3 week2 homework

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -44,7 +44,7 @@
     });
   }
 
-  function addTable(table, body, value) {
+  function addRow(table, body, value) {
     const tr = createAndAppend('tr', table);
     createAndAppend('th', tr, { text: body });
     createAndAppend('td', tr, { text: value });
@@ -54,24 +54,48 @@
   function renderRepoDetails(repo) {
     const leftSide = document.getElementById('left-side');
     leftSide.innerHTML = '';
-    const ul = createAndAppend('ul', leftSide);
-    const li = createAndAppend('li', ul);
-    const table = createAndAppend('table', li);
-    const tr1 = addTable(table, 'Repository:', '');
+    const table = createAndAppend('table', leftSide);
+    const tr1 = addRow(table, 'Repository:', '');
     createAndAppend('a', tr1.lastChild, {
       href: repo.html_url,
       text: repo.name,
     });
-    addTable(table, 'Description:', repo.description);
-    addTable(table, 'Fork: ', repo.forks);
-    addTable(table, 'Updated:', formatDate(repo.updated_at));
+    addRow(table, 'Description:', repo.description);
+    addRow(table, 'Fork: ', repo.forks);
+    addRow(table, 'Updated:', formatDate(repo.updated_at));
   }
 
+  function renderContributor(contributor, contributorList) {
+    const li = createAndAppend('li', contributorList);
+    const list = createAndAppend('a', li, {
+      class: 'contributor-item',
+      href: contributor.html_url,
+      target: '_blank',
+    });
+    createAndAppend('img', list, {
+      class: 'image',
+      src: contributor.avatar_url,
+    });
+    createAndAppend('span', list, {
+      class: 'contributor-name',
+      text: contributor.login,
+    });
+    createAndAppend('span', list, {
+      class: 'contribution-count',
+      text: contributor.contributions,
+    });
+    list.addEventListener('click', () => {
+      window.open(contributor.html_url, '_blank');
+    });
+    list.addEventListener('key', event => {
+      if (event.key === 'Enter') window.open(contributor.html_url, '_blank');
+    });
+  }
   function fetchAndRender(repo) {
-    const rightSide = document.getElementById('right-side');
-    rightSide.innerHTML = '';
     fetchJSON(repo.contributors_url)
       .then(contributors => {
+        const rightSide = document.getElementById('right-side');
+        rightSide.innerHTML = '';
         createAndAppend('span', rightSide, {
           text: 'Contributions',
           class: 'title',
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -28,26 +30,133 @@
...
+    createAndAppend('th', tr, { text: body });
+    createAndAppend('td', tr, { text: value });
+    return tr;
+  }
+
+  function renderRepoDetails(repo) {
+    const leftSide = document.getElementById('left-side');
+    leftSide.innerHTML = '';
+    const ul = createAndAppend('ul', leftSide);
+    const li = createAndAppend('li', ul);
```
remarcmij: Since there is no longer a list of repositories here, there is no need for a `ul` with a single `li` child. You can make your table a direct child of `leftSide`.

homework/index.js
```diff
@@ -28,26 +30,133 @@
...
   }
 
-  function main(url) {
-    fetchJSON(url, (err, repos) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, {
-          text: err.message,
-          class: 'alert-error',
+  function addTable(table, body, value) {
```
remarcmij: You are not actually adding a table here, just a table row. So a better name would be `addRow` or `addTableRow`.

homework/index.js
```diff
@@ -28,26 +30,133 @@
...
+    createAndAppend('a', tr1.lastChild, {
+      href: repo.html_url,
+      text: repo.name,
+    });
+    addTable(table, 'Description:', repo.description);
+    addTable(table, 'Fork: ', repo.forks);
+    addTable(table, 'Updated:', formatDate(repo.updated_at));
+  }
+
+  function fetchAndRender(repo) {
```
remarcmij: This name `fetchAndRender` (and the name `startPage` below) remind me of the homework of previous classes 🤔.

homework/index.js
```diff
@@ -28,26 +30,133 @@
...
+    const rightSide = document.getElementById('right-side');
+    rightSide.innerHTML = '';
+    fetchJSON(repo.contributors_url)
+      .then(contributors => {
+        createAndAppend('span', rightSide, {
```
remarcmij: You get a better user experience (less flicker) if you clear out the previous content **after** you get the contributors data rather clearing it out first and then doing the network request.

```js
function fetchAndRender(repo) {
  fetchJSON(repo.contributors_url)
    .then(contributors => {
      const rightSide = document.getElementById('right-side');
      rightSide.innerHTML = '';
      createAndAppend('span', rightSide, {
        ...
```

homework/index.js
```diff
@@ -28,26 +30,133 @@
...
+            text: contributor.contributions,
+          });
+          list.addEventListener('click', () => {
+            window.open(contributor.html_url, '_blank');
+          });
+          list.addEventListener('key', event => {
+            if (event.key === 'Enter')
+              window.open(contributor.html_url, '_blank');
+          });
+        });
```
remarcmij: This `fetchAndRender()` function is rather large. For readability, I would advise to split this function by creating a separate function that contains all the code that you now have in the arrow function in lines 82-105. This function needs two parameters:

```js
function renderContributor(contributor, contributorList) {
  const list = createAndAppend('li', contributorList, {
    class: 'contributor-item',
  });
  ...
}
```

Then, in line 82 you can call this function like this:

```js
contributors.forEach(contributor =>
  renderContributor(contributor, contributorList),
);
```

homework/index.js
```diff
@@ -28,26 +30,133 @@
...
+            class: 'contribution-count',
+            text: contributor.contributions,
+          });
+          list.addEventListener('click', () => {
+            window.open(contributor.html_url, '_blank');
+          });
+          list.addEventListener('key', event => {
+            if (event.key === 'Enter')
+              window.open(contributor.html_url, '_blank');
+          });
```
remarcmij: This second event listener for the Enter key should allow the user to open the contributor's page. Unfortunately, there is no way for a user to navigate to the list item in order press Enter on it. You need to include the list item into the keyboard tab order. You can do this by setting the `tabindex` attribute when you create the list item in line 83:

```js
const list = createAndAppend('li', contributorList, {
  class: 'contributor-item',
  tabindex: 0,
});
```

But another approach to enable both keyboard and mouse click for opening the contributor's page without using any event listeners is to wrap the whole content of the list item into an `<a>` tag with an appropriate `href`. Responding to mouse clicks and the Enter key is standard behaviour for an `<a>` tag.

The proposed function `renderContributor()` would then look like this:

```js
function renderContributor(contributor, contributorList) {
  const li = createAndAppend('li', contributorList);
  const anchor = createAndAppend('a', li, {
    class: 'contributor-item',
    href: contributor.html_url,
    target: '_blank',
  });
  createAndAppend('img', anchor, {
    class: 'image',
    src: contributor.avatar_url,
  });
  createAndAppend('span', anchor, {
    class: 'contributor-name',
    text: contributor.login,
  });
  createAndAppend('span', anchor, {
    class: 'contribution-count',
    text: contributor.contributions,
  });
}
```

To get rid of the underline normally present for a `<a>` tag you could set the `text-decoration: none;` for the `contributor-item` class.

homework/index.js
```diff
@@ -28,26 +30,133 @@
...
+        });
+      })
+      .catch(error => renderError(error));
+  }
+
+  function startPage(repository) {
+    const select = document.getElementById('select');
+    repository.sort((a, b) => a.name.localeCompare(b.name));
+    repository.forEach((repositories, index) => {
```
remarcmij: The first parameter of the `.forEach()` method is a single repository, therefore the name should be in the singular form: `repository` or `repo`. Readers of your code will be put on the wrong foot if you don't use the correct form.

homework/index.js
```diff
@@ -28,26 +30,133 @@
...
+      fetchAndRender(repository[select.value]);
     });
+    return repository[select.value];
   }
 
+  function main(url) {
+    const root = document.getElementById('root');
+    const header = createAndAppend('header', root);
+    createAndAppend('p', header, {
+      class: 'Header',
```
remarcmij: CSS names should be in **kebab-case**: all lowercase (and with hyphens to separate words).

homework/style.css
```diff
@@ -1,3 +1,142 @@
...
+  width: 32%;
+  line-height: 1.3;
+  height: 2rem;
+  padding: 3px;
+  max-width: 50%;
+  margin-left: 1rem;
+  cursor: pointer;
+  outline: none;
+}
+
```
remarcmij: You get a less jumpy user experience if you fix the left and right side to each take up half of the width:

```css
#left-side,
#right-side {
  flex: 1;
}
```

homework/index.js
```diff
@@ -28,26 +30,133 @@
...
+          list.addEventListener('key', event => {
+            if (event.key === 'Enter')
+              window.open(contributor.html_url, '_blank');
+          });
+        });
+      })
+      .catch(error => renderError(error));
+  }
+
+  function startPage(repository) {
```
remarcmij: This function receives an array of repository objects. Therefore the parameter name should be in plural form: `repositories` (or `repos`);

abdullahkanatsiz7:
Thank you remarcmij all this reviewed. I recognized and understood my many simple or huge mistake thanks to you. I changed my homework as your said and I sent. And I want to ask that I need make more practice about the js3, do you have any suggestion for that?

homework/index.js
```diff
@@ -28,26 +30,136 @@
...
+    createAndAppend('span', list, {
+      class: 'contribution-count',
+      text: contributor.contributions,
+    });
+    list.addEventListener('click', () => {
+      window.open(contributor.html_url, '_blank');
+    });
+    list.addEventListener('key', event => {
+      if (event.key === 'Enter') window.open(contributor.html_url, '_blank');
+    });
```
remarcmij: Now that you used the `<a>` tag that wraps the complete content of the list item, you no longer need these event listeners. You can remove lines 87-92.

homework/index.js
```diff
@@ -28,26 +30,136 @@
     });
     return elem;
   }
+  function formatDate(dateString) {
```
remarcmij: Use a blank line between function definitions for improved readability.

PR #360, opened 2019-09-25
Homework Week2

msnmz:


homework/index.js
```diff
@@ -29,26 +37,163 @@
...
+        description: repositoryInfo.description,
+        forks: repositoryInfo.forks,
+        lastUpdateTime: getDateTimeText(repositoryInfo.updated_at),
+        contributorsUrl: repositoryInfo.contributors_url,
+      }));
+  }
+
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const repoDetailsListItem = createAndAppend('li', ul, {
```
remarcmij: You are now rendering the details for a single repository, so it no longer makes sense to use an unordered list(with a single list item). You can make your table a direct child of the parent section.

homework/index.js
```diff
@@ -29,26 +37,163 @@
...
+    appendRepoDetail('Description:', repo.description, repoDetailsTable);
+    appendRepoDetail('Forks:', repo.forks, repoDetailsTable);
+    appendRepoDetail('Updated:', repo.lastUpdateTime, repoDetailsTable);
+  }
+
+  function renderContributorDetails(contributor, ul, index) {
+    const repoContributorListItem = createAndAppend('li', ul, {
+      class: 'contributors-item',
+    });
+    if (index === 0) {
```
remarcmij: Rather than complicating this function with the exception of dealing with the **Contributions** header, it is better to let the called take care of the exception and take out the `if` statement here. And this header does not have to be a list item anyway. It can just an `h6` that sits directly above the unordered list.

homework/index.js
```diff
@@ -29,26 +37,163 @@
...
+      createAndAppend('h6', repoContributorListItem, {
+        text: 'Contributions:',
+      });
+    }
+    createAndAppend('img', repoContributorListItem, {
+      src: contributor.avatar_url,
+      alt: 'Contributor avatar picture',
+    });
+    createAndAppend('a', repoContributorListItem, {
```
remarcmij: Instead of creating a link here, you could create a link to encloses all the content of the list item. In that way, you could click anywhere on the list item to go to the contributor's page.

homework/index.js
```diff
@@ -29,26 +37,163 @@
...
+  }
+
+  function renderMainSectionWithSubSections(parent) {
+    const mainSection = createAndAppend('main', parent, { class: 'container' });
+    renderSubSection(REPOSITORY_DETAIL_SECTION_ID, mainSection);
+    renderSubSection(REPOSITORY_CONTRIBUTORS_SECTION_ID, mainSection);
+  }
+
+  function getUnorderedListAsEmpty(sectionId) {
+    const section = document.getElementById(sectionId);
```
remarcmij: Here you are creating using `id` attributes to get access to elements that are created elsewhere in your code. You could have saved references to those elements when you created them, rather than placing "flag poles" in the DOM to find them again later with `.getElementById()`. You could either pass references to the relevant elements through parameters to the functions that need them or, alternatively, you could create a global object, and store the references there:

```js
'use strict';

{
  const dom = {};
  ...

function renderSubSection(sectionId, parent) {
  dom[sectionId] = createAndAppend('section', parent, {
    class: sectionId,
  });
  createAndAppend('ul', subSection, { class: 'sub-container' });
}
```

Personally, I would prefer explicitly passing the references using the parameter lists.

homework/index.js
```diff
@@ -29,26 +37,163 @@
...
+      repository,
+      getUnorderedListAsEmpty(REPOSITORY_DETAIL_SECTION_ID),
+    );
+  }
+
+  function renderRepositoryContributorsSection(repository) {
+    const ul = getUnorderedListAsEmpty(REPOSITORY_CONTRIBUTORS_SECTION_ID);
+    fetchJSON(repository.contributorsUrl).then(contributors => {
+      // In case of non contributors situation
+      if (contributors) {
```
remarcmij: If there are no contributors, you are not rendering anything on the right-hand side. This may leave users wondering why they don't see anything there. I think rendering just an empty list would better

homework/index.js
```diff
@@ -29,26 +37,163 @@
...
 
   function main(url) {
-    fetchJSON(url, (err, repos) => {
-      const root = document.getElementById('root');
-      if (err) {
+    const root = document.getElementById('root');
+    const select = renderHeaderSection('HYF Repositories', root);
+    fetchJSON(url)
+      .then(repos => {
+        const repositoryInfos = sortAndMapRepositories(repos);
```
remarcmij: This is just a nit-picking comment, but that there is not plural form for 'information'. See for instance this link:

https://blog.harwardcommunications.com/2010/11/09/information/

msnmz:
Hi @remarcmij, Thank you for your great comments. In this homework, I was left with the dilemma of whether I made the code as small functions or held it as a whole. And sometimes could not find the right verbs for naming functions and variables. Thank you for your great effort.

homework/index.js
```diff
@@ -29,26 +39,142 @@
...
+      class: 'header-title',
+    });
+    return createAndAppend('select', header);
+  }
+
+  function createMainSectionWithSubSections(parent) {
+    const mainSection = createAndAppend('main', parent, { class: 'container' });
+    dom[DETAIL_SECTION] = createAndAppend('section', mainSection, {
+      class: DETAIL_SECTION,
+    });
```
remarcmij: I doubt the added value of using these CONSTANTS. I would myself hard code the class name and use dot notation for the property names (it also reduced the amount of SHOUTING in your code 😉):

```js
dom.detailSection = createAndAppend('section', mainSection, {
  class: 'repository-detail-container',
});
```

homework/index.js
```diff
@@ -29,26 +39,142 @@
...
+    dom[CONTRIBUTORS_LIST].innerHTML = '';
+    fetchJSON(repository.contributors_url).then(contributors => {
+      // In case of non contributors situation
+      if (contributors) {
+        contributors.forEach(contributor => {
+          renderContributorDetails(contributor, dom[CONTRIBUTORS_LIST]);
+        });
+      } else {
+        createAndAppend('li', dom[CONTRIBUTORS_LIST], {
+          text: 'There is not any contributors for this repository.',
```
remarcmij: Nit-picking on English: 'There are no contributors for this repository.'

PR #361, opened 2019-09-26
Week-2 Homework 

ahmthsrvsmz:


homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+    addRow(table, 'Updated:', date.toLocaleDateString(), 'td');
+  }
+
+  function createOptions(repo, parent, index) {
+    createAndAppend('option', parent, {
+      text: repo.name,
+      value: index,
+    });
+  }
+  function selections(selection, data, parent) {
```
remarcmij: 1. This function name is rather vague. What does this function do? That is what the name should tell us.
2. The name `selection` (first param) is also rather vague. What does it contain? I think the name `selectedIndex` would make it clear what it contains.
3. Don't use the generic name `data` (second param) if you know that this function actually receives an array of repository objects. Therefore,  the name `repositories` would seem a good choice.

homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+    const ul = createAndAppend('ul', parent, {
+      class: 'ul-root',
+    });
+    const repo = data[selection];
+    renderRepoDetails(repo, ul);
```
remarcmij: Now that you are rendering the details for a single repository rather than a list of repositories as you did in week 1, it is now redundant to wrap your HTML table in an `ul`. This is what your code generates now:

```html
<section class="repo-container">
  <ul class="ul-root">
    <li class="li-root">
      <table>
        <!-- etc -->
      </table>
    </li>
  </ul>
</section>
```

But all you need is:

```html
<section class="repo-container">
  <table>
    <!-- etc -->
  </table>
</section>

```

homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+  }
+  function createContributorsSection(url, parent) {
+    fetchJSON(url).then(data => {
```
remarcmij: Again, don't use the name `data` if you have more precise knowledge about what you are getting here: you are fetching an array of contributor objects, therefore the logical name would be `contributors`.

homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+      data.forEach(repo => {
```
remarcmij: If you had used the correct name for the contributors array, you probably would not have made the mistake of using the name `repo` here instead of  `contributor`. Please make sure that you don't confuse other reading your code (or perhaps even yourself) by using incorrect names.

homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+          text: repo.login,
+          href: repo.html_url,
+          target: '_blank',
+          class: 'cont-name',
+        });
+        createAndAppend('p', li, {
+          text: repo.contributions,
+        });
+      });
+    });
```
remarcmij: You are not handling the possible error from the `fetchJSON()` promise.

homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+      'section',
+      mainSection,
+      {
+        class: 'contributor-container',
+      },
+    );
+    const p = createAndAppend('p', contributorsContainerSection, {
+      class: 'contributors-title',
+    });
+    p.innerHTML = 'Contributions';
```
remarcmij: Why are you not using the `text` property to set the text content using the `createAndAppend` call in line 117?

homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+    const ulCont = createAndAppend('ul', contributorsContainerSection, {
+      class: 'ul-cont-root',
+    });
+
+    const select = createAndAppend('select', header, {
+      class: 'selection-module',
+    });
+
+    fetchJSON(url)
+      .then(data => {
```
remarcmij: Here again, take away the guessing by using a descriptive name for the data you receive from `fetchJSON()`. In this case, you are fetching an array of repository objects, therefore the name `repositories` or `repos` would seem appropriate.

homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+        renderRepoDetails(data[17], ul);
```
remarcmij: @ersarue mentioned that you guys teamed up and both had difficulties in displaying the initial repo details. Of course, index 17 may no longer be valid if repositories are added or removed. The solution is simple: first sort, then render the first repo which will now always be at index `0`:

```js
fetchJSON(url)
  .then(repos => {
    repos
      .forEach((repo, index) => {
        createOptions(repo, select, index);
      });
    renderRepoDetails(repos[0], ul);
    const urlStart = repos[0].contributors_url;
    ...
```

homework/index.js
```diff
@@ -29,23 +31,127 @@
...
+            createOptions(repo, select, index);
+          });
+        select.addEventListener('change', () => {
+          repoContainerSection.innerHTML = '';
+          ulCont.innerHTML = '';
+          const selection = select.value;
+          selections(selection, data, repoContainerSection);
+          const contUrl = data[selection].contributors_url;
+          createContributorsSection(contUrl, ulCont);
+        });
```
remarcmij: Basically you need to execute the same code to render a repository in two different places:

1. Once, at application start-up, using the repository at index zero.
2. Inside the change event handler, using the index of the selected repo.

When you detect a repeated code like this you should immediately consider writing a function to remove this duplication. Can try and write such a function?

homework/index.js
ahmthsrvsmz: I will remove ul, li but add div inside this section because I need another child for styling this part in CSS.

homework/index.js
ahmthsrvsmz: I really don't know why I did this

ahmthsrvsmz: I really don't know why I did this

homework/index.js
```diff
@@ -29,23 +31,141 @@
...
+  }
+
+  function createOption(repo, parent, index) {
+    createAndAppend('option', parent, {
+      text: repo.name,
+      value: index,
+    });
+  }
+
+  function selectedIndex(selection, repositories, parent) {
```
remarcmij: I think you misunderstood my previous comment with regard to the name of this function. It was first parameter for which I suggested to use the name `selectedIndex`, not the function name itself. Function names should usually indicate 'action', e.g. have an active verb in the name.

But in fact, you do not need this function at all. See my other comment on that below.

homework/index.js
```diff
@@ -29,23 +31,141 @@
...
+      })
+      .catch(err => {
+        createAndAppend('div', parent, {
+          text: err.message,
+          class: 'alert-error',
+        });
+      });
+  }
+
+  function firstScreen(repos, repoParent, contParent, index) {
```
remarcmij: You don't need this function (see my comment below).

homework/index.js
```diff
@@ -29,23 +31,141 @@
...
+    renderRepoDetails(repos[index], repoParent);
+    createContributorsSection(repos[index].contributors_url, contParent);
+  }
+
+  function changeSelection(select, repos, repoParent, contParent) {
+    const selection = select.value;
+    const contUrl = repos[selection].contributors_url;
+    selectedIndex(selection, repos, repoParent);
+    createContributorsSection(contUrl, contParent);
   }
```
remarcmij: You can simplify this function to:

```js
function changeSelection(repo, repoParent, contParent) {
  repoParent.innerHTML = '';
  contParent.innerHTML = '';
  renderRepoDetails(repo, repoParent);
  createContributorsSection(repo.contributors_url, contParent);
}
```

You just need to pass the selected repo as the first parameter. For the first screen this would be the first repo in the repos array. And otherwise, it would be the selected repo from the select box.

homework/index.js
remarcmij: With the modified functions, your code could look like this:

```js
...
fetchJSON(url)
  .then(repos => {
    repos
      .forEach((repo, index) => {
        createOption(repo, select, index);
      });
    changeSelection(repos[0], div, ulCont);

      changeSelection(repos[select.value], div, ulCont);
    });
  })
  ...
```

homework/index.js
ahmthsrvsmz: @remarcmij it is possible for my fault but it didn't work. The first parameter worked, I already do it with your hint but selection parameters didn't work. Result: All contributors shows. I have to add selections function in changeselection function.

ahmthsrvsmz: Got it  👍

PR #362, opened 2019-09-26
Js3 W2 Homework

ersarue:
Dear Jim, even though the whole picture still is not completely okay I understand the logic. Thank you for your teaching!
As you might notice on line 114 I have specified the `repoItem` with a number in order to get first repo (alumni) to be seen when browser loads the page. I wasn't be able to make it without doing so (I have teamed up with @ahmthsrvsmz on this).

### DIFF
--- a/homework/index.js
+++ b/homework/index.js
@@ -21,11 +21,11 @@
   function createAndAppend(name, parent, options = {}) {
     const elem = document.createElement(name);
     parent.appendChild(elem);
-    Object.entries(options).forEach(([key, content]) => {
+    Object.entries(options).forEach(([key, value]) => {
       if (key === 'text') {
-        elem.textContent = content;
+        elem.textContent = value;
       } else {
-        elem.setAttribute(key, content);
+        elem.setAttribute(key, value);
       }
     });
     return elem;
@@ -36,9 +36,9 @@
     return timeFormat.toLocaleString();
   }
 
-  function createTable(table, header, content) {
-    const row = createAndAppend('tr', table);
-    createAndAppend('th', row, { text: header, class: 'header' });
+  function createTableRow(table, header, content) {
+    const row = createAndAppend('tr', table, { class: 'row' });
+    createAndAppend('th', row, { text: header, class: 'table-header' });
     createAndAppend('td', row, { text: content, class: 'content' });
     return row;
   }
@@ -46,35 +46,36 @@
   function renderRepoDetails(repo, ul) {
     const listItem = createAndAppend('li', ul, { class: 'listItem' });
     const table = createAndAppend('table', listItem, { class: 'table' });
-    const topRow = createTable(table, 'Repository: ');
+    const topRow = createTableRow(table, 'Repository: ');
     createAndAppend('a', topRow.lastChild, {
       text: repo.name,
       href: repo.html_url,
     });
-    createTable(table, 'Description: ', repo.description);
-    createTable(table, 'Fork count: ', repo.forks);
-    createTable(table, 'Created on: ', changeDateFormat(repo.created_at));
-    createTable(table, 'Updated on: ', changeDateFormat(repo.updated_at));
+    createTableRow(table, 'Description: ', repo.description);
+    createTableRow(table, 'Fork count: ', repo.forks);
+    createTableRow(table, 'Created on: ', changeDateFormat(repo.created_at));
+    createTableRow(table, 'Updated on: ', changeDateFormat(repo.updated_at));
   }
 
-  function contributorsSection(url, parent) {
+  function getContributorsData(url, parent) {
     fetchJSON(url)
-      .then(repos => {
-        repos.forEach(repo => {
+      .then(contributors => {
+        contributors.forEach(contributor => {
           const li = createAndAppend('li', parent, { class: 'rightList' });
           createAndAppend('img', li, {
             class: 'image',
-            src: repo.avatar_url,
+            src: contributor.avatar_url,
+            alt: `Image of the contributor ${contributor.login}`,
           });
           createAndAppend('a', li, {
-            text: repo.login,
-            href: repo.html_url,
+            text: contributor.login,
+            href: contributor.html_url,
             target: '_blank',
-            class: 'cont-link',
+            class: 'contributors-link',
           });
           createAndAppend('p', li, {
             class: 'contributions',
-            text: repo.contributions,
+            text: contributor.contributions,
           });
         });
       })
@@ -87,34 +88,35 @@
   }
 
   function repoSelect(item, repoItem, parent) {
-    const ul = createAndAppend('ul', parent);
+    const repoUl = createAndAppend('ul', parent, { class: 'repo-ul' });
     const repo = repoItem[item];
-    renderRepoDetails(repo, ul);
+    renderRepoDetails(repo, repoUl);
   }
 
   function main(url) {
     const root = document.getElementById('root');
     const header = createAndAppend('header', root, {
+      class: 'main-header',
       text: 'HYF Repositories',
     });
-    const mainElem = createAndAppend('main', root, { class: 'main-container' });
-    const repoContainer = createAndAppend('section', mainElem, {
+    const select = createAndAppend('select', header, { class: 'select' });
+    const main = createAndAppend('main', root, { class: 'main-container' });
+    const repoContainer = createAndAppend('section', main, {
       class: 'repo-container',
     });
-    const ul = createAndAppend('ul', repoContainer);
-    const rightSection = createAndAppend('section', mainElem, {
+    const contributorContainer = createAndAppend('section', main, {
       class: 'contributors-container',
     });
-    const p = createAndAppend('p', rightSection, { class: 'right-title' });
-    p.innerHTML = 'Contributions';
-    const rightUl = createAndAppend('ul', rightSection, { class: 'right-ul' });
-    const select = createAndAppend('select', header, { class: 'test' });
+    createAndAppend('p', contributorContainer, {
+      text: 'Contributions',
+      class: 'right-title',
+    });
+    const rightUl = createAndAppend('ul', contributorContainer, {
+      class: 'right-ul',
+    });
     fetchJSON(url)
-      .then(repoItem => {
-        renderRepoDetails(repoItem[17], ul);
-        const startItem = repoItem[17].contributors_url;
-        contributorsSection(startItem, rightUl);
-        repoItem
+      .then(repositories => {
+        repositories
           .sort((a, b) => a.name.localeCompare(b.name))
           .forEach((repo, index) => {
             createAndAppend('option', select, {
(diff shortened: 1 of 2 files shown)


homework/index.js
```diff
@@ -1,53 +1,143 @@
...
-      if (xhr.status >= 200 && xhr.status <= 299) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
```
remarcmij: Convert callback to promise; 👍

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
+      xhr.onerror = () => reject(new Error('Network request failed'));
+      xhr.send();
+    });
   }
 
   function createAndAppend(name, parent, options = {}) {
     const elem = document.createElement(name);
     parent.appendChild(elem);
-    Object.entries(options).forEach(([key, value]) => {
+    Object.entries(options).forEach(([key, content]) => {
```
remarcmij: What is the added value of this change? Can you that, as a consequence, this change now is included in the pull request, needlessly requiring the attention of the reviewer?

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
     });
     return elem;
   }
 
+  function changeDateFormat(date) {
+    const timeFormat = new Date(date);
+    return timeFormat.toLocaleString();
+  }
+
+  function createTable(table, header, content) {
```
remarcmij: This function does not create a table: it creates a table **row**. So a more precise name would be `createTableRow`.

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
+    const row = createAndAppend('tr', table);
+    createAndAppend('th', row, { text: header, class: 'header' });
+    createAndAppend('td', row, { text: content, class: 'content' });
+    return row;
+  }
+
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    const listItem = createAndAppend('li', ul, { class: 'listItem' });
```
remarcmij: In the week 1 homework, you were rendering a list of repository details. In week 2 you only render a single repository. Therefore, you no longer need to wrap the table containing the details in a `ul`. This is what you generate now:

```html
<section class="repo-container">
  <ul>
    <li class="listItem">
      <table class="table">
        <!-- etc -->
      </table>
    </li>
  </ul>
</section>
```

But you only need this:

```html
<section class="repo-container">
  <table class="table">
    <!-- etc -->
  </table>
</section>
```

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
+    });
+    createTable(table, 'Description: ', repo.description);
+    createTable(table, 'Fork count: ', repo.forks);
+    createTable(table, 'Created on: ', changeDateFormat(repo.created_at));
+    createTable(table, 'Updated on: ', changeDateFormat(repo.updated_at));
+  }
+
+  function contributorsSection(url, parent) {
+    fetchJSON(url)
+      .then(repos => {
```
remarcmij: Unfortunately, you are not telling the truth here. You are suggesting to the reader that you are fetching an array of repository objects, wherein truth you are fetching an array of contributor objects. This is misleading both for the readers of your code as for yourself.

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
+        repos.forEach(repo => {
```
remarcmij: Similarly, the argument for the arrow function should be `contributor`.

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
+    const mainElem = createAndAppend('main', root, { class: 'main-container' });
+    const repoContainer = createAndAppend('section', mainElem, {
+      class: 'repo-container',
+    });
+    const ul = createAndAppend('ul', repoContainer);
+    const rightSection = createAndAppend('section', mainElem, {
+      class: 'contributors-container',
+    });
+    const p = createAndAppend('p', rightSection, { class: 'right-title' });
+    p.innerHTML = 'Contributions';
```
remarcmij: Why not use a `text` property in the call the `createAndAppend()` in line 108?

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
+    const rightUl = createAndAppend('ul', rightSection, { class: 'right-ul' });
+    const select = createAndAppend('select', header, { class: 'test' });
+    fetchJSON(url)
+      .then(repoItem => {
```
remarcmij: You are fetching an array of repository objects here, therefore the parameter name should be in plural form: `repoItems`, although I would prefer `repositories` or `repos` for short.

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
+        const startItem = repoItem[17].contributors_url;
+        contributorsSection(startItem, rightUl);
+        repoItem
+          .sort((a, b) => a.name.localeCompare(b.name))
+          .forEach((repo, index) => {
+            createAndAppend('option', select, {
+              text: repo.name,
+              value: index,
+            });
+          });
```
remarcmij: As I mentioned in the feedback of @ahmthsrvsmz with whom you teamed up on this, the solution to rendering the first repo is simple: first sort, then render:

```js
fetchJSON(url)
  .then(repos => {
    repos
      .forEach((repo, index) => {
          text: repo.name,
          value: index,
        });
      });
    renderRepoDetails(repos[0], ul);
    const startItem = repos[0].contributors_url;
    contributorsSection(startItem, rightUl);
```

homework/index.js
```diff
@@ -1,53 +1,143 @@
...
+        select.addEventListener('change', () => {
+          repoContainer.innerHTML = '';
+          rightUl.innerHTML = '';
+          const selectInput = select.value;
+          repoSelect(selectInput, repoItem, repoContainer);
+          const contUrl = repoItem[selectInput].contributors_url;
+          contributorsSection(contUrl, rightUl);
+        });
+      })
```
remarcmij: I'm repeating the same comment I gave  @ahmthsrvsmz:

Basically you need to execute the same code to render a repository in two different places:

1. Once, at application start-up, using the repository at index zero.
2. Inside the change event handler, using the index of the selected repo.

When you detect a repeated code like this you should immediately consider writing a function to remove this duplication. Can try and write such a function?

homework/index.js
ersarue: It should have been unintentionally changed while changing a parameter called `value` to `content` in somewhere else by using F2 key.

homework/index.js
ersarue: That's right, makes it redundant.

homework/index.js
ersarue: I' am editing it, no need to have it separately.

homework/index.js
ersarue: First of all, I should indicate it better Jim. I have only got help from  @ahmthsrvsmz for a small part of homework, such as listing only one repo on page load. Sorry for possible misleading if it's the case.
And the feedback for rendering, it's been piece of cake, thanks for the insight!

homework/index.js
ersarue: Right. I am gonna revisit the lines for DRY.

ersarue:
Dear Jim, thank you for the feedbacks! Things are much more clear now!

I should indicate it better Jim in first place. I have only got help from  @ahmthsrvsmz for a small part of homework, such as listing only one repo on page load. Sorry for possible misleading if it's the case.

homework/index.js
remarcmij: Are there any updates that I can review? This PR is still not approved.

ersarue: Hi Jim, indeed there are but still not ready to push. I really wished that I've submitted a renewed homework at least today. I'm planning to do so tomorrow afternoon. Thank you for your message!

ersarue:
Dear Jim, I've tried to implement your feedbacks. Now it's clearer and more functional, thanks! 

> Basically you need to execute the same code to render a repository in two different places:
> 
> Once, at application start-up, using the repository at index zero.
> Inside the change event handler, using the index of the selected repo.
> When you detect a repeated code like this you should immediately consider writing a function to remove this duplication. Can try and write such a function?

I actually did not dully understand here the point. Did you mean this for `fetchJSON`? And there appears a bullet when the browser uploads the page at first, I could not be able to get rid of that even though I have set `list-style: none;`. Could you help me with that?

homework/index.js
remarcmij: Your code is still not DRY (enough). If you carefully read my hint in the comment below, you can write a function that handles both the rendering of the first repo at startup and the rendering of the selected repo when the user changes the selection in the select box. When the user changes the selection, you need to wipe out the details of the previous selection (i.e. clearing out `.innerHTML`). However, although redundant because the containers are initially empty, there it is no problem if you also do that at start-up. With that in mind, you could write a function likes this:

```js
function render(repo, repoContainer, rightUl) {
  repoContainer.innerHTML = '';
  rightUl.innerHTML = '';
  renderRepoDetails(repo, repoContainer);
  getContributorsData(repo.contributors_url, rightUl);
}
```

In your main function, you could use it like this:

```js
...
render(repositories[0], repoContainer, rightUl);
  render(repositories[select.value], repoContainer, rightUl);
});
...
```

The function `renderSelect` is no longer needed.

homework/style.css
```diff
@@ -1,3 +1,119 @@
+* {
```
remarcmij: On a wide-screen monitor, your UI looks a bit stretched out. Perhaps set a `max-width` and horizontally center the content?

image

ersarue:
Hi Jim, I have applied your feedbacks and pushed the final state. Again thank you for the guiding feedbacks!

homework/index.js
```diff
@@ -29,25 +31,113 @@
...
+  }
+
+  function createTableRow(table, header, content) {
+    const row = createAndAppend('tr', table, { class: 'row' });
+    createAndAppend('th', row, { text: header, class: 'table-header' });
+    createAndAppend('td', row, { text: content, class: 'content' });
+    return row;
+  }
+
+  function renderRepoDetails(repo, repoName) {
```
remarcmij: This second parameter does not expect the name of the repo. It expect the container (parent) element in which to render the repo details. Please make an effort to be precise in your naming. Otherwise you are confusing the people trying to understand your code.

homework/style.css
```diff
@@ -1,3 +1,131 @@
...
+  margin: 0;
```
remarcmij: On a mobile device you should not use large left and right margins. Take full advantage of the limited screen width.

ersarue:
Thanks! I'll correct my codes based on the very last comments. 
I have a question though, in createAndAppend, wouldn’t it be better if appending child to parent (`parent.appendChild(elem);`) is executed after `Object.entries` is completed?
```
    Object.entries(options).forEach(([key, value]) => {
      if (key === 'text') {
        elem.textContent = value;
      } else {
        elem.setAttribute(key, value);
      }
    });
    parent.appendChild(elem);
    return elem;
  } 
```
So elem is made ready with all its content.

remarcmij:
That is a good observation, thanks. But I don't think it makes any difference in practice. While JavaScript is executing the browser is not rendering. It is only when the JavaScript call stack is empty that the browser starts examining what outstanding rendering work needs to be done and does it. But, intuitively it would make sense to do it after adding any attributes. I will change it in the repo when class23 is done with the homework. Otherwise we may get undesired merge conflicts. Thanks again.

PR #363, opened 2019-09-26
Week2

malnaqeeb:


homework/index.js
```diff
@@ -1,25 +1,28 @@
...
+      xhr.onerror = () => reject(new Error('Network request failed'));
+      xhr.send();
+    });
   }
 
   function createAndAppend(name, parent, options = {}) {
     const elem = document.createElement(name);
     parent.appendChild(elem);
-    Object.entries(options).forEach(([key, value]) => {
+    Object.keys(options).forEach(key => {
```
remarcmij: Can you please explain where this change came from? This code was used in an older version of the JavaScrip3 repo, as used by previous classes but not in the one used by class 23.

homework/index.js
```diff
@@ -29,25 +32,150 @@
...
+  function createRow(table, label, content) {
+    const row = createAndAppend('tr', table);
+    createAndAppend('th', row, { text: label });
+    createAndAppend('td', row, { text: content });
+    return row;
+  }
+
+  // 2- create details of repositories
+  function renderRepoDetails(detailsSection, repo) {
+    const li = createAndAppend('table', detailsSection, { class: 'item' });
```
remarcmij: Your variable name `li` suggests that you are creating an `<li>` element, whereas in truth you are erronously creating a `<table>` element. As you are also creating a `<table>` element in line 64, the resulting HTML is invalid:

```html
<section class="left-section">
  <table class="item">
    <table>
      <tr>
        <!-- etc -->
      </tr>
    </table>
  </table>
</section>
```

homework/index.js
```diff
@@ -29,25 +32,150 @@
...
+      href: repo.html_url,
+      text: repo.name,
+      target: '_blank',
+    });
+    // 2-2 add description
+    if (repo.description) {
+      createRow(table, 'Description:', repo.description);
+    } else {
+      createRow(table, 'Description:', 'N/A');
+    }
```
remarcmij: Instead of the `if` statement in lines 55-59 you can use a _short-circuit_ expression like this:

```js
createRow(table, 'Description:', repo.description || 'N/A');
```

I already gave this feedback in week 1 but somehow you seem to have missed it.

homework/index.js
```diff
@@ -29,25 +32,150 @@
...
+            });
+            createAndAppend('div', contributorDetails, {
+              text: contributor.contributions,
+              class: 'contributions-number',
+            });
+          });
+        } else {
+          createAndAppend('div', contributorSection, {
+            text: 'N/A',
+            class: 'alert-error',
```
remarcmij: It is not an error if there are no contributors. By using the `alert-error` class here you are giving a wrong visual clue to the user.

homework/index.js
```diff
@@ -29,25 +32,150 @@
...
+            });
+            createAndAppend('div', contributorDetails, {
+              text: contributor.login,
+              class: 'contributor-login',
+            });
+            createAndAppend('div', contributorDetails, {
+              text: contributor.contributions,
+              class: 'contributions-number',
+            });
+          });
```
remarcmij: This function is getting fairly large. It would be better to put the code from the arrow of the `.forEach()` (lines 83-107) into a separate, named function and use it instead of the arrow function:

```js
function renderContributor(contributor, ul) {
  ...
}

function createContributors(contributorSection, url) {
  ...
  contributors.forEach(contributor => renderContributor(contributor, ul));
  ...
}
```

homework/style.css
```diff
@@ -1,3 +1,137 @@
-.alert-error {
-  color: red;
+.alert-error {
```
remarcmij: On a mobile phone you are not using the full (scarce) screen width:

image

malnaqeeb:
Hi @remarcmij, Thanks for your feedbacks.

homework/index.js
remarcmij: I was hoping for an explanation.

homework/index.js
```diff
@@ -29,25 +32,148 @@
...
+        } else {
+          createAndAppend('div', contributorSection, {
+            text: 'N/A',
+            class: 'not-available',
+          });
+        }
+      })
+      .catch(err =>
+        createAndAppend('div', contributorSection, {
+          text: err.message,
```
remarcmij: Where did the `alert-error` class go?

homework/style.css
```diff
@@ -1,3 +1,139 @@
-.alert-error {
```
remarcmij: Why did you delete this class? It still necessary for errors. You use it in line 173 for your JavaScript file.

homework/index.js
malnaqeeb: HI @remarcmij, I forgot to add after I applied the modification. now it is there again. 

Thanks  :+1:

PR #364, opened 2019-09-27
add JS3 Week2 Homework

mehmetekinci:


homework/index.js
```diff
@@ -29,26 +34,128 @@
...
         });
-        return;
-      }
-      const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+      });
+  }
+  function renderPage(event, reposUl, contUl) {
+    const url = HYF_REPOS_URL;
+    fetchJSON(url)
```
remarcmij: There is no need to fetch the repository information each time the user changes the selected repo. You have fetched it already in function main. It would be wasteful, network-wise, to fetch it again.

homework/index.js
```diff
@@ -29,26 +34,128 @@
...
+    );
+
+    fetchJSON(url).then(repos => {
+      repos.sort(sortReposByNameProp).forEach((repo, index) => {
+        createAndAppend('option', select, {
+          value: index,
+          text: repo.name,
+        });
+      });
+      renderPage(event, reposUl, contUl);
```
remarcmij: The `event` argument used here is undefined. It produces an ESLint error causing your PR to fail the Travis test. It also causes a problem in your `renderPage` function which you tried to fix by adding `0` like this:

```js
const selectedRepo = sortedRepos[event.target.value || 0];
```

This, of course, is not the way to solve this.

In general, you should not pass events to functions other than event handlers. In this case, you could have passed `event.target.value` in the event handler and simply `0` for rendering the first repo at application startup.

However, as I said, you should only fetch the repositories once, at application startup:

```js
function main(url) {
  ...
  fetchJSON(url).then(repos => {
    repos.sort(sortReposByNameProp).forEach((repo, index) => {
        value: index,
        text: repo.name,
      });
    });
    select.addEventListener('change', event =>
      renderPage(repos[event.target.value], reposUl, contUl),
    );
    renderPage(repos[0], reposUl, contUl);
  });
}
```

Then, your `renderPage()` function could be simplified to:

```js
function renderPage(selectedRepo, reposUl, contUl) {
  renderRepoDetails(selectedRepo, reposUl);
  renderContributors(selectedRepo, contUl);
}
```

mehmetekinci:
hi @remarcmij thank you so much for your helpful feedbacks. I really want to learn how I can make my codes clear.  I am trying to do my best. I improved my codes.

homework/index.js
```diff
@@ -29,22 +31,115 @@
...
+    createAndAppend('a', firstRow.lastChild, {
+      text: repo.name,
+      href: repo.html_url,
+    });
+    addRepoDetailsInfo(table, 'Description :', repo.description);
+    addRepoDetailsInfo(table, 'Forks :', repo.forks);
+    addRepoDetailsInfo(table, 'Updated :', formatDate(repo.updated_at));
+  }
+
+  function renderContDetails(selectedRepo, contributionsUl) {
```
remarcmij: I may have missed this one on my previous review, but this parameter name is incorrect. In line 97 you are passing `contributors` as the first parameter, so that parameter name should also be `contributors`. As you can see, JavaScript does not care about the names you choose, it will run your code regardless. But for humans reading your code it becomes utterly confusing is you say A but mean B.

BTW in my previous comment in this line, I made a remark about 'Cont', and I was hoping that you concluded that a better name for this function would be `renderContributorDetails`.

homework/index.js
```diff
@@ -29,22 +31,115 @@
...
+    });
+
+    return trElm;
   }
 
-  function main(url) {
-    fetchJSON(url, (err, repos) => {
-      const root = document.getElementById('root');
-      if (err) {
+  function formatDate(stringDateAndTime) {
```
remarcmij: This escaped my eye in the previous review. This function does not take a string as an argument. Instead, it expects a Date object. Therefore this argument is best named simply `date`. I keep stressing on using accurate and precise names because your code becomes confusing if the names do not accurately reflect the data they represent.

mehmetekinci: thank you for the corrections. I just thought,I have to use string in my parameter when I use toLocaleString. now it is clear for me. I made it  _date_.

homework/index.js
mehmetekinci: I noticed after I read your feedback. I made corrections .

mehmetekinci:
hi @remarcmij. I forget to leave a comment. I completed last changes at last night.

mehmetekinci:
thank you @remarcmij. I am the happiest.

PR #365, opened 2019-09-28
week2

husenjanobul:


homework/index.js
```diff
@@ -29,23 +31,136 @@
...
+      { text: dateChange(repo.updated_at) },
+    );
+  }
+
+  function renderContributes(repositories, ul) {
+    const url = repositories.contributors_url;
+    fetchJSON(url)
+      .then(contributions => {
+        ul.innerHTML = ' ';
+        contributions.forEach(contributor => {
```
remarcmij: Rename to `contributors.forEach(contributor => {`

homework/index.js
```diff
@@ -29,23 +31,136 @@
...
+          const li = createAndAppend('li', ul);
+          const infoImg = createAndAppend('div', li, { class: 'infoImg' });
+          createAndAppend('img', infoImg, {
+            src: contributor.avatar_url,
+            alt: contributor.login,
+          });
+          const infoName = createAndAppend('div', li, { class: 'infoName' });
```
remarcmij: By convention, CSS identifiers (e.g. class names and ids) should be in **kebab-case**: all lowercase and words separated with hyphens: 

```js
const infoName = createAndAppend('div', li, { class: 'info-name' });
```

There are several places in your code where this needs to be corrected.

homework/index.js
```diff
@@ -29,23 +31,136 @@
...
+            target: '_blank',
+          });
+          createAndAppend('span', infoName, {
+            text: contributor.contributions,
+            class: 'number',
+          });
+        });
+      })
+      .catch(err => {
+        createAndAppend('div', root, {
```
remarcmij: The variable `root` is undefined here. This produces an ESLint error, causing the Travis check on your PR to fail. You can use `getElementById('root')` to avoid this.

```js
.catch(err => {
  createAndAppend('div', document.getElementById('root'), {
    text: err.message,
    class: 'alert-error',
  });
});
```

homework/index.js
```diff
@@ -29,23 +31,136 @@
...
+        renderContributes(repos[0], ulContributes);
+
+        select.addEventListener('click', () => {
+          const repo = repos[select.value];
+          renderRepoDetails(repo, ul);
+          renderContributes(repo, ulContributes);
+        });
+      })
+      .catch(err => {
         createAndAppend('div', root, {
```
remarcmij: See my previous comment on the variable `root`.

husenjanobul: Thank you for your feedback and advice. i'll check again homework and fix those details.

PR #366, opened 2019-09-28
N.Ince-JS3-HW2

NesibeInce:


homework/index.js
```diff
@@ -43,8 +90,44 @@
...
+        class: 'repos-container',
+      });
+      const contributorContainer = createAndAppend('section', mainContainer, {
+        class: 'contributor-container',
+      });
+      repos
+        .sort((a, b) => {
+          return a.name.localeCompare(b.name);
+        })
+        .map((elem, index) => {
```
remarcmij: You should the `.map()` method if you want to create a new array consisting of the transformed elements of an existing array. This is not the case here. Instead, you should `.forEach()` here. That will also get rid of the ESLint error that causes the Travis check to fail on this PR.

homework/index.js
```diff
@@ -43,8 +90,44 @@
...
+
+      select.addEventListener('change', () => {
+        reposContainer.innerHTML = '';
+        contributorContainer.innerHTML = '';
+        const repo = repos[select.value];
+        renderRepoDetails(repo, reposContainer);
+        contributorContainer.innerHTML = 'Contributors';
+        renderContributors(repo.contributors_url, contributorContainer);
+      });
     });
```
remarcmij: You are not displaying the details of the first repo at start-up, as required by the assignment. Because the `change` event only fires when the user actively changes the selection in select box, but not at start-up, you need run the same code that you have in the event handler once, in function main, outside of the event handler, using the repository at index zero:

```js
  reposContainer.innerHTML = '';
  contributorContainer.innerHTML = '';
  const repo = repos[select.value];
  renderRepoDetails(repo, reposContainer);
  contributorContainer.innerHTML = 'Contributors';
  renderContributors(repo.contributors_url, contributorContainer);
});

reposContainer.innerHTML = '';
contributorContainer.innerHTML = '';
const repo = repos[0];
renderRepoDetails(repo, reposContainer);
contributorContainer.innerHTML = 'Contributors';
renderContributors(repo.contributors_url, contributorContainer);
```

And it looks like you could write a function to eliminate this code duplication (make the code DRY).

homework/index.js
NesibeInce: hi @remarcmij ,I converted .map() into .forEach(). 
Thank you for the correction.

homework/index.js
NesibeInce: I defined a new function then I called it twice for 0 and the select.value.

homework/index.js
```diff
@@ -1,19 +1,21 @@
...
-      if (xhr.status >= 200 && xhr.status <= 299) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
```
remarcmij: - [x] Replace callback with promise

homework/index.js
```diff
@@ -29,23 +31,115 @@
     return elem;
   }
 
-  function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+  function renderRepoTable(table, name, path) {
+    const tr = createAndAppend('tr', table);
+    createAndAppend('th', tr, {
+      text: name,
```
remarcmij: The first column of the table containing the label text looks slightly better if you align them to the left. You can achieve that by adding another attribute for the `th` element: `align: 'left',`

homework/index.js
```diff
@@ -29,23 +31,115 @@
...
+          createAndAppend('h4', linkContributor, {
+            text: repo.login,
+          });
+          createAndAppend('p', linkContributor, {
+            text: repo.contributions,
+            class: 'contribution-count',
+          });
+        });
+      })
+      .catch(err => {
```
remarcmij: - [x] Handle errors

homework/index.js
```diff
@@ -29,23 +31,115 @@
...
+            });
+          });
+        function renderContent(elem) {
+          reposContainer.innerHTML = '';
+          contributorContainer.innerHTML = '';
+          const repo = repos[elem];
+          renderRepoDetails(repo, reposContainer);
+          contributorContainer.innerHTML = 'Contributors';
+          renderContributors(repo.contributors_url, contributorContainer);
+        }
```
remarcmij: You have correctly understood that you can create this function to make your code DRY. However, you avoid embedded named functions like this, because it is not easy to determine to which variables from the outer function they need access. Furthermore, it increases the size of the outer function, which makes it less readable.

If you move this function out of the `main` function, you will immediately see in VSCode various line having red underlines, indicating issues. Notably, this is the case with the variables `reposContainer`, `contributorContainer`, and `repos`.

```js
function renderContent(elem) {
  reposContainer.innerHTML = '';
  contributorContainer.innerHTML = '';
  const repo = repos[elem];
  renderRepoDetails(repo, reposContainer);
  contributorContainer.innerHTML = 'Contributors';
  renderContributors(repo.contributors_url, contributorContainer);
}

function main(url) {
  ...
}
```

Therefore, you need to add the missing variables as parameters to the function definition:

```js
function renderContent(elem, repos, reposContainer, contributorContainer) {
  ...
}
```

But if you look closer, you can see that you are using `elem` and `repos` inside the function only to get the repo that is to be rendered:

```js
const repo = repos[elem];
```

Therefore, it is better to let the caller of the function pass this repo as a parameter so that you can simplify the function definition to:

```js
function renderContent(repo, reposContainer, contributorContainer) {
  reposContainer.innerHTML = '';
  contributorContainer.innerHTML = '';
  renderRepoDetails(repo, reposContainer);
  contributorContainer.innerHTML = 'Contributors';
  renderContributors(repo.contributors_url, contributorContainer);
}
```

Then, in your main function, you can call this function with:

```js
renderContent(repos[0], reposContainer, contributorContainer)
```

homework/index.js
```diff
@@ -29,23 +31,115 @@
...
+          class: 'repos-container',
+        });
+        const contributorContainer = createAndAppend('section', mainContainer, {
+          class: 'contributor-container',
+        });
+        repos
+          .sort((a, b) => {
+            return a.name.localeCompare(b.name);
+          })
+          .forEach((elem, index) => {
```
remarcmij: A better name for the parameter `elem` is `repo`. After all, the `.forEach()` method is iterating through the (sorted) `repos` array, and a single element of that array is a `repo`.

Once you have used the correct name, it is easy to see that you can use the `repo` variable to set the `text` property:

```js
.forEach((repo, index) => {
    value: index,
    text: repo.name,
    class: 'select-class',
  });
});
```

homework/style.css
```diff
@@ -1,3 +1,114 @@
...
+
+.main-container {
+  display: flex;
+  align-items: flex-start;
+  justify-content: center;
+  flex-wrap: wrap;
+  padding-top: 10px;
+}
+
+a {
```
remarcmij: This CSS rule should only apply to the `<a>` in the contributor's section. It screws up the `<a>` in the repo table. Therefore you should restrict to children of the `contributor-details` class:

```css
.contributor-details a {
  ...
}
```

homework/index.js
NesibeInce: Thank you.

homework/index.js
NesibeInce: I added it to the CSS part, since I have not written any styling in .js page. Hope it is also ok like this.

homework/index.js
NesibeInce: I am happy that I was abla to do it.

homework/index.js
NesibeInce: I removed the function outside the main function, and rendered with :

renderContent(repos[0], reposContainer, contributorContainer);
renderContent(repos[select.value], reposContainer, contributorContainer)

homework/index.js
NesibeInce: elem is converted into repo, sir.

homework/style.css
NesibeInce: This is also done.
Thank you for your eagle eye.

homework/style.css
```diff
@@ -1,3 +1,118 @@
...
+
+td {
+  vertical-align: top;
+}
+
+td:first-child {
+  width: 100px;
+  min-width: 100px;
+  max-width: 100px;
+}
```
remarcmij: This comment from my previous feedback got lost, but you can improve the appearance of the table by replacing these rules with:

```css
td,
th {
  vertical-align: top;
}

th {
  width: 100px;
  min-width: 100px;
  max-width: 100px;
}
```

NesibeInce:
Ok, I will do so.
Thank you for your time and care.

homework/style.css
```diff
@@ -1,3 +1,118 @@
+body {
```
remarcmij: This comment was again somehow lost from my feedback of today, but this is what I also like you to improve: the UI when viewed on a mobile phone:

image

homework/index.js
```diff
@@ -29,23 +31,119 @@
...
+    createAndAppend('th', tr, {
+      text: name,
+    });
+    createAndAppend('td', tr, {
+      text: path,
+    });
+    return tr;
   }
 
+  function renderRepoDetails(repo, table) {
```
remarcmij: In the HTML you generate you are missing a `table` HTML tag. The parameter name `table` suggests that you are receiving a `table` element through the parameter list, but actually when you call the function you give it a `section` element. 

```html
<section class="repos-container">
  <tr>
    <th>Repository:</th>
    <td>
      <a href="https://github.com/HackYourFuture/alumni" target="_blank"
        >alumni</a
      >
    </td>
  </tr>
  <tr>
    <th>Description:</th>
  </tr>
  <tr>
    <th>Forks:</th>
    <td>9</td>
  </tr>
  <tr>
    <th>Updated:</th>
    <td>10/3/2019, 2:53:12 PM</td>
  </tr>
</section>
```

homework/style.css
```diff
@@ -1,3 +1,129 @@
...
+
+.repos-container,
+.contributor-container {
+  margin-bottom: 10px;
+  border: none;
+  border-radius: 5px;
+  background-color: white;
+  box-shadow: 0 1px 8px 0 rgba(0, 0, 0, 0.2), 0 3px 4px 0 rgba(0, 0, 0, 0.14), 0 3px 3px -2px rgba(0, 0, 0, 0.12);
+  padding: 10px;
+  width: 47%;
```
remarcmij: This 47% seems to be an arbitrary value. It doesn't work well on a mobile phone.

```suggestion
  flex: 1;
  margin: 8px 4px;
```

homework/style.css
```diff
@@ -1,3 +1,129 @@
...
+  width: 250px;
+  height: 40px;
+  padding: 3px;
+  font-size: 14px;
+  background-color: #f3f3f3;
+}
+
+th {
+  /* width: 80px; */
+  line-height: 25px;
```
remarcmij: This line-height causes misalignment.

```suggestion
```

homework/style.css
```diff
@@ -1,3 +1,129 @@
...
+
+.avatar {
+  border-radius: 5px;
+  margin-right: 15;
+  width: 100px;
+}
+
+.main-container {
+  display: flex;
+  align-items: flex-start;
```
remarcmij:
```suggestion
  align-items: stretch;
```

homework/style.css
```diff
@@ -1,3 +1,129 @@
...
+  justify-content: space-between;
```
remarcmij:
```suggestion
  justify-content: flex-start;
```

PR #367, opened 2019-09-28
JS3Week2 Homework

ihk2018:


homework/index.js
```diff
@@ -1,19 +1,21 @@
...
-      if (xhr.status >= 200 && xhr.status <= 299) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  function fetchJSON(url) {
+    return new Promise((resolve, reject) => {
```
remarcmij: Convert callback to promise: 👍

homework/index.js
```diff
@@ -29,23 +31,160 @@
...
+  }
+
+  function sortRepositoriesByName(a, b) {
+    return a.name.localeCompare(b.name);
+  }
+
   function renderRepoDetails(repo, ul) {
-    createAndAppend('li', ul, { text: repo.name });
+    ul.textContent = ' ';
+    const repoList = createAndAppend('li', ul, { class: 'rep-list' });
```
remarcmij: Be consistent. You have a variable `repoList` but a class name `rep-list`. Better make that `repo-list`. (I gave you this feedback already in week 1).

homework/index.js
```diff
@@ -29,23 +31,160 @@
...
+      createAndAppend('span', description, {
+        text: 'No Description Added Yet',
+        class: 'texts',
+      });
+    } else {
+      createAndAppend('span', description, {
+        text: repo.description,
+        class: 'texts',
+      });
+    }
```
remarcmij: Rather than repeating almost identical code in the `if` and `else` parts, you can use a so-called _short-circuit_ expression to deal with a missing description:

```js
createAndAppend('span', description, {
  text: repo.description || 'No Description Added Yet',
  class: 'texts',
});
```

The `text` attribute will be set to `repo.description` if it is truthy (a non-empty string is truthy), otherwise it will be set to `No Description Added Yet`.

homework/index.js
```diff
@@ -29,23 +31,160 @@
...
+    });
+
+    createAndAppend('span', updatedTime, {
+      text: convertDateTime(repo.updated_at),
+      class: 'texts',
+    });
+  }
+
+  function getContributorsInfo(repositories, ul) {
+    ul.textContent = ' ';
```
remarcmij: You must use `.innerHTML` here instead of `.textContent`. And set it to an empty string (`''`) instead of a space (`' '`).

homework/index.js
```diff
@@ -29,23 +31,160 @@
...
+    const url = repositories.contributors_url;
+    fetchJSON(url)
+      .then(contributions => {
+        contributions.forEach(contributor => {
```
remarcmij: Shorter function are preferred over longer functions, as it improves the readability of your code. In this case, I would replace the arrow function for this `.forEach()` with a named function like this:

```js
function renderContributor(contributor, ul) {
  const li = createAndAppend('li', ul, { class: 'contributors-li' });
  const contributorImage = createAndAppend('div', li, {
    class: 'contributor-image',
  });
  ...
}

function getContributorsInfo(repositories, ul) {
  ul.textContent = '';
  const url = repositories.contributors_url;
  fetchJSON(url)
    .then(contributions => {
      contributions.forEach(contributor =>
        renderContributor(contributor, ul),
      );
    })
   ...
}
```

homework/index.js
```diff
@@ -29,23 +31,160 @@
...
+        });
+
+        createAndAppend('h2', contributorContainer, {
+          text: 'Contributors',
+        });
+
+        const ulContributors = createAndAppend('ul', contributorContainer);
+        getContributorsInfo(repos[0], ulContributors);
+
+        select.addEventListener('click', () => {
```
remarcmij: You don't want this event handler to fire whenever this user clicks on the select box. It should fire only when the user changes the selection. The event to list for is `change`.

homework/index.js
ihk2018: thank you

homework/index.js
ihk2018: I corrected the class name, thanks @remarcmij

homework/index.js
ihk2018: I changed the rewrote the codes based on your feedback, and it is better now. I both save space and the codes looks nicer now 👍

homework/index.js
ihk2018: I changed the code based on the feedback, hopefully it is correct :), thank you @remarcmij

PR #371, opened 2019-09-30
Week3

rabrad:
Hi Jim, 
For your review. The first part of week3 homework.


homework/index.js
```diff
@@ -1,19 +1,14 @@
...
-    xhr.send();
+  async function fetchJSON(url) {
+    const response = await fetch(url);
+    if (!response.ok) {
+      throw new Error(
+        `Network error: ${response.status} - ${response.statusText}`,
+      );
+    }
+    return response.json();
   }
```
- [x] Use async/await

homework/style.css
```diff
@@ -1,3 +1,141 @@
+*,
```
remarcmij: I may have missed this on my previous reviews but on a mobile device your UI does not look responsive (enough). On an iPhone 6 it would like this, which is already cramped:

image


On an iPhone 5 with its smaller screen it becomes unusable:

image

rabrad: Thank you Jim. It was correct in the previous PR. 
I missed to update the container ID in my css.

I will send the updated version along with the rest of the homework as you suggested.

homework-classes/ContributorsView.js
```diff
@@ -19,8 +19,39 @@
...
     render(contributors) {
-      // TODO: replace this comment and the console.log with your own code
-      console.log('ContributorsView', contributors);
+      this.container.innerHTML = '';
+      createAndAppend('p', this.container, {
+        text: `Contributors:`,
+        id: 'contributors-title',
+        role: 'heading',
+      });
+      const div = createAndAppend('div', this.container, {
```
remarcmij: You seem to have missed the point that I made in slack about ARIA-compliance:

>For instance, for what is semantically supposed to be a list, a `ul` or `ol` element would be read out as a list. However, creating a list as a collection of `div` elements would not. In that case you would need to add ARIA attributes to the relevant HTML elements to help a screen reader recognize the semantics.

rabrad: Thank you so much for everything you've done over the past weeks. 
I've learned a lot from you!
I will continue to dig into the ARIA documentation until I find a way to use its attributes correctly.

remarcmij: Glad to hear, Rabie 👍.

For ARIA-compliance in this case, you only need to replace your nested divs with a `ul` with `li` children. This is semantically a list that is ARIA-compliant. A screen reader will identify that as a list. But it would not know what to make of a `div` unless you add ARIA attributes to explain its role. But why choose that over a `ul`?

rabrad: I thought I shouldn't replace the divs with `ul` and `li `  in the borrowed snippets. Because of that I was looking for alternative attributes to identify my divs. 

Well, this is what beginners would mistakenly do :)

I will update the work

PR #372, opened 2019-09-30
Week3

husenjanobul:


homework/index.js
```diff
@@ -1,19 +1,14 @@
...
-    xhr.send();
+  async function fetchJSON(url) {
+    const response = await fetch(url);
+    if (!response.ok) {
+      throw new Error(
+        `Network error: ${response.status} - ${response.statusText}`,
+      );
+    }
+    return response.json();
   }
```
- [x] Use async/await

homework/index.js
```diff
@@ -29,23 +24,110 @@
...
-    fetchJSON(url, (err, repos) => {
-      const root = document.getElementById('root');
-      if (err) {
-        createAndAppend('div', root, {
-          text: err.message,
-          class: 'alert-error',
+  async function renderContributors(repositories, ul) {
+    const url = repositories.contributors_url;
+    try {
+      const data = await fetchJSON(url);
```
remarcmij: The word `data` is rather meaningless. The usage of that word should be reserved for cases where you absolutely know nothing about the nature of the data it refers to. In this case however, you know that you are fetching an array of contributors. Therefore, you should use the more precise name `contributors` here.

Then, your `.forEach()` down below can read as:

```js
```

homework/index.js
```diff
@@ -29,23 +24,110 @@
...
+    try {
+      const root = document.getElementById('root');
+      const header = createAndAppend('header', root, {
+        text: 'HYF Repositories',
+        class: 'title',
+      });
+      const select = createAndAppend('select', header, {
+        class: 'selection',
+      });
+      const data = await fetchJSON(url);
```
remarcmij: `data` -> `repositories` (or `repos`)

homework/style.css
```diff
@@ -1,3 +1,176 @@
+* {
```
remarcmij: On a mobile phone this looks wrong:

image

homework/index.js
husenjanobul: thank you for feedbacks , i will review and update those problems.

homework-classes/RepoView.js
```diff
@@ -19,8 +19,27 @@
...
      */
     render(repo) {
-      // TODO: replace this comment and the console.log with your own code
-      console.log('RepoView', repo);
+      this.container.innerHTML = ' ';
+      const ul = createAndAppend('ul', this.container, {
+        class: 'list-container',
+      });
+      
+      function createDetailLine(ul, labelText, value) {
```
remarcmij: Instead of creating an embedded function like this, it is a best practice when using classes to create a method. In this case, that can be a **static** method, as the function/method body does not reference the **this** value:

```js
render(repo) {
  const first = RepoView.createDetailLine(ul, 'Repository', '');
...
}

static createDetailLine(ul, labelText, value) {
  ...
}
```

PR #373, opened 2019-09-30
Week3

msnmz:


homework-classes/ContributorsView.js
```diff
@@ -19,8 +19,33 @@
...
      */
     render(contributors) {
-      // TODO: replace this comment and the console.log with your own code
-      console.log('ContributorsView', contributors);
+      this.container.innerHTML = '';
+      createAndAppend('label', this.container, {
+        text: 'Contributions:',
+      });
+      const ul = createAndAppend('ul', this.container, {
+        class: 'contributers-list',
```
remarcmij: VSCode flags a spelling error here.

homework-classes/RepoView.js
```diff
@@ -19,8 +19,39 @@
...
+        href: repo.html_url,
+        text: repo.name,
+        target: '_blank',
+      });
+      this.appendRepoDetail('Description:', repo.description, repoDetailsTable);
+      this.appendRepoDetail('Forks:', repo.forks, repoDetailsTable);
+      const timeRow = this.appendRepoDetail('Updated:', '', repoDetailsTable);
+      createAndAppend('time', timeRow.lastChild, {
+        text: this.getDateTimeText(repo.updated_at),
+        datetime: repo.updated_at,
```
remarcmij: What do you expect to achieve with this `datetime` attribute?

msnmz:
Hi @remarcmij, thank you for your intensive effort on us. I searched for ARIA compliance for HTML elements, and could not find a way to make human-readable date-time text. I encountered with the `<time> ` tag and its `datetime`  attribute. But it did not work at all. And I changed the way of displaying time string.

homework-classes/RepoView.js
```diff
@@ -19,8 +19,38 @@
...
+      });
+      this.appendRepoDetail('Description:', repo.description, repoDetailsTable);
+      this.appendRepoDetail('Forks:', repo.forks, repoDetailsTable);
+      const timeRow = this.appendRepoDetail('Updated:', '', repoDetailsTable);
+      createAndAppend('time', timeRow.lastChild, {
+        text: this.getDateTimeText(repo.updated_at),
+      });
+    }
+
+    appendRepoDetail(header, description, parentTable) {
```
remarcmij: Strictly speaking, this method could be `static` as it does not reference inside its method body. The same goes for the `getDateTimeText` method. Static methods are attached to the class itself (actually, the underlying constructor function of the class). Regular methods are attached to the prototype object.

This is just for info. No need to change anything here.

PR #374, opened 2019-09-30
Week 3

ahmthsrvsmz:


homework/index.js
```diff
@@ -1,19 +1,14 @@
...
-    xhr.send();
+  async function fetchJSON(url) {
+    const response = await fetch(url);
+    if (!response.ok) {
+      throw new Error(
+        `Network error: ${response.status} - ${response.statusText}`,
+      );
+    }
+    return response.json();
   }
```
- [x] Use async/await

homework/index.js
```diff
@@ -29,23 +24,131 @@
...
+  function createOption(repo, parent, index) {
+    createAndAppend('option', parent, {
+      text: repo.name,
+      value: index,
+    });
+  }
+
+  async function createContributorsSection(url, parent) {
+    try {
+      const contributes = await fetchJSON(url);
```
remarcmij: You are fetching an array of **contributors** (people that have contributed to the repo), not contributes. Such a single person is a **contributor**.

homework/index.js
```diff
@@ -29,23 +24,131 @@
...
+        });
+      });
+    } catch (err) {
+      const root = document.getElementById('root');
+      createAndAppend('div', root, {
+        text: err.message,
+        class: 'alert-error',
+      });
+    }
+  }
```
remarcmij: - [x] async/await
- [x] try/catch

homework/index.js
```diff
@@ -29,23 +24,131 @@
...
+        changeSelection(sortedRepos[select.value], repoParent, contParent);
+      });
+    } catch (err) {
+      const root = document.getElementById('root');
+      createAndAppend('div', root, {
+        text: err.message,
+        class: 'alert-error',
+      });
+    }
+  }
```
remarcmij: - [x] async/await
- [x] try/catch

homework/index.js
ahmthsrvsmz: I will fix 👍 Thanks

homework-classes/ContributorsView.js
```diff
@@ -19,8 +19,46 @@
      * @param {Object[]} contributors An array of contributor objects
      */
     render(contributors) {
-      // TODO: replace this comment and the console.log with your own code
       console.log('ContributorsView', contributors);
```
remarcmij: Remember to take out debug-type console.log statements out of production code (which a PR always is).

homework-classes/RepoView.js
```diff
@@ -18,9 +18,40 @@
      * Renders the repository details.
      * @param {Object} repo A repository object.
      */
+
     render(repo) {
-      // TODO: replace this comment and the console.log with your own code
       console.log('RepoView', repo);
```
remarcmij: Remove console.log.

homework-classes/ContributorsView.js
```diff
@@ -19,8 +19,46 @@
...
+      const contributorsContainer = document.querySelector(
```
remarcmij: This container is already available to this class in the `this.container` class member. Therefore don't try and get it with `document.querySelector`.

homework-classes/ContributorsView.js
```diff
@@ -19,8 +19,46 @@
...
+        '.contributors-container',
+      );
+      const select = document.querySelector('.repo-select');
```
remarcmij: Handling the `select` element is the responsibility of the `HeaderView` class. You should not reference it here.

homework-classes/ContributorsView.js
```diff
@@ -19,8 +19,46 @@
...
+      const p = createAndAppend('p', contributorsContainer, {
+        text: 'Contributions',
+      });
+
+      async function createContributorsSection(contributors, parent) {
```
remarcmij: This need not be an `async` function, You are not using `await` inside the function body. Fetching data is the responsibility of the **Model** class. Therefore, the `try/catch` block is also unwanted. In fact, the responsibility to render error is that of the `ErrorView` class.

Finally, this function should not be a function at all as you call immediately after its definition (see line 57).

homework-classes/ContributorsView.js
```diff
@@ -19,8 +19,46 @@
      * @param {Object[]} contributors An array of contributor objects
      */
     render(contributors) {
-      // TODO: replace this comment and the console.log with your own code
       console.log('ContributorsView', contributors);
+      const contributorsContainer = document.querySelector(
+        '.contributors-container',
+      );
+      const select = document.querySelector('.repo-select');
+      const p = createAndAppend('p', contributorsContainer, {
```
remarcmij: The variable `p` is unused and should therefore not exist.

homework-classes/ContributorsView.js
```diff
@@ -19,8 +19,46 @@
...
+          createAndAppend('div', root, {
+            text: err.message,
+            class: 'alert-error',
+          });
+        }
+      }
+      createContributorsSection(contributors, contributorsContainer);
+
+      select.addEventListener('change', () => {
+        contributorsContainer.innerHTML = '';
```
remarcmij: You should no reference the `select` element here. It is the exclusive responsibility of the `HeaderView` to deal with that element. Besides, you do not need an event handler simply to erase previous content on each call to the `render` method. Just use this at the top the `render` method:

```js
this.container.innerHTML = '';
```

homework-classes/Model.js
```diff
@@ -39,13 +39,13 @@
...
 
-    static fetchJSON(url) {
-      return fetch(url).then(res => {
-        if (!res.ok) {
-          return new Error(`HTTP ${res.status} - ${res.statusText}`);
-        }
-        return res.status === 200 ? res.json() : null;
-      });
+    static async fetchJSON(url) {
+      const response = await axios.get(url);
```
remarcmij: - [x] Replace fetch with axios

homework-classes/Model.js
```diff
@@ -39,13 +39,13 @@
...
+      if (!response.status === 200) {
```
remarcmij: As @eidosam pointed out in the lecture, you don't need this test here as **axios** does it for you.

homework-classes/RepoView.js
```diff
@@ -18,9 +18,40 @@
...
+      const repoContainer = document.querySelector('.repo-container');
+      const select = document.querySelector('.repo-select');
+
+      function addRow(table, title, repoText) {
```
remarcmij: When using classes, it is a best practice to use class methods instead of embedded functions. In this case, because the function body does not reference the `this` value, this could be a **static** method:

```js
render(repo) {
  ...
  RepoView.addRow(table, 'Description:', repo.description, 'td');
  ...
}

static addRow(table, title, repoText) {
  ...
}
```

homework-classes/RepoView.js
```diff
@@ -18,9 +18,40 @@
      * Renders the repository details.
      * @param {Object} repo A repository object.
      */
+
     render(repo) {
-      // TODO: replace this comment and the console.log with your own code
       console.log('RepoView', repo);
+      const repoContainer = document.querySelector('.repo-container');
+      const select = document.querySelector('.repo-select');
```
remarcmij: Do not get references like this from the DOM. The container is already available as `this.container`. And you should not reference the `select` element here. That is the responsibility of the `HeaderView` class.

homework-classes/RepoView.js
```diff
@@ -18,9 +18,40 @@
...
+      function addRow(table, title, repoText) {
+        const tr = createAndAppend('tr', table);
+        createAndAppend('th', tr, {
+          text: title,
+        });
+        createAndAppend('td', tr, { text: repoText });
+        return tr;
+      }
+
+      function renderRepoDetails(repo, parent) {
```
remarcmij: Besides my earlier point about embedded functions, defining this code as a function is unneeded as you call it immediately in line 50.

homework-classes/RepoView.js
```diff
@@ -18,9 +18,40 @@
...
+        });
+
+        addRow(table, 'Description:', repo.description, 'td');
+        addRow(table, 'Forks:', repo.forks, 'td');
+        addRow(table, 'Updated:', date.toLocaleDateString(), 'td');
+      }
+      renderRepoDetails(repo, repoContainer);
+
+      select.addEventListener('change', () => {
+        repoContainer.innerHTML = '';
```
remarcmij: You don't need to have this code inside an event listener. Just erase the previous content each time the `render` method is called by placing this line at the top of that function:

```js
this.container.innerHTML = '';
```

ahmthsrvsmz:
Dear @remarcmij
I think I still work result-oriented. I will work on it again.

homework-classes/RepoView.js
ahmthsrvsmz: I don't understand the static mode but I understand class and write this js page again

remarcmij: You can also use a 'regular' method like this:

```js
render(repo) {
  ...
  this.addRow(table, 'Description:', repo.description, 'td');
  ...
}

addRow(table, title, repoText) {
  ...
}
```

Static methods are preferred (but not required) if the method body does not contain a reference to the `this` value, which holds true for your `addRow` method. If you _do_ use a static method then you can't call it with `this.myMethodName()`. Instead, you must the class name, e.g. `MyClass.myMethodName()`.

See also: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static

PR #378, opened 2019-10-01
JS3-WEEK3-PART1

malnaqeeb:


homework/index.js
```diff
@@ -1,25 +1,19 @@
...
+    }
+    return response.json();
   }
 
   function createAndAppend(name, parent, options = {}) {
     const elem = document.createElement(name);
     parent.appendChild(elem);
-    Object.entries(options).forEach(([key, value]) => {
+    Object.keys(options).forEach(key => {
+      const value = options[key];
```
remarcmij: I'm still expecting an answer from you where this code change come from. It is code that we used in previous classes but not in your class.

homework/index.js
```diff
@@ -1,25 +1,19 @@
...
-      if (xhr.status >= 200 && xhr.status <= 299) {
-        cb(null, xhr.response);
-      } else {
-        cb(new Error(`Network error: ${xhr.status} - ${xhr.statusText}`));
-      }
-    };
-    xhr.onerror = () => cb(new Error('Network request failed'));
-    xhr.send();
+  async function fetchJSON(url) {
+    const response = await fetch(url);
```
- [x] Use async/await

homework/index.js
```diff
@@ -29,25 +23,141 @@
...
+  function createRow(table, label, content) {
+    const row = createAndAppend('tr', table);
+    createAndAppend('th', row, { text: label });
+    createAndAppend('td', row, { text: content });
+    return row;
+  }
+
+  // 2- create details of repositories
+  function renderRepoDetails(detailsSection, repo) {
+    const li = createAndAppend('li', detailsSection, { class: 'item' });
```
remarcmij: An `li` tag cannot be the child of a `section` tag. That constitutes invalid HTML.

homework/index.js
```diff
@@ -29,25 +23,141 @@
...
+      class: 'contributions-number',
+    });
+  }
+
+  async function createContributors(contributorSection, url) {
+    createAndAppend('h4', contributorSection, {
+      text: 'Contributions:',
+    });
+    try {
+      const contributors = await fetchJSON(url);
```
remarcmij: - [x] try/catch
- [x] async/await

homework/index.js
```diff
@@ -29,25 +23,141 @@
...
       }
-      const ul = createAndAppend('ul', root);
-      repos.forEach(repo => renderRepoDetails(repo, ul));
+    } catch (err) {
+      createAndAppend('div', contributorSection, {
+        text: err.message,
+      });
+    }
+  }
+  async function main(url) {
```
remarcmij: Separate function definitions with a blank line.

homework/index.js
```diff
@@ -29,25 +23,141 @@
...
+    const header = createAndAppend('header', root, {
+      class: 'header',
+    });
+    const title = createAndAppend('p', header);
+    title.innerText = 'HYF Repositories';
+    const dropList = createAndAppend('select', header, {
+      id: 'drop-list',
     });
+    try {
+      const repositories = await fetchJSON(url);
```
remarcmij: - [x] try/catch
- [x] async/await

homework/index.js
malnaqeeb: Hi @remarcmij, yes it's a code from previous classes i guess. because when i was trying to do the homework i was need to see the old pull request because i didn't understand the homework correctly. I saw few olds codes and I copied it to my  VS code just to understand it.

remarcmij: Of course, I already suspected that. If you don't understand the homework, you should ask in slack, Copying from others, whether it be classmates or homework of previous classes goes against the HYF code of conduct as stated in the Student Manual.

malnaqeeb: Yes but I just was trying to understand one or two point because it will not be efficient if i take the whole code.  i have never did that

